PR #2, opened 2023-04-24
feat: initial implementation

olevski:
The initial implementation without any tests.

I will open a followup PR for the tests.

Refer to the Makefile and README on how to run the server and then also the tests with `schemathesis`.


src/renku_crac/app.py
```diff
@@ -0,0 +1,369 @@
...
+                resource_pool_id=resource_pool_id, resource_class_id=class_id, **body.dict(exclude_none=True)
+            )
+            return json(apispec.ResourceClassWithId.from_orm(cls).dict(exclude_none=True))
+
+        return handler
+
+    def _post_resource_pool_class(self):
+        @validate(json=apispec.ResourceClass)
+        async def handler(_: Request, body: apispec.ResourceClass, resource_pool_id: int):
+            rps = await self.config.db.get_resource_pools(id=resource_pool_id)
```
Panaetius, line 170: Are duplicate names for resource classes an issue? If so we should check for it here and add a unique constraint on the DB

src/renku_crac/app.py
```diff
@@ -0,0 +1,369 @@
...
+        async def handler(_: Request):
+            res = await self.config.db.get_resource_pools()
+            return json([apispec.ResourcePoolWithId.from_orm(r).dict(exclude_none=True) for r in res])
+
+        return handler
+
+    def _post_resource_pools(self):
+        @validate(json=apispec.ResourcePool)
+        async def handler(_: Request, body: apispec.ResourcePool):
+            rp = models.ResourcePool.from_dict(body.dict())
```
Panaetius, line 43: should we check for duplicate names? And add a unique constraint on the DB for name

src/users/dummy.py
```diff
@@ -0,0 +1,45 @@
...
+
+    @property
+    def user_always_exists(self) -> bool:
+        """If true then every query for a user will make a user that matches the query."""
+        return self._user_always_exists
+
+    @user_always_exists.setter
+    def user_always_exists(self, value: bool):
+        """Control over the users that are queried should exist or not."""
+        self._user_always_exists = value
```
Panaetius, line 25: since the setter doesn't do anything special, can this just be a regular @property?

Panaetius (changes requested):
Looks really nice. I think you may want to change `import src.a`/`from src import a` to just `import a` and not mention `src` as src isn't a real package. And include them in the packages in pyproject.toml

src/users/dummy.py
olevski: @Panaetius by regular `@property` you mean not an `@property` but rather just simply have `self.user_always_exists`, right? Because if I remove the setter then I cannot change the user store to give me a 404 when testing via the setter. And I will want to do this when running tests. That is the main reason for this. This would not be used in production only for testing or development.

```python
class DummyUserStore:
    """A dummy adapter for keycloak. By default it will create and return users that do not exist."""

    def __init__(self, *, user_always_exists: bool = True):
        self._user_always_exists = user_always_exists

    @property
    def user_always_exists(self) -> bool:
        """If true then every query for a user will make a user that matches the query."""
        return self._user_always_exists
```

src/renku_crac/app.py
olevski: I dont want to enforce unique names for resource pools. We have ids to uniquely determine resource pools. So I want users to be able to name resource pools however they want. For making the k8s resources I will append either the id from the db or a uuid. I think with this in mind it may be better to then use uuid in the endpoints rather than the ids from the database. You already mentioned this in the previous review. I will open a separate PR for this.

src/renku_crac/app.py
olevski: I do not want to enforce unique names on resource classes at all. So if users want to make several resource classes with the same name they can do so. We will always go by the unique id in the db. The names are always for display purposes only.

src/renku_crac/app.py
```diff
@@ -0,0 +1,369 @@
...
+                user = users[0]
+                if user.username != body.username:
+                    # The username in keycloak is different, so update the user to have the username from keycloak
+                    user = await self.config.db.update_user(keycloak_id=cast(int, user.id), username=body.username)
+                return json(
+                    apispec.UserWithId(id=user.keycloak_id, username=cast(EmailStr, user.username)).dict(
+                        exclude_none=True
+                    ),
+                    200,
+                )
```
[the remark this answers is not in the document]
olevski: I think this is overly strict and inconvenient for whoever is using the API. Especially for the users endpoint. Here the POST is idempotent because we cannot have duplicates for the user keycloak id or username. So if someone wants to create the same user that is already in the db I can simply return it. There is no reason to error out when the state that the caller wants is already there.

I just think it is really weird and unfair to send a 409 and force the caller to send another GET request when the state they want the db to be in is already there.

src/users/dummy.py
olevski: Let me know what you think. Either way this is really inconsequential because it will never run in production.

src/renku_crac/app.py
```diff
@@ -0,0 +1,369 @@
...
+
+from src import models
+from src.models import errors
+from src.renku_crac.config import Config
+from src.renku_crac.error_handler import CustomErrorHandler
+from src.schemas import apispec
+
+
+@dataclass
+class Server:
```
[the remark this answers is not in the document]
olevski: Ah I did not notice this in the docs. This should work nicely.

olevski:
> I think you may want to change import src.a/from src import a to just import a and not mention src as src isn't a real package.

Python package setup is always a bit mysterious to me. Removing the `src.` part form the imports worked. But then whatever I did I could not keep mypy happy. And I kept getting this error:

```
src/renku_crac/config.py:9:1: error: Skipping analyzing "db.adapter": module is
installed, but missing library stubs or py.typed marker  [import]
    from db.adapter import ResourcePoolRepository, UserRepository
    ^
```

I am really not sure what I am missing. It seems that every little cosmetic change like this I make really upsets mypy. How would you do this?

src/users/dummy.py
Panaetius: I meant 
```
class DummyUserStore:
    user_always_exists = True

    def __init__(self, *, user_always_exists: bool = True):
        self.user_always_exists = user_always_exists

```

Since there's not really anything gained by having setter/getters. But as you say, it doesn't really matter :slightly_smiling_face:

src/renku_crac/app.py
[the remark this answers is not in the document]
Panaetius: fair enough.

Another thing in the code above that confuses me is that if the user doesn't exist and the username doesn't match the keycloak username, we error out. but if the user already exists and the names don't match, we set the name to the supplied name, contradicting the comment "update the user to have the username from keycloak". Right now it's "update the user to have the username from the request".
And if the names need to match, do we even need to pass `username` in the request, if all that matters is the keycloak `id`?

src/users/dummy.py
olevski: Ok got it. I can change that.

src/renku_crac/app.py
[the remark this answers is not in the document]
olevski: Ah yeah that is a very good point. Sorry I think I got lost in my own logic there. Let me fix this.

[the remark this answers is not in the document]
olevski: So after looking at this again and thinking in general I think it is better to remove the username. We are just asking for trouble because we can easily get into a situation where the username in our db is different form keycloak for the same user. And then we can have all sorts of unpredictable behaviour. So until we are certain we need to have the usernames from keycloak in our db it is better to simply leave them out.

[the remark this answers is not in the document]
olevski: So I removed the username field from our db and also simplified the logic you were talking about. I think it is now clearer. But let me know I can modify or clarify thing further.

Panaetius (approved):

PR #5, opened 2023-05-03
chore: add actions

olevski:
So the coverage shows up on coveralls. I am pretty certain that the comment in the PR does not show up because there is nothing to compare it against. I.e. the coverage on main has not been uploaded. This should be fixed once we merge this.


.github/workflows/tests.yml
```diff
@@ -0,0 +1,41 @@
...
+      - uses: actions/checkout@v3
+      - name: Install poetry
+        run: pipx install poetry
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.11"
+          cache: "poetry"
+      - name: Install dependencies
+        run: poetry install
```
Panaetius, line 23: I'm confused how this actually works. I mean, the test passes, but since we don't call it with `--with dev`, where does it get `pytest` etc. from?

olevski: I think they changed this a few times recently. But also locally I never passed any flags like `--with-dev` and I have all the packages installed.

Panaetius: ah I got so used to having to do it manually all the time :slightly_smiling_face:

Panaetius (approved):

PR #7, opened 2023-05-08
feat: add authentication

olevski:
Adds authentication and checks against keycloak.


src/renku_crac/app.py
```diff
@@ -4,347 +4,534 @@
...
+        @only_admins
+        @validate(json=apispec.UserWithId)
+        async def _post(request: Request, body: apispec.UserWithId, user: models.APIUser):
+            if user.access_token is None:
+                raise errors.Unauthorized()
+            users_db, user_kc = await asyncio.gather(
+                self.repo.get_users(keycloak_id=body.id, api_user=user),
+                self.user_store.get_user_by_id(body.id, user.access_token),
+            )
+            user_db = users_db[0] if len(users_db) >= 1 else None
```
Panaetius, line 370: `next(users_db, None)` should work and is a bit more concise.

src/renku_crac/auth.py
```diff
@@ -0,0 +1,43 @@
+"""Authentication decorators for Sanic."""
+from functools import wraps
+
+from sanic import Request
+
+from models import APIUser, Authenticator, errors
+
+
+def authenticate(authneticator: Authenticator):
```
Panaetius, line 9:
```suggestion
def authenticate(authenticator: Authenticator):
```

src/renku_crac/config.py
```diff
@@ -31,11 +47,47 @@ def from_env(cls):
...
+                    message="The JWKS url for Keycloak cannot be found from the OIDC discovery endpoint."
+                )
+            algorithms = os.environ.get(f"{prefix}KEYCLOAK_TOKEN_SIGNATURE_ALGS")
+            if algorithms is None:
+                raise errors.ConfigurationError(message="At least one token signature algorithm is requried.")
+            algorithms_lst = [i.strip() for i in algorithms.split(",")]
+            jwks = PyJWKClient(jwks_url)
+            authenticator = KeycloakAuthenticator(jwks=jwks, algorithms=algorithms_lst)
+            user_store = KcUserStore(keycloak_url=keycloak_url, realm=keycloak_realm)
+
```
Panaetius, line 86: Should we use `dataconf` to make it easier to do this kind of thing instead of all the manual `os.environ` calls?

Panaetius (changes requested):

src/renku_crac/app.py
```diff
@@ -4,347 +4,534 @@
...
-            raise errors.ValidationError(message="The provided user ID does not match the ID from keycloak.")
-        # The user exists in the db and the request body matches what is the in the db, simply return the user.
-        if user_db is not None and user_db.keycloak_id == body.id:
+
+        @authenticate(self.authenticator)
+        @only_admins
+        @validate(json=apispec.UserWithId)
+        async def _post(request: Request, body: apispec.UserWithId, user: models.APIUser):
+            if user.access_token is None:
+                raise errors.Unauthorized()
```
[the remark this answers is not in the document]
olevski: This is actually not even needed here. This case is handled by the `@only_admins` decorator.

src/renku_crac/config.py
olevski: Ralf I would like to address this in a separate PR if that is ok wiht you?

Panaetius: yes that's perfectly fine

src/db/adapter.py
```diff
@@ -190,19 +180,27 @@ async def get_classes(
...
             )
+            if resource_pool_id is not None:
+                stmt = stmt.where(schemas.ResourcePoolORM.id == resource_pool_id)
+            if id is not None:
+                stmt = stmt.where(schemas.ResourceClassORM.id == id)
+            if name is not None:
+                stmt = stmt.where(schemas.ResourceClassORM.name == name)
+            # NOTE: The line below ensures that the right users can access the right resources, do not remove.
+            stmt = _classes_user_access_control(api_user, stmt)
+            stmt = stmt
```
Panaetius, line 196: I think this line is a leftover that should be removed

Panaetius (changes requested):
One tiny change, otherwise it looks great now :+1:

PR #8, opened 2023-05-09
chore: add docker build

olevski:
Depends on #7 and should be merged after it.


Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM python:3.11-bullseye as builder
```
Panaetius, line 1: did you give https://hub.docker.com/r/sanicframework/sanic a try?

olevski: @Panaetius I did not know about those images. But now I see that they have medium and/or high vulnerabilities when I scan then with snyk. Whereas the images we use only have low vulnerabilities. 

Also the images are all based on alpine. And installing python packages with alpine takes forever.

Lastly, the sanic image sizes (without any of out stuff installed) are already at ~120MB whereas the python/debian images here with our stuff all installed and ready to go are ~80MB.

So it is not worth using the Sanic images.

Panaetius (approved):

PR #9, opened 2023-05-10
chore: changes for Postgres and Helm

olevski:


Panaetius (approved):

src/renku_crc/server_options.py
```diff
@@ -0,0 +1,167 @@
...
+        max_elems = 0
+        largest_list = ""
+        for k, v in self.dict().items():
+            options = v.get("options", [])
+            if k == "disk_request":
+                continue
+            if len(options) > max_elems:
+                max_elems = len(options)
+                largest_list = k
+        return largest_list
```
Panaetius, line 100:
```suggestion
       options = ((k, v.get("options", [])) for k,v in self.dict().items() if k != "disk_request")
       largest_list = max(options, key=lambda t: len(t[1]))[0]
       return largest_list
```
a bit shorter, not that it really makes a difference though

src/renku_crc/server_options.py
```diff
@@ -0,0 +1,167 @@
...
+    def __iter__(self):
+        self._current_ind = 0
+        return self
+
+    def __next__(self):
+        curr_ind = self._current_ind
+        self._current_ind += 1
+        if curr_ind >= len(self._vals):
+            return ("x" * (curr_ind - len(self._vals) + 1)) + self._vals[-1]
+        return self._vals[curr_ind]
```
Panaetius, line 117:
```suggestion
def _get_classname():
    yield "small"
    yield "medium"
    yield "large"
    count = 1
    while True:
        yield "x"*count + "large"
        count += 1
```

and you can use it the same as your current iterator (` class_names = _get_classname()` and then calling next on it)

src/renku_crc/server_options.py
```diff
@@ -0,0 +1,167 @@
...
+    largest_attribute = server_options.find_largest_attribute()
+    options_xref = {
+        "cpu_request": "cpu",
+        "mem_request": "memory",
+        "gpu_request": "gpu",
+    }
+    class_names = _ClassNameIter()
+    largest_attribute_options = getattr(getattr(server_options, largest_attribute), "options")
+    max_storage = round(sorted(server_options.disk_request.options)[-1] / 1_000_000_000)
+    for ival, val in sorted(enumerate(largest_attribute_options)):
```
Panaetius, line 134: I think it should be `enumerate(sorted(...))` otherwise it just sorts by the enumerate, i.e. it stays in the same order

src/renku_crc/server_options.py
```diff
@@ -0,0 +1,167 @@
...
+    clses: Set[models.ResourceClass] = set()
+    largest_attribute = server_options.find_largest_attribute()
+    options_xref = {
+        "cpu_request": "cpu",
+        "mem_request": "memory",
+        "gpu_request": "gpu",
+    }
+    class_names = _ClassNameIter()
+    largest_attribute_options = getattr(getattr(server_options, largest_attribute), "options")
+    max_storage = round(sorted(server_options.disk_request.options)[-1] / 1_000_000_000)
```
Panaetius, line 133:
```suggestion
    max_storage = round(max(server_options.disk_request.options) / 1_000_000_000)
```

src/k8s/clients.py
```diff
@@ -0,0 +1,154 @@
+"""Different implementations of k8s clients."""
+from copy import deepcopy
+from threading import Lock
```
Panaetius, line 3: does threading.lock work since there might be multiple worker processes on sanic?

olevski: That is a good point. I am not sure but just to be safe I switched it to multiprocessing.Lock. It is awesome that both support context managers. Also this did not matter all that much because it is only used in the dummy adapters. But still it is better to be safe than get stuck chasing weird errors because of some bug like this.

src/k8s/quota.py
```diff
@@ -0,0 +1,122 @@
...
+    _label_value: str = field(init=False, default="renku")
+
+    def _quota_from_manifest(self, manifest: client.V1ResourceQuota) -> models.Quota:
+        gpu = 0
+        gpu_kind = models.GpuKind.NVIDIA
+        for igpu_kind in models.GpuKind:
+            key = f"requests.{igpu_kind}/gpu"
+            if key in manifest.spec.hard:
+                gpu = int(manifest.spec.hard.get(key))
+                gpu_kind = igpu_kind
```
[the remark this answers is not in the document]
olevski: Yes

src/renku_crc/server_options.py
olevski: Yes you are right. I fixed it.

olevski:
@Panaetius I made all the changes you requested. Also I got the coverage back up to where it was before.

Panaetius (approved):

PR #14, opened 2023-07-18
feat: renku storage service

Panaetius:
Moves code to using poetry polylith plugin to allow separating components and reusing them between services.


olevski (commented):
The only thing that is a bit annoying is that isort considers the internal library as if it was external? Is that right? Can we fix this somehow with configuration in the change somewhere?

olevski (approved):

PR #17, opened 2023-08-09
feat: add storage url parsing

Panaetius:


components/renku_data_services/storage_models/core.py
```diff
@@ -15,17 +17,117 @@ class CloudStorage:
 
     storage_id: str | None = None
 
+    source_path: str
+    target_path: str
```
olevski, line 21: I think it would be nice make the type for the source and target path `Path` from the `pathlib` library. Or at least should we add some validation to make sure they are proper valid paths?

components/renku_data_services/storage_models/core.py
```diff
@@ -15,17 +17,117 @@ class CloudStorage:
...
             configuration=data["configuration"],
-            storage_type=data["configuration"].get("type"),
+            storage_type=data["configuration"]["type"],
+            source_path=data["source_path"],
+            target_path=data["target_path"],
         )
 
     @classmethod
-    def from_url(cls, storage_url: str, project_id: str) -> "CloudStorage":
+    def from_url(cls, storage_url: str, project_id: str, target_path: str) -> "CloudStorage":
```
olevski, line 51: Can we document which standard or where the documentation for this url format is? Or are we making it up ourselves?

components/renku_data_services/storage_schemas/core.py
```diff
@@ -91,7 +124,22 @@ class RCloneOption(BaseModel):
     value_str: str = Field(alias="ValueStr")
     type: str = Field(alias="Type")
 
-    def validate_config(self, value, provider: str):
+    def matches_provider(self, provider: str | None) -> bool:
+        """Check if this option applies for a provider."""
```
olevski, line 128:
```suggestion
        """Check if this option applies for a provider.
        
        Note that the field can contain multiple providers separated by comma and can be preceded by a '!' which flips the matching logic.
```

olevski (changes requested):
A few changes/questions. Most notably I think we should not be storing the bucket name in the source path because it is not a real thing that exists in the storage. I would rather store it separtely and be able to use the source path directly than having to remove the bucket name from it.

components/renku_data_services/storage_adapters/schemas.py
```diff
@@ -25,6 +25,8 @@ class CloudStorageORM(BaseORM):
     project_id: Mapped[str] = mapped_column("project_id", String(), index=True)
     storage_type: Mapped[str] = mapped_column("storage_type", String(20))
     configuration: Mapped[dict[str, Any]] = mapped_column("configuration", JSONVariant)
+    source_path: Mapped[str] = mapped_column("source_path", String())
+    target_path: Mapped[str] = mapped_column("target_path", String())
```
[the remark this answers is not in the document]
Panaetius: yes that's correct

components/renku_data_services/storage_schemas/api.spec.yaml
```diff
@@ -165,9 +170,18 @@ components:
               description: Dictionary of rclone key:value pairs (based on schema from '/storage_schema')
               additionalProperties:
                 type: string
+            source_path:
+              type: string
+              description: the source path to mount, usually starts with bucket/container name
+              example: bucket/my/storage/folder/
```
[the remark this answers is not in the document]
Panaetius: as discussed, rclone config does not contain bucket

components/renku_data_services/storage_models/core.py
Panaetius: `Path` doesn't really validate the code and validation fails if the folder doesn't exist (which it does not).

I've added validation in a "unify databases" follow-up PR

test/bases/renku_data_services/storage_api/test_storage.py
```diff
@@ -40,38 +42,145 @@ def storage_test_client(storage_repo: StorageRepository) -> SanicTestClient:
...
+):
     _, res = storage_test_client.post(
         "/api/storage/storage",
         headers={"Authorization": "bearer test"},
         data=json.dumps(payload),
     )
     assert res
     assert res.status_code == expected_status_code
+    if res.status_code < 300:
+        assert res.json["storage_type"] == expected_storage_type
```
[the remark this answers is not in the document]
Panaetius: I added a lot of tests in the "add tests" PR

components/renku_data_services/storage_models/core.py
Panaetius: there's no central standard, it's based on the different formats by the different providers. I'll add some documentation.

olevski (approved):

olevski (approved):

PR #18, opened 2023-08-11
feat: add storage edit endpoints

Panaetius:
adds put/patch/delete endpoints


components/renku_data_services/storage_adapters/core.py
```diff
@@ -36,10 +37,57 @@ async def get_storage(self, id: str | None = None, project_id: str | None = None
...
         return orm.dump()
+
+    async def update_storage(self, storage_id: str, **kwargs) -> models.CloudStorage:
+        """Update a cloud storage entry."""
+        async with self.session_maker() as session:
+            async with session.begin():
+                res = await session.execute(
+                    select(schemas.CloudStorageORM).where(schemas.CloudStorageORM.storage_id == storage_id)
+                )
+                storage = res.one_or_none()
```
olevski, line 67: This is really neat. Did not know about this method.

components/renku_data_services/storage_adapters/core.py
```diff
@@ -36,10 +37,57 @@ async def get_storage(self, id: str | None = None, project_id: str | None = None
...
+
+                if storage is None:
+                    raise errors.MissingResourceError(message=f"The storage with id '{storage_id}' cannot be found")
+                storage = storage[0]
+
+                for key, value in kwargs.items():
+                    setattr(storage, key, value)
+
+                if "configuration" in kwargs and "type" in kwargs["configuration"]:
+                    storage.storage_type = kwargs["configuration"]["type"]
```
olevski, line 77: So the storage type is set both in `storage.storage_type` and `storage.configuration.type`? Is there a reason for this. Can we get away with having just one and not both. And if we have to have both we should check for consistency somewhere?

olevski (commented):
One small question. All else looks good.

olevski (approved):

PR #19, opened 2023-08-11
chore: add tests

Panaetius:


bases/renku_data_services/crc_api/config.py
```diff
@@ -92,6 +92,11 @@ def __post_init__(self):
                 defaults = ServerOptionsDefaults.parse_obj(safe_load(f))
             self.default_resource_pool = generate_default_resource_pool(options, defaults)
 
+    @property
+    def repo(self):
+        """Used by alembic to find repo."""
```
olevski, line 97:
```suggestion
        """Used by alembic for database migrations to get the object that satisfies the DataRepository protocol."""
```

Also please add the type annotation for what is returned. I see that you have a `DataRepository` protocol. We should put that as the return type annotation. This change makes the change to the docstring unnecessary so you can keep the docstring short.

bases/renku_data_services/storage_api/app.py
```diff
@@ -76,6 +78,10 @@ def put(self) -> BlueprintFactoryResponse:
         @authenticate(self.authenticator)
         @only_admins
         async def _put(request: Request, storage_id: str, validator: RCloneValidator, user: base_models.User):
+            if not request.json:
+                raise errors.ValidationError(message="The request body is empty. Please provide a valid JSON object.")
+            if not isinstance(request.json, dict):
+                raise errors.ValidationError(message="The request body is not a valid JSON object.")
```
olevski, line 84: The schemas for the api spec do not enforce this?

test/components/renku_data_services/crc_api/test_config.py
```diff
@@ -0,0 +1,15 @@
+from renku_data_services.crc_api.config import Config
+from renku_data_services.users.dummy import DummyAuthenticator
+
+
+def test_config_dummy(monkeypatch):
```
olevski, line 5: Thanks for adding such a test!

Makefile
```diff
@@ -16,12 +16,22 @@ style_checks:
...
 	@rm -f .tmp.pid
+	@echo "===========================================STORAGE API==========================================="
+	DUMMY_STORES=true poetry run coverage run -a -m sanic --debug --single-process renku_data_services.storage_api.main:create_app --factory & echo $$! > .tmp.pid
+	@sleep 10
+	-poetry run st run http://localhost:8000/api/storage/spec.json --validate-schema True --checks all --hypothesis-max-examples 20 --data-generation-method all --show-errors-tracebacks --hypothesis-suppress-health-check data_too_large --max-response-time 100 -v --header "Authorization: bearer some-random-key-123456"
+	cat .tmp.pid | xargs kill
+	@rm -f .tmp.pid
+	@echo "===========================================TEST DOWNGRADE==========================================="
+	DUMMY_STORES=true poetry run coverage run -m alembic -c components/renku_data_services/migrations/alembic.ini --name=storage downgrade base
+	DUMMY_STORES=true poetry run coverage run -m alembic -c components/renku_data_services/migrations/alembic.ini --name=resource_pools downgrade base
```
olevski, line 34: That is awesome! This tests if we can roll back the mgirations?

olevski (commented):
Just one really picky change. Everything else is great. Thank you for adding tests for the config. I have been burned by that before and I still forgot to do it for the CRC.

Makefile
Panaetius: yes. Mainly to bump coverage but it does perform a full rollback back to empty

bases/renku_data_services/crc_api/config.py
Panaetius: I've addressed this in https://github.com/SwissDataScienceCenter/renku-data-services/pull/21/commits/83d4098936b03ca33e0a94ef03c02268f3eb90d0

bases/renku_data_services/storage_api/app.py
Panaetius: `**request.json` fails in that case, hypothesis found this issue.

Panaetius: And I can't `@validate()` since that doesn't work when there's multiple schemas. or at least I didn't find a way to do that

olevski (approved):

olevski (approved):

PR #22, opened 2023-08-22
fix: adding users and showing quotas

olevski:
A few issues here that @rokroskar found:
- The response from some endpoints fail because we expect to get a full quota in the response but internally in some cases the quota could also just be a string. So these cases were breaking the API. Now I have addressed this.
- Patching a resource pool was calling the method to delete the resource pool, and luckily this was failing, I have corrected this now so it should work

### DIFF
--- a/.github/workflows/test_publish.yml
+++ b/.github/workflows/test_publish.yml
@@ -7,6 +7,9 @@ on:
     tags:
       - 'v*'
   pull_request:
+    branches:
+      - 'main'
+      - 'hotfix/**'
 
 jobs:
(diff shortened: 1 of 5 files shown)


olevski:
Note that the base is different since this is will be pushed out as a bugfix release.

I will post the link to a renku CI deployment with the image from this PR.

olevski:
Here is the test deployment with the image in this PR:
https://ci-renku-3209.dev.renku.ch/swagger

Panaetius (approved):

PR #23, opened 2023-08-23
feat: add storage sensitive field handling

Panaetius:
Also fixes many smaller bugs I found


olevski (approved):

Panaetius:
I got coverage back up in https://github.com/SwissDataScienceCenter/renku-data-services/pull/29

bases/renku_data_services/data_api/main.py
```diff
@@ -31,6 +32,7 @@ async def do_migrations(*_):
         run_migrations_for_app("storage", config.rp_repo)
         config.rp_repo.initialize(config.default_resource_pool)
 
+    app.ext.add_dependency(RCloneValidator)
```
olevski, line 35: So this gets created for every request or once when the server starts? Because from looking at the RCloneValidator it seems that with this the RCloneValidator will be instantiated with every request. And it is not that nice because instantiating the object reads a file. So every request that needs the validator will wait for the file to be read.

I found this example in the Sanic docs:
```
@app.before_server_start
async def setup_db(app, _):
    db_conn = FakeConnection()
    app.ext.dependency(db_conn)
```

In this case the class instance that is injected is created once.

See here (under the `The higher level API using dependency` section): https://sanic.dev/en/plugins/sanic-ext/injection.html#injecting-services

olevski (approved):

PR #29, opened 2023-08-28
feat: add name field and proper authz

### LINKED ISSUE #26: add name field to storage
add a name field to CloudStorage entity

Panaetius:
closes #26


components/renku_data_services/base_models/core.py
```diff
@@ -29,6 +29,13 @@ def is_authenticated(self):
         return self.id is not None
 
 
+@dataclass(kw_only=True)
+class GitlabAPIUser(APIUser):
+    """The model for a user of the API for Gitlab authenticated requests."""
+
+    project_id: str
```
olevski, line 36: Making the project ID part of the GitlabAPIUser object is really confusing. A Gitlab user is made to have access to many projects. Not sure why the object makes it seem like that it is not the case.

components/renku_data_services/storage_adapters/core.py
```diff
@@ -68,7 +90,19 @@ async def update_storage(self, storage_id: str, **kwargs) -> models.CloudStorage
 
                 if storage is None:
                     raise errors.MissingResourceError(message=f"The storage with id '{storage_id}' cannot be found")
+                if storage[0].project_id != user.project_id:
```
olevski, line 93: Along with my comment about the GitlabAuthenticator it would be nice to hide the authorization complexity in the Authenticator and not see it here. The Authenticator or a decorator on an endpoint handler can raise this error if things do not match.

components/renku_data_services/base_models/core.py
olevski: ..Ok so I think I understand where the awkwardness comes from. The GitlabAuthenticator is not only doing authentication but also authorization in "one go". And the authenticator is not really validating the access token in all cases but only in some (when a project id is passed).

We should split this up so that the authenticator has two public methods - one that only authenticates and another that validates if an authenticated user can access a specific project. The second will implicitly also authenticate the user. But then it makes it clearer what is going on. Also then we do not have to tie the GitlabAPIUser to a Gitlab project. And when we need to we can validate the access for a specific gitlab user to the project.

olevski: ... I know that some of this code did not change with this PR. But now that I see how we are starting to use the Authenticator I think we should clean things up a bit.

olevski (changes requested):
I think we should add a decorator for the http handlers (like `is_admin`) to check if a user has access to a specific gitlab project. And the authenticator for Gitlab should have clearer separation between authn and authz. Lastly we should make sure we always check the validity of the provided access token for gitlab when the authenticate method is called.

olevski (dismissed):

olevski (approved):

PR #35, opened 2023-08-31
feat: add old crc migration and fix openapi schema

### LINKED ISSUE #25: properly serve schemas for crc and storage
right now they're just merged but that doesn't work, they should be properly combined or served separately

### LINKED ISSUE #28: add migration for existing CRC database
Since we now moved migrations to their own independent schema, any existing database from the CRC service has to have its data copied over and cleaned up.

Panaetius:
closes #28 
closes #25

### DIFF
--- a/components/renku_data_services/crc_migrations/versions/0044b54a89b9_migrate_old_crc_tables.py
+++ b/components/renku_data_services/crc_migrations/versions/0044b54a89b9_migrate_old_crc_tables.py
@@ -1,0 +1,95 @@
+"""migrate old crc tables
+
+Revision ID: 0044b54a89b9
+Revises: 95ce5418d4d9
+Create Date: 2023-08-31 09:53:55.488210
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = "0044b54a89b9"
+down_revision = "95ce5418d4d9"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    connection = op.get_bind()
+    result = connection.execute(
+        sa.sql.text(
+            """SELECT EXISTS (
+                   SELECT FROM information_schema.tables
+                   WHERE  table_schema = 'public'
+                   AND    table_name   = 'resource_pools'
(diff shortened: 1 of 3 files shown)


components/renku_data_services/crc_migrations/versions/0044b54a89b9_migrate_old_crc_tables.py
```diff
@@ -0,0 +1,95 @@
...
+    op.drop_table("users", schema="public")
+    op.drop_index(op.f("ix_resource_pools_resource_pools_quota"), table_name="resource_pools", schema="public")
+    op.drop_index(op.f("ix_resource_pools_resource_pools_public"), table_name="resource_pools", schema="public")
+    op.drop_index(op.f("ix_resource_pools_resource_pools_name"), table_name="resource_pools", schema="public")
+    op.drop_index(op.f("ix_resource_pools_resource_pools_default"), table_name="resource_pools", schema="public")
+    op.drop_table("resource_pools", schema="public")
+
+
+def downgrade() -> None:
+    pass
```
olevski, line 95: I think we should raise a `NotImplementedError` or something like that here to let people know that the downgrade did not work. Otherwise I think that with a `pass` alembic will consider that the downgrade succeeded.

olevski (changes requested):

components/renku_data_services/crc_migrations/versions/0044b54a89b9_migrate_old_crc_tables.py
Panaetius: but we have a test that downgrades work and that would fail then.

We could of course manually implement a proper downgrade function that puts it back where it came from, but I'm not sure that's valuable. Especially since data in a crc deployment could have come from the crc schema or from the public/default schema and there's no way to tell on downgrade where it came from.

I think a nop is the right choice here, since this isn't really a migration in the alembic sense anyways, it's just getting in data from somewhere else, and just done as a migration for convenience.

olevski (approved):

olevski (approved):

PR #44, opened 2023-09-10
feat: add tolerations and node affinities

olevski:
closes https://github.com/SwissDataScienceCenter/renku-notebooks/issues/1582
closes https://github.com/SwissDataScienceCenter/renku-notebooks/issues/1583


components/renku_data_services/resource_pool_adapters/adapter.py
```diff
@@ -129,7 +129,18 @@ async def get_resource_pools(
     ) -> List[models.ResourcePool]:
         """Get resource pools from database."""
         async with self.session_maker() as session:
-            stmt = select(schemas.ResourcePoolORM).options(selectinload(schemas.ResourcePoolORM.classes))
+            stmt = select(schemas.ResourcePoolORM).options(
+                selectinload(
+                    schemas.ResourcePoolORM.classes,
+                ).selectinload(
```
Panaetius, line 135: Should we just define that affinities/tolerations are loaded by default on the ORM class level. if we always load it anyways it'd clean up the code in this file a bit.

I.e. `... = relationship(..., lazy="selectin")`

components/renku_data_services/resource_pool_adapters/adapter.py
```diff
@@ -309,23 +342,16 @@ async def update_resource_pool(self, api_user: base_models.APIUser, id: int, **k
                                         message="More fields than the id of the class "
                                         "should be provided when updating it"
                                     )
-                                stmt_cls = (
-                                    update(schemas.ResourceClassORM)
-                                    .where(schemas.ResourceClassORM.id == class_id)
-                                    .where(schemas.ResourceClassORM.resource_pool_id == id)
-                                    .values(**cls)
-                                    .returning(schemas.ResourceClassORM)
+                                new_cls = await self.update_resource_class(
```
Panaetius, line 345: This seems like a nice place to use `asyncio.gather()` outside the loop without awaiting here

Panaetius (changes requested):

components/renku_data_services/resource_pool_adapters/adapter.py
olevski: Done.

components/renku_data_services/resource_pool_adapters/adapter.py
olevski: Yes this cleaned up a lot of crap. Thank you!

Panaetius (approved):

PR #46, opened 2023-09-13
feat: allow patching name and always return private fields

Panaetius:


leafty:
`components/renku_data_services/storage_blueprints/core.py` needs to be updated to return `sensitive_fields`

Panaetius:
@leafty I addressed that in https://github.com/SwissDataScienceCenter/renku-data-services/pull/49

olevski (approved):

PR #53, opened 2023-10-05
fix gitlab graphql query

Panaetius:


components/renku_data_services/base_models/core.py
```diff
@@ -51,12 +51,24 @@ class GitlabAPIUser(APIUser):
...
 
+    @property
+    def gitlab_graphql_url(self):
+        """Gets the graphql url for gitlab."""
+        gitlab_url = self.gitlab_url
+
+        if not gitlab_url.startswith("http") and "://" not in gitlab_url:
+            gitlab_url = f"https://{gitlab_url}"
+
+        gitlab_url = gitlab_url.rstrip("/")
```
olevski, line 62: I think it is better to check whether gitlab_url is a valid/proper url in a `__post_init__` method of thi dataclass and crash if you dont like what you see. You can also run the `.rstrip` to remove trailing slashes there. Then you can just freely use the value in other methods/places.

olevski (changes requested):
One minor change. All else is good.

components/renku_data_services/base_models/core.py
```diff
@@ -51,17 +51,16 @@ class GitlabAPIUser(APIUser):
...
         gitlab_url = self.gitlab_url
 
         if not gitlab_url.startswith("http") and "://" not in gitlab_url:
-            gitlab_url = f"https://{gitlab_url}"
+            raise errors.ConfigurationError(message=f"Gitlab URL should start with 'http(s)://', got: {gitlab_url}")
 
         gitlab_url = gitlab_url.rstrip("/")
 
-        return f"{gitlab_url}/api/graphql"
+        self.gitlab_url = f"{gitlab_url}/api/graphql"
```
olevski, line 63: I think we should be more specific here. This should be called `self.gitlab_graphql_url`. As a user of this class without reading the code I would expect that `self.gitlab_url` is the base url of gitlab. And that is what it used to be prior to this change. So this could lead to confusion and problems down the line.

olevski (changes requested):

olevski (approved):

PR #54, opened 2023-10-11
fix: use custom certificates when making requests

### LINKED ISSUE #43: Enable/check SSL certificates
Self-signed SSL certificates for BIT for the data services is not working.

Panaetius:
closes #43 

I hardcoded the path since it's hardcoded in other services as well, but we really don't have a good solution for passing a setting to a cross-cutting concern like this and I don't see an easy way to address this in the current setup.


components/renku_data_services/utils/core.py
```diff
@@ -0,0 +1,9 @@
+"""Shared utility functions."""
+import ssl
+
+
+def get_ssl_context():
+    """Get an SSL context supporting mounted custom certificates."""
+    context = ssl.create_default_context()
+    context.load_verify_locations(capath="/usr/local/share/ca-certificates")
```
olevski, line 8:
```suggestion
    if os.environ.get("SSL_CERT_FILE", None"):
        context.load_verify_locations(cafile=os.environ["SSL_CERT_FILE"])
```

olevski: Based on this: https://github.com/SwissDataScienceCenter/renku/blob/master/helm-chart/renku/templates/_certificates-env.tpl

components/renku_data_services/utils/core.py
```diff
@@ -0,0 +1,9 @@
+"""Shared utility functions."""
+import ssl
```
olevski, line 2:
```suggestion
import os
import ssl

```

components/renku_data_services/utils/core.py
```diff
@@ -0,0 +1,9 @@
...
+
+
+def get_ssl_context():
```
olevski, line 5: Also lets toss in a functools.lru_cache on the `get_ssl_context` function. Because we really dont need this to execute every single time. I just worry about it because now we will open, read and parse a file before every http request sent with httpx.

olevski (changes requested):

components/renku_data_services/utils/core.py
Panaetius: Is this actually correct? Going by https://github.com/SwissDataScienceCenter/renku/blob/master/helm-chart/renku/templates/_certificates-volume-mounts.tpl#L3 and https://github.com/SwissDataScienceCenter/renku/blob/master/helm-chart/renku/templates/_certificates-volumes.tpl#L11 it could be multiple and the env should just be /etc/ssl/certs/, no?

olevski: @Panaetius the `"SSL_CERT_FILE"` contains the combination of all SSL certs available in `/etc/ssl/certs/` as well as extra stuff that is added via `/usr/local/share/etc/certs`.

olevski: For python we only inject the environment variable that points to the cafile that contains all certs. If you want you can also set capath as well. And you can set the capth to always be `/etc/ssl/certs` - regardless of what the evnrionment variable is and that should work. But I think that as you have things right now we will only load any custom/additional certificates from `/usr/local/share/ca-certificates` and completely ignore the system stuff at `etc/ssl/certs`

olevski (dismissed):

olevski (approved):

PR #59, opened 2023-10-19
fix: run sanic in single process

olevski:
The defaults we had before were weird. also we were running with `--fast` which takes up way too many replicas of itself because it decides this based on the total number of cpus and when it runs on k8s it sees the total cpus on the node not the total cpus allocated to the pod.

### DIFF
--- a/projects/renku_data_service/Dockerfile
+++ b/projects/renku_data_service/Dockerfile
@@ -33,5 +33,5 @@ RUN apt-get update && apt-get install -y \
 USER $USER_UID:$USER_GID
 WORKDIR /app
 COPY --from=builder /app/env ./env
-ENTRYPOINT ["tini", "-g", "--"]
-CMD ["tini", "-g", "--", "env/bin/python", "-m", "renku_data_services.data_api.main", "--fast"]
+ENTRYPOINT ["tini", "-g", "--", "env/bin/python", "-m", "renku_data_services.data_api.main"]
+CMD ["--single-process"]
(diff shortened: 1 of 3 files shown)


projects/renku_data_service/Dockerfile
```diff
@@ -33,5 +33,5 @@ RUN apt-get update && apt-get install -y \
 USER $USER_UID:$USER_GID
 WORKDIR /app
 COPY --from=builder /app/env ./env
-ENTRYPOINT ["tini", "-g", "--"]
-CMD ["tini", "-g", "--", "env/bin/python", "-m", "renku_data_services.data_api.main", "--fast"]
+ENTRYPOINT ["tini", "-g", "--", "env/bin/python", "-m", "renku_data_services.data_api.main"]
+CMD ["--single-process"]
```
Panaetius, line 37: I think we should just run without an argument:
> By default, Sanic runs a main process and a single worker process (see [worker manager](https://sanic.dev/en/guide/deployment/manager.html) for more details).
from [here](https://sanic.dev/en/guide/deployment/running.html#workers)

Because on single process mode, it says:
> ou will also not have access to any features that require processes (auto-reload, the inspector, etc).
and I don't know if I'm comfortable with not having "etc."...

olevski: Oh cool I did not know that. I will remove that. I did not know about these limitations.

Panaetius (approved):

Panaetius (approved):

olevski:
Just for reference, results of some very simple k6 load tests...

I used this configuration

```json
scenarios: {
    lecture: {
      executor: "per-vu-iterations",
      vus: 50,
      iterations: 10,
    },
  },
```

And I only listed all resource pools as an anonymous user, sleeping for 1 second between each iteration.

The memory consumption stayed pretty constant at ~350MB, the cpu usage spiked from almost zero at rest to about 150m before settling down.

PR #66, opened 2023-10-28
fix: properly validate quotas

olevski:
This fixes several instances in the code where:
- a quota can be added in the db even if it is not compatible with existing resource classes
- a class can be added if it is not compatible with an existing quota
- the quota field in the ResourcePool dataclass has be either of type Quota or None, prior to this it could also be a string that represented the name of the quota in k8s, but that complicates the code quite a bit - the Quota type contains the k8s name as its id
- fixes to the tests to work with the new changes
- changes to the tests so that the config is used when certain things that it initializes are needed - prior to this we initialized certain db adapeters or clients separately but the config type can do these even for a test scenario

/deploy


mohammad-alisafaee (approved):
Great work, thank you! I've just made a single comment. Feel free to ignore.

PR #68, opened 2023-10-28
fix: creating users

olevski:
Fixes #67 

/deploy

PR #70, opened 2023-10-28
fix: migrate primary key resource_pool sequences

### LINKED ISSUE #67: User database consistency
When trying to add a user to a pool the following can happen:

```
sqlalchemy.dialects.postgresql.asyncpg.AsyncAdapt_asyncpg_dbapi.IntegrityError: <class 'asyncpg.exceptions.UniqueViolationError'>: duplicate key value violates unique constraint "users_pkey"
DETAIL:  Key (id)=(1) already exists.
```

```
  File "/app/env/lib/python3.11/site-packages/renku_data_services/crc/blueprints.py", line 466, in _post
    kc_user = await self.repo.insert_user(api_user=user, user=base_models.User(keycloak_id=body.id))
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/env/lib/python3.11/site-packages/renku_data_services/crc/db.py", line 104, in decorated_function
    response = await f(self, *args, **kwargs)
  File "/app/env/lib/python3.11/site-packages/renku_data_services/crc/db.py", line 447, in insert_user
    async with session.begin():
```

olevski:
Fixes #67

### DIFF
--- a/components/renku_data_services/crc_migrations/versions/5403953f654f_fix_primary_key_sequences.py
+++ b/components/renku_data_services/crc_migrations/versions/5403953f654f_fix_primary_key_sequences.py
@@ -1,0 +1,42 @@
+"""fix primary key sequences
+
+Revision ID: 5403953f654f
+Revises: 024b215f9a14
+Create Date: 2023-10-28 13:23:45.947375
+
+"""
+import sqlalchemy as sa
+from alembic import op
+from psycopg import sql
+
+# revision identifiers, used by Alembic.
+revision = "5403953f654f"
+down_revision = "024b215f9a14"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    connection = op.get_bind()
+    inspector = sa.inspect(connection)
+    schema = "resource_pools"
+    tables = inspector.get_table_names(schema)
+
+    for table in tables:
+        if table in ["alembic_version", "resource_pools_users"]:
+            continue
+        full_name = sql.Identifier(schema, table)
+        statement = sql.SQL("SELECT MAX(id) FROM {}").format(full_name).as_string(connection)
+        res = connection.exec_driver_sql(statement)
+        row = res.fetchone()
+        if row is not None and len(row) > 0 and row[0] is not None:
+            last_id = row[0]
+            sequence_name = sql.Identifier("resource_pools", f"{table}_id_seq")
+            statement = (sql.SQL("ALTER SEQUENCE {} RESTART WITH ").format(sequence_name).as_string(connection)) + str(
+                last_id
+            )  # Could not figure out how to add this to a query with Alembic
+            connection.exec_driver_sql(statement)
+
+
+def downgrade() -> None:
+    pass
--- a/components/renku_data_services/crc_migrations/versions/5403953f654f_fix_primary_key_sequences.py
+++ b/components/renku_data_services/crc_migrations/versions/5403953f654f_fix_primary_key_sequences.py
@@ -33,7 +33,7 @@ def upgrade() -> None:
             last_id = row[0]
             sequence_name = sql.Identifier("resource_pools", f"{table}_id_seq")
             statement = (sql.SQL("ALTER SEQUENCE {} RESTART WITH ").format(sequence_name).as_string(connection)) + str(
-                last_id
+                last_id + 1
             )  # Could not figure out how to add this to a query with Alembic
             connection.exec_driver_sql(statement)
 


olevski:
replaces #68

aledegano (dismissed):

components/renku_data_services/crc_migrations/versions/5403953f654f_fix_primary_key_sequences.py
```diff
@@ -0,0 +1,42 @@
...
+from alembic import op
+from psycopg import sql
+
+# revision identifiers, used by Alembic.
+revision = "5403953f654f"
+down_revision = "024b215f9a14"
+branch_labels = None
+depends_on = None
+
+
```
mohammad-alisafaee, line 18: I'm not sure why this error is happening. Shouldn't a sequence field be unique by default? If no, does it mean that we need to re-run this migration in future again?

components/renku_data_services/crc_migrations/versions/5403953f654f_fix_primary_key_sequences.py
```diff
@@ -0,0 +1,42 @@
...
+            continue
+        full_name = sql.Identifier(schema, table)
+        statement = sql.SQL("SELECT MAX(id) FROM {}").format(full_name).as_string(connection)
+        res = connection.exec_driver_sql(statement)
+        row = res.fetchone()
+        if row is not None and len(row) > 0 and row[0] is not None:
+            last_id = row[0]
+            sequence_name = sql.Identifier("resource_pools", f"{table}_id_seq")
+            statement = (sql.SQL("ALTER SEQUENCE {} RESTART WITH ").format(sequence_name).as_string(connection)) + str(
+                last_id
```
mohammad-alisafaee, line 36: Shouldn't we increase `last_id` to `last_id + 1`?

mohammad-alisafaee (changes requested):

components/renku_data_services/crc_migrations/versions/5403953f654f_fix_primary_key_sequences.py
olevski: No because the `last_id` is the last in-use id. Postgres increments it and then issues a new id.

So on a brand new deployment with nothing ever inserted in it, this is what you get:

```
renku=> select * from resource_pools.tolerations_id_seq;
 last_value | log_cnt | is_called
------------+---------+-----------
          1 |       0 | f
(1 row)

renku=> select * from resource_pools.tolerations;
 key | resource_class_id | id
-----+-------------------+----
(0 rows)

```

There is nothing in the table but the last_id is 1 and not 0

olevski: I think you are right. Let me check on a brand new deployment.

components/renku_data_services/crc_migrations/versions/5403953f654f_fix_primary_key_sequences.py
olevski: It is unique and that is the failure. The problem occurs because we did this migration: https://github.com/SwissDataScienceCenter/renku-data-services/blob/main/components/renku_data_services/crc_migrations/versions/0044b54a89b9_migrate_old_crc_tables.py

Here we moved the tables and indices from the public schema to a specific one (called resource_pools) but when we did this we did not move the primary key sequences. And because we did not move these then postgres just automatically created them and it started from 1 rather.

components/renku_data_services/crc_migrations/versions/5403953f654f_fix_primary_key_sequences.py
olevski: I think it is just `last_id`.

When the table is empty it is at 1. But when I add more things it goes to the last value being used:

```
renku=> select * from resource_pools.users;
             keycloak_id              | id | no_default_access
--------------------------------------+----+-------------------
 9d7f30b2-55cb-4a2a-a4ee-b574d44888ef |  1 | f
(1 row)

renku=> select * from resource_pools.users_id_seq;
 last_value | log_cnt | is_called
------------+---------+-----------
          1 |      32 | t
(1 row)
```

Then I add a user in the API

```
renku=> select * from resource_pools.users;
             keycloak_id              | id | no_default_access
--------------------------------------+----+-------------------
 9d7f30b2-55cb-4a2a-a4ee-b574d44888ef |  1 | f
 f088eac3-5703-4f40-988a-6447b851500b |  2 | f
(2 rows)

renku=> select * from resource_pools.users_id_seq;
 last_value | log_cnt | is_called
------------+---------+-----------
          2 |      31 | t
(1 row)
```

olevski: If you do the same thing on renkulab.io you will see that there are ~80 or something users in the users table but the `resource_pools.users_id_seq` has a last_value of 1. Now it is probably at something higher beacuse with every failed request to create a user postgres still increments the sequence.

mohammad-alisafaee: I did this simple example:
```
books=# CREATE TABLE person (name VARCHAR NOT NULL, id INT NOT NULL);
CREATE TABLE
books=# CREATE SEQUENCE person_id START 1 INCREMENT 1 OWNED BY person.id;
CREATE SEQUENCE
books=# INSERT INTO person(name, id) VALUES ('A', NEXTVAL('person_id')), ('B', NEXTVAL('person_id'));
INSERT 0 2
books=# SELECT * FROM person;
 name | id 
------+----
 A    |  1
 B    |  2
(2 rows)

books=# SELECT MAX(id) FROM person;
 max 
-----
   2
(1 row)

books=# ALTER SEQUENCE person_id RESTART WITH 2;
ALTER SEQUENCE
books=# SELECT NEXTVAL('person_id');
 nextval 
---------
       2
(1 row)

```
As you can see, when we restart the sequence with maximum `id`'s value, the next sequence still returns that max id value which should cause a conflict again.

I couldn't find how we use the sequence in the code but if we use `NEXTVAL` it returns the restart value in the first time.

olevski: Ok honestly it does not matter. In the worst case we skip a number in the ids.

olevski: I made it `last_id + 1`

mohammad-alisafaee (approved):
Thanks for your explanations!

PR #73, opened 2023-10-30
fix: add user patch and put endpoints

olevski:
This is needed so that admins can actually modify users to have or not have access to the default resource pool.

We forgot to add these endpoints when we were adding the initial functionality.


olevski:
@mohammad-alisafaee this is ready now

mohammad-alisafaee (approved):
LGTM, thank you!

PR #74, opened 2023-10-31
feat: add user preferences service

### LINKED ISSUE #64: Create a backend and API to store and retrieve user preferences
Part of https://github.com/SwissDataScienceCenter/renku/issues/3325.

Add new models and APIs to store and retrieve pinned projects.

leafty:
Fixes #64.

Note: merging into a feature branch.


olevski:
Is this ready for review @leafty?

leafty:
> Is this ready for review @leafty?

Yes it is 😄

components/renku_data_services/user_preferences/db.py
```diff
@@ -0,0 +1,124 @@
...
+                    if project_slug.lower() == slug.lower():
+                        return user_preferences.dump()
+
+                new_project_slugs = list(project_slugs) + [project_slug]
+                pinned_projects = models.PinnedProjects(project_slugs=new_project_slugs).model_dump()
+                user_preferences.pinned_projects = pinned_projects
+                return user_preferences.dump()
+
+    async def remove_pinned_project(self, user: base_models.APIUser, project_slug: str) -> models.UserPreferences:
+        """Adds a pinned project from the user's preferences."""
```
olevski, line 105:
```suggestion
        """Removes a pinned project from the user's preferences."""
```

components/renku_data_services/user_preferences/orm.py
```diff
@@ -0,0 +1,44 @@
...
+class UserPreferencesORM(BaseORM):
+    """Stored user preferences."""
+
+    __tablename__ = "user_preferences"
+
+    user_id: Mapped[str] = mapped_column("user_id", String(), primary_key=True)
+    """Id of the user."""
+
+    pinned_projects: Mapped[dict[str, Any]] = mapped_column("pinned_projects", JSONVariant)
+    """Pinned projects."""
```
olevski, line 30: Just add a regular int-based `ID` key as the primary key. Make the `user_id` field an index and make it unique as well.

olevski: Also shouldn't pinned_projects be `Mapped[List[str]]`? I assume this will always be a list of strings in json format.

olevski (changes requested):

components/renku_data_services/user_preferences/api.spec.yaml
```diff
@@ -0,0 +1,142 @@
...
+      properties:
+        project_slugs:
+          type: array
+          items:
+            $ref: "#/components/schemas/ProjectSlug"
+    ProjectSlug:
+      type: string
+      description: The slug used to identify a project
+      minLength: 3
+      example: "user/my-project"
```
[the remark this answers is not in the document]
leafty: This would be over-validation. I do not care if the project slug is valid or not, GitLab will give me the project data or not (in which case, the pin is ignored).

components/renku_data_services/user_preferences/orm.py
leafty: > Just add a regular int-based `ID` key as the primary key. Make the `user_id` field an index and make it unique as well.

Is there a reason for this? I personally think int sequences as primary key are bad because they expose too much data to attackers.

Panaetius: storage already uses ULID and we should use that everywhere

.github/workflows/test_publish.yml
```diff
@@ -5,11 +5,12 @@ on:
...
     tags:
-      - 'v*'
+      - "v*"
   pull_request:
     branches:
-      - 'main'
-      - 'hotfix/**'
+      - "main"
+      - "hotfix/**"
+      - "feat-add-user-preferences"
```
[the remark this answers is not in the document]
leafty: Done in [1bcf908](https://github.com/SwissDataScienceCenter/renku-data-services/pull/74/commits/1bcf908780377daff1bec7c3d2b7323313836fbf).

components/renku_data_services/user_preferences/blueprints.py
```diff
@@ -0,0 +1,53 @@
...
+        @authenticate(self.authenticator)
+        async def _get(_: Request, user: base_models.APIUser):
+            user_preferences: models.UserPreferences | None
+            user_preferences = await self.user_preferences_repo.get_user_preferences(user=user)
+            return json(apispec.UserPreferences.model_validate(user_preferences).model_dump())
+
+        return "/user_preferences", ["GET"], _get
+
+    def post_pinned_projects(self) -> BlueprintFactoryResponse:
+        """Add a pinned project to user preferences for the logged in user."""
```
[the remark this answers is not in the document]
leafty: This will be added in another PR.

components/renku_data_services/user_preferences/api.spec.yaml
```diff
@@ -0,0 +1,142 @@
...
+                $ref: "#/components/schemas/UserPreferences"
+        default:
+          $ref: "#/components/responses/Error"
+      tags:
+        - user_preferences
+    delete:
+      summary: Remove a pinned project
+      parameters:
+        - in: query
+          name: project_slug
```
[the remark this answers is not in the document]
leafty: Done in c60f16d and d7c84eb.

components/renku_data_services/user_preferences/orm.py
olevski: I dont think you expose any IDs anywhere @leafty. And I agree that we should not use the database IDs in the API.

My comment was to keep the `user_id` and if needed expose that in your API. But, you should not use that as the primary key in the table. Just add a simple `id` field that postgres can autoincrement and and that we will rarely touch directly - but postgres will be happy. And if we have to change the `user_id` field in some way we dont have to worry about dropping the primary key in the table or messing with it and messing up the whole table as a result.

components/renku_data_services/user_preferences/api.spec.yaml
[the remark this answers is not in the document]
olevski: Ok you have to handle errors or validation somewhere. I guess that will happen on the UI side.

components/renku_data_services/user_preferences/orm.py
leafty: > But, you should not use that as the primary key in the table.

Is there a compelling argument here? These are unique within Renku and user preferences will have a 1:1 mapping to user instances.

olevski (approved):

PR #75, opened 2023-11-01
fix(crc): listing users

olevski:
More bug fixes.

/deploy


rokroskar (approved):
Thanks @olevski - I tested the API in the CI deployment and the bug seems to have been fixed.

mohammad-alisafaee (approved):
Thank you!

PR #76, opened 2023-11-02
feat: renku native projects

mohammad-alisafaee:


components/renku_data_services/project/api.spec.yaml
```diff
@@ -0,0 +1,382 @@
...
+        visibility:
+          $ref: "#/components/schemas/Visibility"
+        members:
+          $ref: "#/components/schemas/MembersWithRoles"
+      required:
+        - "id"
+        - "name"
+        - "slug"
+        - "created_by"
+        - "visibility"
```
olevski, line 188: These two should not be requried. Visibility can default to `private`.

For created by it should just not be an option at all to change. You will fill this in from the access token of the request. This way I can create a project logged in as myself but put Sekhar as the creator.

components/renku_data_services/project/api.spec.yaml
```diff
@@ -0,0 +1,382 @@
...
+          $ref: "#/components/schemas/Name"
+        slug:
+          $ref: "#/components/schemas/Slug"
+        created_by:
+          $ref: "#/components/schemas/User"
+        repositories:
+          $ref: "#/components/schemas/RepositoriesList"
+        visibility:
+          $ref: "#/components/schemas/Visibility"
+        members:
```
olevski, line 211: As discussed the members should not be shown here.

components/renku_data_services/project/api.spec.yaml
```diff
@@ -0,0 +1,382 @@
...
+        - role
+    Role:
+      # TODO: Can have multiple owners? Owner here vs ``created_by``?
+      # TODO: Better names for roles: Maintainer, admin, member, ... instead of reader and writer
+      description: Possible roles of users in a project
+      type: string
+      enum:
+        - "reader"
+        - "writer"
+        - "owner"
```
olevski, line 344: Lets just follow what the notion page says. We have `owner`, `member` and `admin` roles. 

It is this secition in notion:
image


The `created_by` field is the id of the user who created the project. That should not be changeable. And it could be that the original creator is not in the project members, owners, etc at all anymore.

components/renku_data_services/project/api.spec.yaml
```diff
@@ -0,0 +1,382 @@
...
+        name: "Renku R Project"
+        slug: "r-project"
+        created_by:
+          id: "owner-KC-id"
+        visibility: "public"
+    ProjectPost:
+      type: object
+      description: Project metadata to be created in Renku
+      additionalProperties: false
+      properties:
```
olevski, line 200: We are missing the project description field here.

components/renku_data_services/project/api.spec.yaml
```diff
@@ -0,0 +1,382 @@
...
+        name:
+          $ref: "#/components/schemas/Name"
+        slug:
+          $ref: "#/components/schemas/Slug"
+        created_by:
```
olevski, line 205: I would just put the keycloak user id of the user that created the project here, rather than full set of info about the user. Again an anonymous user who lists a public project should not get a full set of information about the project owner.

olevski (changes requested):

components/renku_data_services/project/api.spec.yaml
ciyer: I agree with the above and also think that "id" should not be a parameter -- the server should generate it.

components/renku_data_services/project/api.spec.yaml
```diff
@@ -0,0 +1,382 @@
...
+          $ref: "#/components/schemas/User"
+        repositories:
+          $ref: "#/components/schemas/RepositoriesList"
+        visibility:
+          $ref: "#/components/schemas/Visibility"
+        members:
+          $ref: "#/components/schemas/MembersWithRoles"
+      required:
+        - "id"
```
ciyer, line 184: I don't think this should be provided by a client. I think only "name" **really** needs to be required (slug could be generated from the name), but if prefer to get the slug from the client and just validate it, that is also fine with me.

components/renku_data_services/project/api.spec.yaml
```diff
@@ -0,0 +1,382 @@
...
+      type: object
+      properties:
+        version:
+          type: string
+      required:
+        - "version"
+
+    ErrorResponse:
+      type: object
+      properties:
```
ciyer, line 356: What do you think about making this an array field, `errors` which contains one or more `Error` objects, typed as specified here?

olevski: Sekhar that is an error field that is used throughout the whole API and changing its format will require significant changes to the centralized error handling. Also this error is already returned by other parts of the API and used throughout the UI and other components. I would rather keep this uniform and avoid changing this at this stage.

If we want to change the errors returned we should do it across the whole API all in one go along with the changes needed in the UI and any other components.

components/renku_data_services/project/api.spec.yaml
ciyer: ☝️

components/renku_data_services/project/api.spec.yaml
mohammad-alisafaee: This is `Project` schema which is always used in responses. These fields should be required for a Renku project.

I believe your comments refers to `ProjectPatch` and `ProjectPut` schemas that mark some of these fields as required. I fixed those two.

components/renku_data_services/project/api.spec.yaml
mohammad-alisafaee: I thought UI is generating a slug from name since it is showing that slug to users when they edit the name. Since UI will send us a slug, I've marked it as required here.
I'll make it optional in case other clients access the API.

components/renku_data_services/project/api.spec.yaml
mohammad-alisafaee: As discussed, we'll keep this as it is since `User` only includes KeyCloak ID. The other thing is that we can send different responses based on user being logged-in or not.

components/renku_data_services/project/api.spec.yaml
```diff
@@ -0,0 +1,471 @@
...
+      type: array
+      items:
+        $ref: "#/components/schemas/User"
+      minItems: 0
+      uniqueItems: true
+    MembersWithRoles:
+      description: List of users and their access level to the project
+      type: array
+      items:
+        $ref: "#/components/schemas/MemberWithRole"
```
ciyer, line 410:
```suggestion
        $ref: "#/components/schemas/MemberWithRoles"
```

components/renku_data_services/project/api.spec.yaml
```diff
@@ -0,0 +1,471 @@
...
+          name: project_id
+          required: true
+          schema:
+            type: string
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/Members"
```
ciyer, line 219: Delete requests should not have a request body.

components/renku_data_services/project/api.spec.yaml
```diff
@@ -0,0 +1,471 @@
...
+  /projects:
+    get:
+      summary: Get all projects
+      responses:
+        "200":
+          description: The project
+          content:
+            "application/json":
+              schema:
+                $ref: "#/components/schemas/Project"
```
ciyer, line 22: Is there handling of paging of results here? If so, I do not follow how that works.

components/renku_data_services/project/api.spec.yaml
```diff
@@ -0,0 +1,466 @@
...
+        $ref: "#/components/schemas/GitUrl"
+      minItems: 0
+      uniqueItems: true
+      example:
+        - https://github.com/SwissDataScienceCenter/project-1.git
+        - git@github.com:SwissDataScienceCenter/project-2.git
+    GitUrl:
+      description: URL of a Git repository
+      type: string
+      format: uri  # TODO: This should support other formats like ``git@`` and ``https+git``
```
ciyer, line 393: As discussed in the meeting today, we **might** need a field in addition to the URI, which is backend type, which would be an enum of the backends we support (just "gitlab" at first, but later perhaps "github", "bitbucket", etc).

ciyer: Discussed in today's meeting, and we decided that we are not sure that we need this, so let us not make this change now. If it later turns out to be necessary, it should be easy to make the change at that time.

olevski (changes requested):
Lets discuss the pagination implementation further... Also perhaps we can keep going as things are right now and then improve things down the line.

components/renku_data_services/project/db.py
```diff
@@ -0,0 +1,123 @@
...
+            projects_orm = result.scalars().all()
+
+            stmt = select(func.count()).select_from(schemas.ProjectORM)
+            result = await session.execute(stmt)
+            n_total_elements = result.scalar()
+            total_pages, reminder = divmod(n_total_elements, per_page)
+            if reminder:
+                total_pages += 1
+
+            pagination = PaginationResponse(page, per_page, n_total_elements, total_pages)
```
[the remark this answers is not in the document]
ciyer: Yes, there are race conditions with pagination, but I do not think there is much we can do about that. I do not think this should be a big deal in practice.

components/renku_data_services/project/orm.py
```diff
@@ -0,0 +1,60 @@
...
+    __tablename__ = "projects"
+
+    id: Mapped[str] = mapped_column("id", String(26), primary_key=True, default_factory=lambda: str(ULID()), init=False)
+    name: Mapped[str] = mapped_column("name", String(99))
+    slug: Mapped[str] = mapped_column("slug", String(99))
+    visibility: Mapped[Visibility]
+    created_by_id: Mapped[str] = mapped_column("created_by_id", String())
+    creation_date: Mapped[datetime] = mapped_column("creation_date", DateTime(timezone=True))
+    repositories: Mapped[List[str]] = mapped_column("repositories", ARRAY(String))
+    description: Mapped[str] = mapped_column("description", String(5000))
```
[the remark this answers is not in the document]
ciyer: I agree. I have this restricted to 500 in the UI right now. Descriptions are UTF-8 encoded, right?

components/renku_data_services/project/db.py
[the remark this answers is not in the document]
olevski: We can change the implementation to use a project id as the last or first item in the page - rather than only just using offsets. Then offsetting based on this constant IDs would prevent some of these issues. But I think we can let this be for now and come back to this later if we see it becoming a problem.

components/renku_data_services/project/orm.py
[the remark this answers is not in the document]
mohammad-alisafaee: I set this based on a value of `README.md` file in a newly-created renku project which is about 1500 characters. I'll change it to 500.

components/renku_data_services/project/api.spec.yaml
```diff
@@ -0,0 +1,470 @@
...
+            role: "owner"
+          - id: "another-keycloak-user-id"
+            role: "member"
+    ProjectPost:
+      description: Project metadata to be created in Renku
+      type: object
+      additionalProperties: false
+      properties:
+        name:
+          $ref: "#/components/schemas/Name"  # TODO: Should name be unique?
```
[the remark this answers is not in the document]
mohammad-alisafaee: I'll add these to the notion page and we can discuss them on the next sync meeting.

components/renku_data_services/project/orm.py
```diff
@@ -0,0 +1,60 @@
...
+
+    __tablename__ = "projects"
+
+    id: Mapped[str] = mapped_column("id", String(26), primary_key=True, default_factory=lambda: str(ULID()), init=False)
+    name: Mapped[str] = mapped_column("name", String(99))
+    slug: Mapped[str] = mapped_column("slug", String(99))
+    visibility: Mapped[Visibility]
+    created_by_id: Mapped[str] = mapped_column("created_by_id", String())
+    creation_date: Mapped[datetime] = mapped_column("creation_date", DateTime(timezone=True))
+    repositories: Mapped[List[str]] = mapped_column("repositories", ARRAY(String))
```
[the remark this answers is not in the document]
mohammad-alisafaee: As discussed, let's decide on this later before the pitch is over.

olevski (approved):

PR #77, opened 2023-11-02
chore: prevent duplicate list entries on schema merge

### LINKED ISSUE #61: The openapi spec is not valid
We have two open api specs and then we merge them into 1 file.

But when we do this any arrays in the spec get merged. 

So we get duplicate array items for components like `ErrorResponse` which are used in both the CRC and the storage service. The problem occurs in the `required` field which is a list of field names that are required and cannot have duplicates.

Because of this the automated schemathesis tests do not run because the spec is not valid.

image

rokroskar:
This makes sure the merged merged API schema complies with OpenAPI spec.

closes #61 

/deploy

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -11,7 +11,7 @@
 		"ghcr.io/devcontainers/features/git:1": {},
 		"ghcr.io/jungaretti/features/make:1": {},
 		"ghcr.io/rocker-org/devcontainer-features/apt-packages:1": {
-			"packages": "procps"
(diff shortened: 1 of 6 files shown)


leafty:
Note: would probably be nice to split in utils like I did here: https://github.com/SwissDataScienceCenter/renku-data-services/pull/74/files#diff-93648cabdae9e8c50ee0d7afe159dbaba59a9e1417ae48a21cce3e74d1cc1d6b

olevski (approved):
Thank you for fixing this @rokroskar and @leafty !

PR #80, opened 2023-11-03
chore: add swagger to devcontainer

rokroskar:
Adds swagger for development.

~~**warning**: sets up CORS in a permissive way, not sure if this is desired... but swagger needs the cors headers to work~~

CORS is enabled via an environment variable and off by default.


pyproject.toml
```diff
@@ -137,6 +137,11 @@ mypy_path = ["components", "bases"]
 files = ["bases/**/*.py", "components/**/*.py"]
 namespace_packages = true
 explicit_package_bases = true
+exclude = [
+  "components/renku_data_services/crc_migrations/versions",
```
leafty, line 141: personally, I feel it's a bit suspicious if migrations aren't typed properly

olevski: It was failing for missing annotations on stuff like `branch = None`. And we never used the variables where it was failing but I fixed it and removed the ignore statement in the pyproject file.

olevski (approved):

PR #89, opened 2023-11-15
chore: use async_sessionmaker

leafty:
See: https://docs.sqlalchemy.org/en/20/orm/extensions/asyncio.html#synopsis-orm.

This allows `mypy` to properly see types returned by `session.execute()` as being async, etc.

/deploy

### DIFF
--- a/components/renku_data_services/crc/db.py
+++ b/components/renku_data_services/crc/db.py
@@ -7,12 +7,12 @@ it all in one place.
 """
 from asyncio import gather
 from functools import wraps
-from typing import Any, Dict, List, Optional, Tuple, cast
+from typing import Any, Dict, List, Optional, Sequence, Tuple, cast
 
 from sqlalchemy import create_engine, delete, select
-from sqlalchemy.ext.asyncio import AsyncEngine, AsyncSession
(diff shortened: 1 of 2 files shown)


components/renku_data_services/crc/db.py
```diff
@@ -514,11 +512,11 @@ async def get_users(
...
                 else:
-                    stmt = select(schemas.UserORM)
+                    stmt2 = select(schemas.UserORM)
                     if keycloak_id is not None:
-                        stmt = stmt.where(schemas.UserORM.keycloak_id == keycloak_id)
-                    res = await session.execute(stmt)
-                    orms = res.scalars().all()
+                        stmt2 = stmt2.where(schemas.UserORM.keycloak_id == keycloak_id)
+                    res2 = await session.execute(stmt2)
+                    orms = res2.scalars().all()
```
leafty, line 519: Note here: it seems that `mypy` doesn't like re-using `stmt` and `res` (thinks they are related to type `ResourcePoolORM`).

olevski (approved):
Thats really nice. I did not know about it. And mypy was definitely struggling with quite a few of the sqlalchemy things.

PR #91, opened 2023-11-15
chore: move the config into its own component

olevski:
This does a few things:
- reduces (further) the risk that we spawn multiple connection pools (by assigning the connection pool as a class variable on the config class)
- pulls out the `Config` class into its own component - previously it was part of the data services base


olevski:
This is the followup from #83 @Panaetius

components/renku_data_services/config/config.py
```diff
@@ -66,25 +77,79 @@ def _oidc_discovery(url: str, realm: str) -> Dict[str, Any]:
...
+class DBConfig:
+    """Database configuration."""
+
+    password: str = field(repr=False)
+    host: str = "localhost"
+    user: str = "renku"
+    port: str = "5432"
+    db_name: str = "renku"
+    _async_engine: ClassVar[AsyncEngine | None] = field(default=None, repr=False, init=False)
+    _async_engine_write_lock: ClassVar[asyncio.Lock] = asyncio.Lock()
```
Panaetius, line 90: this isn't used anywhere

components/renku_data_services/migrations/env.py
```diff
@@ -44,6 +42,8 @@ def run_migrations_offline(target_metadata, sync_sqlalchemy_url: str) -> None:
         with context.begin_transaction():
             context.run_migrations()
 
+    engine.dispose()
```
Panaetius, line 45: I thought we didn't need dispose?

Panaetius (changes requested):

components/renku_data_services/migrations/env.py
olevski: Yes I forgot we dont need it.

olevski: Removed.

components/renku_data_services/config/config.py
olevski: Removed. I forgot it.

Panaetius (approved):

PR #92, opened 2023-11-20
feat: fix rclone schema and add support for validating storage connections

Panaetius:
examples wasn't returned.

updates apispec
adds check to ensure apispec is up to date
adds testing of connections using rclone
fixes bug in apispec with nullable values for patch endpoint


olevski (changes requested):
@Panaetius I am not sure if this changed in this PR or not. But the generated code for the storage apispec does not match the current apispec yaml file. I.e. if I run `make schemas` then I get changes in the generated python code for the storage apispec.

Panaetius:
> @Panaetius I am not sure if this changed in this PR or not. But the generated code for the storage apispec does not match the current apispec yaml file. I.e. if I run `make schemas` then I get changes in the generated python code for the storage apispec.

for some reason I thought that CI creates those. I've updated them now

olevski:
@Panaetius thanks for making the update. But the tests are failing now. Some endpoint is responding with a 422 whereas a 200 is expected. I think this is because of the apispec.py update.

And yeah we dont have a CI action that updates the code generated from the apispec.yaml file. We can add one that generates the code or one that simply checks that the generated code and the stuff in the yaml file match.

Panaetius:
it was due to a change rok made a while ago that broke things, but we never updater the apispec.py file so it didn't have any effect. Would be nice to have a test that that is up to date in CI

Panaetius:
@olevski I've added a test to ensure the schema is updated

olevski (approved):

olevski (approved):

PR #98, opened 2023-11-26
authorization

mohammad-alisafaee:


components/renku_data_services/project/db.py
```diff
@@ -52,21 +60,21 @@ async def get_projects(
...
-            # TODO: Filter based on the APIUser
-
             return [p.dump() for p in projects_orm], pagination
 
-    async def get_project(self, user: base_models.APIUser, id: str) -> models.Project:
+    async def get_project(self, user: base_models.APIUser, project_id: str) -> models.Project:
         """Get one project from the database."""
+        authorized = await self.project_authz.has_permission(user=user, project_id=project_id, scope=Scope.READ)
+        if not authorized:
+            raise errors.Unauthorized(message="You do not have the required permissions to access this project.")
```
olevski, line 69: This should be a 404. If you do not have access to a project and the project you should not even know it exists.


```suggestion
        if not authorized:
            raise errors.MissingResourceError(message=f"Project with id '{project_id}' does not exist or you do not have access to it.")

File: components/renku_data_services/project/db.py [hunk 2]
```diff
@@ -80,41 +88,53 @@ async def insert_project(self, user: base_models.APIUser, project: models.Projec
             async with session.begin():
                 session.add(project_orm)
 
+                # TODO: Test that raising an exception causes the transaction to be aborted or not
```
olevski, line 91: You can write a test about this if you want or test manually. But it would be nice to do this before we merge.

olevski (changes requested):

components/renku_data_services/authn/dummy.py
```diff
@@ -47,9 +47,12 @@ async def authenticate(access_token: str, request: Request) -> base_models.APIUs
...
+
+        is_set = bool(user_props.get("id") or user_props.get("name") or user_props.get("is_admin") is not None)
+
         return base_models.APIUser(
             is_admin=user_props.get("is_admin", False),  # type: ignore[arg-type]
-            id=user_props.get("id") if user_props.get("id") else "some-id",
+            id=user_props.get("id") if user_props.get("id") else "some-id" if is_set else None,
             access_token=access_token,
-            name=user_props.get("name") if user_props.get("name") else "John Doe",
+            name=user_props.get("name") if user_props.get("name") else "John Doe" if is_set else None,
```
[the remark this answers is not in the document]
mohammad-alisafaee: As discussed, this is required to allow creating an unauthorized user.

components/renku_data_services/project/db.py
```diff
@@ -80,41 +88,53 @@ async def insert_project(self, user: base_models.APIUser, project: models.Projec
...
             async with session.begin():
-                result = await session.execute(select(schemas.ProjectORM).where(schemas.ProjectORM.id == id))
+                result = await session.execute(select(schemas.ProjectORM).where(schemas.ProjectORM.id == project_id))
                 projects = result.one_or_none()
 
                 if projects is None:
-                    raise errors.MissingResourceError(message=f"The project with id '{id}' cannot be found")
-
-                # TODO: Check if the user can access this project
+                    raise errors.MissingResourceError(message=f"The project with id '{project_id}' cannot be found")
```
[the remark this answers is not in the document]
mohammad-alisafaee: We've already checked if the user has access to the project. So, if it can't be found at this point, it only means that the project doesn't exist.

components/renku_data_services/project/db.py
mohammad-alisafaee: This was a comment for myself to test this by hand. Basically, I wanted to make sure that this block works as a transaction. I believe it's not easy (and also needed) to write an automatic test for it.

mohammad-alisafaee: I'll remove all these TODOs in the final code.

components/renku_data_services/project/db.py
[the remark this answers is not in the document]
olevski: But its ok we can merge like this it is not a big deal.

olevski (approved):

PR #104, opened 2023-11-29
chore: add users endpoint and sync tests

olevski:
Adding tests

Note that the target branch is a feature branch.


components/renku_data_services/base_api/auth.py
```diff
@@ -45,3 +45,38 @@ async def decorated_function(request: Request, user: APIUser, *args, **kwargs):
...
+            api_user = kwargs["requested_by"]
+        elif "user" in kwargs and isinstance(kwargs["user"], APIUser):
+            api_user = kwargs["user"]
+        elif len(args) >= 1:
+            api_user_search = [a for a in args if isinstance(a, APIUser)]
+            if len(api_user_search) == 1:
+                api_user = api_user_search[0]
+            else:
+                raise errors.ProgrammingError(
+                    detail="Found two valid non-keyword APIUser arguments when authenticating user, expected only one."
```
Panaetius, line 72: should this say "found no or more than one"?

components/renku_data_services/base_api/auth.py
```diff
@@ -45,3 +45,38 @@ async def decorated_function(request: Request, user: APIUser, *args, **kwargs):
...
+                )
+
+        if api_user is None or not api_user.is_authenticated:
+            raise errors.Unauthorized(message="You have to be authenticated to perform this operation.")
+
+        # the user is authenticated
+        response = await f(self, *args, **kwargs)
+        return response
```
Panaetius, line 80: i wonder, can you just return the future instead of awaiting it and then the calling code has to await again?

components/renku_data_services/crc/blueprints.py
```diff
@@ -203,14 +203,14 @@ async def _get(_: Request, resource_pool_id: int, user_id: str, user: base_model
...
-                    message=f"The user with id {user_id} or resource pool with id {resource_pool_id} cannot be found."
-                )
-            return json(
-                apispec.UserWithId(id=res[0].keycloak_id, no_default_access=res[0].no_default_access).model_dump(
-                    exclude_none=True
+            if len(res.allowed) > 0:
+                return json(
+                    apispec.UserWithId(
+                        id=res.allowed[0].keycloak_id, no_default_access=res.allowed[0].no_default_access
+                    ).model_dump(exclude_none=True)
```
Panaetius, line 210: should there be an error that the resource pool does not allow any users?

components/renku_data_services/crc/db.py
```diff
@@ -487,35 +497,67 @@ async def get_resource_pool_users(
...
+                return RespositoryUsers(rp.id, allowed, disallowed)
+
+    @_only_admins
+    async def get_users(
+        self,
+        *,
+        api_user: base_models.APIUser,
+        id: str | None = None,
+    ) -> List[base_models.User]:
+        """Get users from the users-resource pools access lists."""
```
Panaetius, line 553: this seems to just get all users, nothing with access lists.

components/renku_data_services/crc/db.py
```diff
@@ -640,7 +682,7 @@ async def delete_resource_pool_user(self, api_user: base_models.APIUser, resourc
 
     @_only_admins
     async def update_resource_pool_users(
-        self, api_user: base_models.APIUser, resource_pool_id: int, users: List[base_models.User], append: bool = True
+        self, api_user: base_models.APIUser, resource_pool_id: int, user_ids: List[str], append: bool = True
     ) -> List[base_models.User]:
         """Update the users that have access to a specific resource pool."""
```
Panaetius, line 687:
```suggestion
        """Update the users to have access to a specific resource pool."""
```

Panaetius (changes requested):

mohammad-alisafaee (approved):
LGTM, thank you!

components/renku_data_services/crc/db.py
olevski: So the `UserORM` object is literally just a list of users that is joined with another table `resource_pools_users` to get ACLs. With the addition of a dedicated table of users synced with Keycloak this table is only storing the users that have (or are prevented from having) access to a resource pool. I will update the `UserORM` object to have a name that reflects this.

components/renku_data_services/crc/blueprints.py
olevski: I am not sure what you mean by this? This is used for a request that tries to get a specific resource pool for a specific user. So if either there is an authorization problem or the user or the resource pool does not exist then we raise an error.

olevski: But I see that for a default resource pool I was not adding the requesting users in the "allowed" list returned from the `get_resource_pool_users` properly. So I fixed that.

Panaetius: i mean if `allowed` is less than 1 this would print `f"The user with id {user_id} or resource pool with id {resource_pool_id} cannot be found."` when actually the pool was found, it just doesn't allow any users

components/renku_data_services/base_api/auth.py
olevski: Without the await this breaks at the sanic level.

```
  File "/home/renku/.cache/pypoetry/virtualenvs/renku-data-services-tbxlFWds-py3.11/lib/python3.11/site-packages/sanic/request/types.py", line 347, in respond
    response = await self.app._run_response_middleware(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "_run_response_middleware", line 17, in _run_response_middleware
    from asyncio.futures import Future
                            ^^^^^^^^^^^
  File "/home/renku/.cache/pypoetry/virtualenvs/renku-data-services-tbxlFWds-py3.11/lib/python3.11/site-packages/sanic_ext/extensions/http/cors.py", line 56, in _add_cors_headers
    if ORIGIN_HEADER not in response.headers:
                            ^^^^^^^^^^^^^^^^
AttributeError: 'coroutine' object has no attribute 'headers'
```

But I am sure we can work something out - maybe with middleware that checks for this case and awaits? Or maybe there is a nicer way. But simple removing the await does not work.

olevski: I would just do this in a follow up issue.

components/renku_data_services/crc/blueprints.py
olevski: As discussed this endpoint is to specifically check that a specific user can access a specific resource pool. If the permissions are not there then the requesting user should not know that the resource pool exists but they have no access to it.

olevski:
p.s. The test for patching a cloud storage entity is failing but not because we changed stuff here. It is being fixed in a separate PR: #107

components/renku_data_services/base_api/auth.py
Panaetius: oh it doesn't matter, i just thought that futures can be passed around without awaiting

Panaetius (approved):

PR #105, opened 2023-12-01
feat: project members

mohammad-alisafaee:


components/renku_data_services/project/api.spec.yaml
```diff
@@ -271,15 +257,13 @@ components:
         creation_date:
           $ref: "#/components/schemas/CreationDate"
         created_by:
-          $ref: "#/components/schemas/User"
+          $ref: "#/components/schemas/Member"
```
olevski, line 260: I don't know when we changed this but the example for this field a few lines below this contains the members property. But the members property is not part of the spec. So we should update that to match.

components/renku_data_services/project/api.spec.yaml
```diff
@@ -223,13 +219,13 @@ paths:
           schema:
             type: string
         - in: path
-          name: user_id
+          name: member_id
```
olevski, line 222: Can we be explicit what this is? It is the keycloak ID right?

components/renku_data_services/project/orm.py
```diff
@@ -53,8 +53,28 @@ def dump(self) -> models.Project:
...
             visibility=self.visibility,
-            created_by=models.User(id=self.created_by_id),
+            created_by=models.Member(id=self.created_by_id),
             creation_date=self.creation_date,
             repositories=self.repositories,
             description=self.description,
         )
+
+
+class ProjectMemberORM(BaseORM):
```
olevski, line 63: What is the purpose of this table? The authz service and database already contains pretty much the same information. 

I was hoping that the authz service and its tables would be the only source of truth for authorization data and membership.

olevski: See `components/renku_data_services/authz/orm.py`

olevski (changes requested):
Lets talk about this tomorrow. I think there is a few pieces of code here that just repeat stuff that is already implemented in the authz component.

components/renku_data_services/project/orm.py
mohammad-alisafaee: As agreed, this will be done in a future PR.

olevski (changes requested):
Can you please fix the tests? Currently only mypy is complaining.

Everything else is ok.

olevski (approved):

PR #106, opened 2023-12-01
chore: add users from access tokens

olevski:
Note that the target branch is not main but a feature branch.

Note this should be reviewed after #104 is merged. This builds on top of #104


components/renku_data_services/authn/dummy.py
```diff
@@ -48,11 +48,21 @@ async def authenticate(access_token: str, request: Request) -> base_models.APIUs
...
+            or user_props.get("email")
+        )
 
         return base_models.APIUser(
             is_admin=user_props.get("is_admin", False),  # type: ignore[arg-type]
             id=user_props.get("id") if user_props.get("id") else "some-id" if is_set else None,
             access_token=access_token,
+            first_name=user_props.get("first_name") if user_props.get("first_name") else "John" if is_set else None,
+            last_name=user_props.get("last_name") if user_props.get("last_name") else "Doe" if is_set else None,
+            email=user_props.get("email") if user_props.get("email") else "john.doe@gmail.com" if is_set else None,
```
Panaetius, line 66: I think you can just put "John" as the second parameter to `get` to get rid of one if,else. same on the other lines. and in general if we need multiple if/else in a line it's nice to add brackets to make the order clear

components/renku_data_services/base_models/core.py
```diff
@@ -23,7 +23,10 @@ class APIUser:
     is_admin: bool = False
     id: Optional[str] = None  # the sub claim in the access token - i.e. the Keycloak user ID
     access_token: Optional[str] = field(repr=False, default=None)
-    name: Optional[str] = None
+    name: Optional[str] = None  # the full name of the user
```
Panaetius, line 26: should we just cal this fullname?

components/renku_data_services/users/db.py
```diff
@@ -65,16 +54,26 @@ async def get_user(self, requested_by: APIUser, id: str) -> UserInfo | None:
             stmt = select(UserORM).where(UserORM.keycloak_id == id)
             res = await session.execute(stmt)
             orm = res.scalar_one_or_none()
+            if not orm and id == requested_by.id:
+                return await self._add_api_user(requested_by)
```
Panaetius, line 58: get_user inserting a user in the db class is a bit odd, shouldn't this be handled by the client or calling code? Seems like the kind of side-effect that could bite us in the future

components/renku_data_services/users/db.py
```diff
@@ -65,16 +54,26 @@ async def get_user(self, requested_by: APIUser, id: str) -> UserInfo | None:
...
     async def get_users(self, requested_by: APIUser, email: str | None = None) -> List[UserInfo]:
-        """Get user from the database."""
+        """Get users from the database."""
         if not email and not requested_by.is_admin:
             raise errors.Unauthorized(message="Non-admin users cannot list all users.")
-        return await self._get_users(email)
+        users = await self._get_users(email)
+
+        def is_api_user_missing() -> bool:
+            return not any([requested_by.id == user.id for user in users])
```
Panaetius, line 71: couldn't this just be `api_user_is_missing = ...` instead of an inner function?

Panaetius (changes requested):

components/renku_data_services/users/db.py
Panaetius: at the very least it should be called get_or_create_user or similar

olevski: The reason for this is to avoid cases where a freshly signed up user hits the /user endpoint in the API and we return a 404 because the data services API has not synced with Keycloak yet. And the user will be considered to not exist for the next 2-3 minutes until the cronjob that syncs keyclaok and the data services fires and completes.

It is not that nice. But I have no idea how to avoid this problem for newly created users. If we do not have this in place we will definitely make it so that newly created users dont exist for a little while.

olevski: I will rename the function though.

components/renku_data_services/users/db.py
olevski: Yeah I have no clue why I did that.

components/renku_data_services/authn/dummy.py
olevski: Yeah that is right. I kinda left this as is before because it is only in the dummy stuff used only for testing. But still it is nice to have it cleaner like this.

Panaetius (approved):

PR #109, opened 2023-12-07
chore: relax permissions for listing members

olevski:
With this change every logged in user can list other users.

I think we should always allow this but limit what fields are returned based on the role of the user (i.e. admin or not).


mohammad-alisafaee (approved):
LGTM, thank you!

PR #112, opened 2023-12-22
fix: allow removing tolerations and affinities

### LINKED ISSUE #95: cannot remove tolerations or affinities with PUT/PATCH on resource class
See [`update_resource_class()`](https://github.com/SwissDataScienceCenter/renku-data-services/blob/main/components/renku_data_services/crc/db.py#L416).

olevski:
closes #95 

There are a few ways to remove tolerations or patches:
- remove all of them for a class by calling the DELETE endpoint
- use PUT and passing in an empty list
- use PUT and removing some tolerations/affinities from the list

Hopefully this is acceptable.

@leafty


components/renku_data_services/crc/blueprints.py
```diff
@@ -328,24 +335,56 @@ async def _patch(
...
+        return "/resource_pools/<resource_pool_id>/classes/<class_id>/tolerations", ["GET"], _get_tolerations
+
+    def delete_tolerations(self) -> BlueprintFactoryResponse:
+        """Delete all tolerations of a resource class."""
+
+        @authenticate(self.authenticator)
+        @only_admins
+        async def _delete_tolerations(_: Request, resource_pool_id: int, class_id: int, user: base_models.APIUser):
+            await self.repo.delete_tolerations(user, resource_pool_id, class_id)
+            return HTTPResponse(status=204)
```
Panaetius, line 363: use `sanic.empty(status=204)` instead as that's what it's made for. instead of the HTTPResponse base class

components/renku_data_services/crc/blueprints.py
```diff
@@ -328,24 +335,56 @@ async def _patch(
...
+        return "/resource_pools/<resource_pool_id>/classes/<class_id>/node_affinities", ["GET"], _get_affinities
+
+    def delete_affinities(self) -> BlueprintFactoryResponse:
+        """Delete all affinities of a resource class."""
+
+        @authenticate(self.authenticator)
+        @only_admins
+        async def _delete_affinities(_: Request, resource_pool_id: int, class_id: int, user: base_models.APIUser):
+            await self.repo.delete_affinities(user, resource_pool_id, class_id)
+            return HTTPResponse(status=204)
```
Panaetius, line 385: use `empty()`

components/renku_data_services/crc/db.py
```diff
@@ -466,6 +487,62 @@ async def update_resource_class(
...
+    @_only_admins
+    async def delete_tolerations(self, api_user: base_models.APIUser, resource_pool_id: int, class_id: int):
+        """Delete all tolerations for a specific resource class."""
+        async with self.session_maker() as session, session.begin():
+            res_classes = await self.get_classes(api_user, class_id, resource_pool_id=resource_pool_id)
+            if len(res_classes) == 0:
+                raise errors.MissingResourceError(
+                    message=f"The resource pool with ID {resource_pool_id} or the resource "
+                    f"class with ID {class_id} do not exist, or they are not related."
+                )
```
Panaetius, line 513: should delete be idempotent?

components/renku_data_services/crc/db.py
```diff
@@ -466,6 +487,62 @@ async def update_resource_class(
...
+    @_only_admins
+    async def delete_affinities(self, api_user: base_models.APIUser, resource_pool_id: int, class_id: int):
+        """Delete all affinities from a resource class."""
+        async with self.session_maker() as session, session.begin():
+            res_classes = await self.get_classes(api_user, class_id, resource_pool_id=resource_pool_id)
+            if len(res_classes) == 0:
+                raise errors.MissingResourceError(
+                    message=f"The resource pool with ID {resource_pool_id} or the resource "
+                    f"class with ID {class_id} do not exist, or they are not related."
+                )
```
Panaetius, line 542: should delete be idempotent?

Panaetius (changes requested):

components/renku_data_services/crc/db.py
olevski: You mean that this error should not be raised at all? 

I think an error should be raised here because the resource class or pool that contains the toleration does exist. So in this case the caller of this method has very wrong assumptions about what they are doing. It is not that they are trying to delete the same toleration again (this will not error). This error is raised when:
- the resource pool does not exist
- the resource class does not exist
- the resource class does not belong to the pool

components/renku_data_services/crc/db.py
olevski: Same comment as above. This is a situation where the caller is making a mistake or has a wrong idea about the state of the pools/classes that contain the toleration or affinity.

Panaetius (approved):

PR #113, opened 2023-12-22
fix: schemathesis tests

olevski:
Fixed a few problems with edge cases that schemathesis throws at the API.

I tried to keep the time limit to 100ms but listing all resource pools took 101ms. So I just increased the limit to 120ms. I think this is still acceptable. Locally on my machine it took 150ms and removing all additional logic and just doing a simple query to list all resource pools without filtering for resource classes or anything still took 150-140ms. So the time spent on that endpoint goes all to the database.

Lastly the reason that the tests were not failing when schemathesis was failing is because in the Makefile we ran the schemathesis tests in the background which ignores the exit code. Now this is not the case and I confirmed when the schemathesis fails the tests also fail.


components/renku_data_services/crc/orm.py
```diff
@@ -152,15 +156,21 @@ def dump(self, quota: models.Quota | None) -> models.ResourcePool:
...
             )
         return models.ResourcePool(
             id=self.id,
             name=self.name,
             quota=quota,
-            classes=[resource_class.dump() for resource_class in classes],
+            classes=[
+                resource_class.dump(
+                    resource_class.dump() >= class_match_criteria if class_match_criteria is not None else None
+                )
```
Panaetius, line 171: **question (non-blocking):** this looks weird to me, we dump the class twice, once for matching and then once for the output. could this be done in one go by passing the class_match_criteria into the dump method or by dumping first and then setting `matching` afterwards?

components/renku_data_services/user_preferences/blueprints.py
```diff
@@ -36,6 +37,8 @@ def post_pinned_projects(self) -> BlueprintFactoryResponse:
         @authenticate(self.authenticator)
         @validate(json=apispec.AddPinnedProject)
         async def _post(_: Request, body: apispec.AddPinnedProject, user: base_models.APIUser):
+            if "\u0000" in body.project_slug:
```
Panaetius, line 40: **suggestion (non-blocking):** should we instead validate project slug to something like `a-zA-Z0-9-_`?

.github/workflows/test_publish.yml
```diff
@@ -17,16 +17,13 @@ jobs:
       - uses: actions/checkout@v3
       - name: Build and run dev container task
         uses: devcontainers/ci@v0.3
-        env:
-          USER_UID: 1001
-          USER_GID: 123
```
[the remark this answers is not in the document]
olevski: With this the installation of polylith in the devcontainer in the github action was failing. After I removed it everything worked and the installation of polylith in the github action was successful.

components/renku_data_services/crc/orm.py
olevski: Yeah I mean to go back to this and clean it up.

Panaetius (dismissed):

Panaetius (approved):

PR #116, opened 2024-01-30
feat: secrets storage

m-alisafaee:


components/secret_storage/secret/orm.py
```diff
@@ -0,0 +1,67 @@
...
+from sqlalchemy import DateTime, MetaData, String
+from sqlalchemy.orm import DeclarativeBase, Mapped, MappedAsDataclass, mapped_column
+from ulid import ULID
+
+metadata_obj = MetaData(schema="secrets")  # Has to match alembic ini section name
+
+
+def encrypt_string(key: bytes, data: str) -> str:
+    """Encrypt a given string."""
+    return Fernet(key).encrypt(data.encode()).decode()
```
Panaetius, line 17: ideally we would use a salt, so if someone has the key they can't just decrypt everything easily.
https://cryptography.io/en/latest/fernet/#using-passwords-with-fernet has an example.
We just need to pick a good salt,like maybe user id for user secrets, group id for group secrets etc.

Panaetius (commented):
this is great. I know this feature is very eagerly awaited by quite a few people :slightly_smiling_face:

Panaetius:
we should also add a follow-up issue to support key rotation: https://cryptography.io/en/latest/fernet/#cryptography.fernet.MultiFernet.rotate
like to automatically do it every 90 days or manually if someone with access to the key leaves the team

components/renku_data_services/secrets_storage/secret/api.spec.yaml
```diff
@@ -0,0 +1,230 @@
...
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/SecretsList"
+        default:
+          $ref: "#/components/responses/Error"
+      tags:
+        - secrets_storage_api
+    post:
+      summary: Add a new secret
```
lorenzo-cavazzi, line 32: Here it's missing the `user_id` parameter:
```suggestion
      summary: Add a new secret
      parameters:
        - in: path
          name: user_id
          required: true
          schema:
            type: string
```

components/renku_data_services/secrets_storage/secret/api.spec.yaml
```diff
@@ -0,0 +1,230 @@
...
+      tags:
+        - secrets_storage_api
+    delete:
+      summary: Remove a secret
+      parameters:
+        - in: path
+          name: user_id
+          required: true
+          schema:
+            type: string
```
lorenzo-cavazzi, line 114:
```suggestion
```
This endpoint doesn't have the `user_id` parameter in the path. Actually, instead of removing the parameter definition, we might want to add it to the path since `/secrets/{user_id}/{secret_id}` might be safer -- or it might be needed if `secret_id` might be a common string

lorenzo-cavazzi (changes requested):
Couple of changes on the API spec.
I didn't check the code, but I assume minimal adjustments are required (for sure if we decide to change the path)

components/renku_data_services/users/api.spec.yaml
```diff
@@ -66,8 +66,145 @@ paths:
...
+          content:
+            "application/json":
+              schema:
+                $ref: "#/components/schemas/SecretWithId"
+        default:
+          $ref: '#/components/responses/Error'
+      tags:
+        - users
+        - secrets
+  /user/{user_id}/secrets/{secret_id}:
```
lorenzo-cavazzi, line 126:
```suggestion
  /users/{user_id}/secrets/{secret_id}:
```

components/renku_data_services/users/api.spec.yaml
```diff
@@ -66,8 +66,145 @@ paths:
...
+          name: secret_id
+          required: true
+          schema:
+            type: string
+      responses:
+        "204":
+          description: The secret was deleted or didn't exist
+        default:
+          $ref: "#/components/responses/Error"
+
```
lorenzo-cavazzi, line 207:
```suggestion
      tags:
        - users
        - secrets

```

lorenzo-cavazzi (commented):
I guess the `/user` instead of `/users` is a typo -- not 100% sure tho since there are some routes with `/user`.

components/renku_data_services/users/db.py
```diff
@@ -277,3 +303,99 @@ async def events_sync(self, kc_api: IKeycloakAPI):
...
+        self,
+        session_maker: Callable[..., AsyncSession],
+    ):
+        self.session_maker = session_maker
+
+    @only_authenticated
+    async def get_secrets(self, requested_by: APIUser, user_id: str) -> list[Secret]:
+        """Get a specific user secret from the database."""
+        if user_id != requested_by.id:
+            raise errors.Unauthorized(message="Cannot list secrets.")
```
sgaist, line 321: Since this check is pertinent to all functions, a decorator would make things simpler. It could have the error message as parameter.

test/components/renku_data_services/utils/test_utils.py
```diff
@@ -0,0 +1,12 @@
...
+
+def test_can_decrypt_correctly():
+    data = "some data"
+    password = b"some password"
+    salt = "some salt"
+    encrypted_data = encrypt_string(password=password, salt=salt, data=data)
+
+    decrypted_data = decrypt_string(password=password, salt=salt, data=encrypted_data)
+
+    assert decrypted_data == "some data"
```
sgaist, line 12:
```suggestion
    assert decrypted_data == data
```

test/bases/renku_data_services/secrets_storage_api/test_secret.py
```diff
@@ -0,0 +1,222 @@
...
+    assert response.status_code == 200, response.text
+
+    _, response = await sanic_client.get(
+        f"/api/secret/secrets/{secret_id}", headers=user_headers
+    )
+
+    assert response.status_code == 200, response.text
+    assert response.json is not None
+    assert response.json["name"] == "new-name"
+    assert response.json["value"] == "new-value"
```
sgaist, line 176:
```suggestion
assert response.json["name"] == payload["name"]
assert response.json["value"] == payload["value"]
```

components/renku_data_services/secrets_storage/app_config/config.py
```diff
@@ -0,0 +1,86 @@
...
+            if jwks_url is None:
+                raise errors.ConfigurationError(
+                    message="The JWKS url for Keycloak cannot be found from the OIDC discovery endpoint."
+                )
+            algorithms = os.environ.get(f"{prefix}KEYCLOAK_TOKEN_SIGNATURE_ALGS")
+            if algorithms is None:
+                raise errors.ConfigurationError(
+                    message="At least one token signature algorithm is required."
+                )
+            algorithms_lst = [i.strip() for i in algorithms.split(",")]
```
sgaist, line 82: While I am not a big fan of single letter variables, wouldn't `a` make more sense if not using `algorithm` ?

components/renku_data_services/app_config/config.py
```diff
@@ -377,4 +401,6 @@ def from_env(cls, prefix: str = ""):
             redis=redis,
             kc_api=kc_api,
             message_queue=message_queue,
+            encryption_key=encryption_key,
+            secret_service_public_key=secret_service_public_key,
```
sgaist, line 405: Shouldn't that rather be `secrets_service_public_key` ?

sgaist (changes requested):
Hi,

While I wanted to test my suggestion to avoid code duplication when `COVERAGE_RUN` is defined, I failed to get started because it seems that the `secret` folder is in the wrong folder and should be in `renku_data_services/secrets_storage` rather than just `renku_data_services/` (based on the number of import that search for it there). I got further after moving it there however, now alembic is pretty unhappy.

The overall implementation looks sane however I think there are several instances where there's a mix of single and plural form of Secrets that makes the code a bit hard to read/reason about. There's a need of secret stuff like the secret key matching the public key for encryption/decryption however the service is dealing with user secrets which can make thing a bit hairy to follow. I have made some suggestions in this direction but there are more that I think are worth considering (maybe document as well to avoid contributors to be taken off guard).

Panaetius:
this is a draft PR that is currently being worked on and isn't done. Everything in secrets_storage* and secrets/ is being reworked, the current code is previous work from before we settled on the new solution. the only part in this PR that has some semblance of being done are the changes in the */users/ folder (i.e. the UI facing part of the solution), together with the chart changes in the main renku repo to get this to work.
Even that isn't done, it lacks tests, and I probably need to go over it again to do cleanup. The main goal was to get the frontend facing part done for now, so the UI team can continue working against it, even if it's not yet polished (i.e. it can store secrets but any code about doing something with them isn't there).
At this point, most of what is on this PR should be considered exploratory work. I'll probably work on the backend facing side of this in a separate PR based off of this one some time in the coming days.

olevski (changes requested):
I have a few changes that I requested.

Some are really picky like naming and can be done at a later time.

One general comment is that in a lot of places we want to hid the value of the string that holds the encryption keys. I think for this we should make a subclass or alias of string, overwrite its `__repr__` field and then use that everywhere. That way we can be sure that if the class that holds this is printed accidentally it will not be shown. And mypy will (maybe?) catch cases where someone is using a string when they should using this "redacted string". Even if mypy does not catch this it is useful because developers can see the intent a lot easier and use the right type.

olevski:
Also maybe I missed this in the tests. But perhaps we should add a test that specifically checks for the double encryption and the fact that neither the data service nor the secret service can independently decrypt a secret once it has been created.

components/renku_data_services/app_config/config.py
```diff
@@ -127,9 +120,13 @@ class Config:
     gitlab_client: base_models.GitlabAPIProtocol
     kc_api: IKeycloakAPI
     message_queue: IMessageQueue
+    secrets_service_public_key: rsa.RSAPublicKey
+    """The public key of the secrets service, used to encrypt user secrets that only it can decrypt."""
     spec: dict[str, Any] = field(init=False, default_factory=dict)
+    encryption_key: bytes = field(repr=False)
+    """The encryption key to encrypt user keys at rest in the database."""
```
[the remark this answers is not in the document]
Panaetius: though the usual way with google style would be to have it in the class docstring, like 
```
class Foo:
    """The foo class.

    Attributes:
        bar(str): the bar attribute
    """

    bar: str
``` 
but I only really like this when actually everything is documented, if it's just for select attributes and others are missing, it's a bit odd.

bases/renku_data_services/secrets_storage_api/main.py
```diff
@@ -0,0 +1,42 @@
...
+    """Create a Sanic application."""
+    config = Config.from_env()
+    app = Sanic(config.app_name)
+    if "COVERAGE_RUN" in environ:
+        app.config.TOUCHUP = False
+    app = register_all_handlers(app, config)
+
+    @app.main_process_start
+    async def do_migrations(*_):
+        logger.info("running migrations")
```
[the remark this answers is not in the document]
Panaetius: ah that's a leftover. renku_data_services does all the migrations, just so they're in one place and we don't get any issues with concurrency or anything

components/renku_data_services/secrets/blueprints.py
```diff
@@ -0,0 +1,66 @@
...
+            with contextlib.suppress(k8s_client.ApiException):
+                # try and delete the secret in case it already existed
+                self.core_client.delete_namespaced_secret(body.name, body.namespace)
+
+            secret = k8s_client.V1Secret(
+                data=decrypted_secrets,
+                metadata={"name": body.name, "namespace": body.namespace, "owner_references": body.owner_references},
+            )
+
+            self.core_client.create_namespaced_secret(body.namespace, secret)
```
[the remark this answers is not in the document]
Panaetius: we already do, through the owner_references field.

components/renku_data_services/users/blueprints.py
```diff
@@ -71,3 +90,108 @@ async def _get_one(request: Request, user_id: str, user: base_models.APIUser):
...
+
+        Since RSA cannot encrypt arbitrary length strings, we use symmetric encryption with a random key and encrypt the
+        random key with RSA to get it to the secrets service.
+        """
+        secret_key = await self.user_repo.get_or_create_user_secret_key(requested_by=requested_by)
+        # encrypt once with user secret
+        encrypted_value = encrypt_string(secret_key.encode(), requested_by.id, value)  # type: ignore
+        # encrypt again with secret service public key
+        encryption_key = generate_random_encryption_key()
+        encrypted_value = encrypt_string(encryption_key, requested_by.id, encrypted_value.decode())  # type: ignore
```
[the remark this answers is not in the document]
Panaetius: as mentioned I created an issue for this. In the meantime I changed it to check and raise and removed the ignore.

components/renku_data_services/utils/cryptography.py
```diff
@@ -0,0 +1,54 @@
...
+
+import base64
+
+from cryptography.fernet import Fernet
+from cryptography.hazmat.primitives import hashes
+from cryptography.hazmat.primitives.asymmetric import padding, rsa
+from cryptography.hazmat.primitives.kdf.pbkdf2 import PBKDF2HMAC
+from sqlalchemy import MetaData
+
+metadata_obj = MetaData(schema="secrets")  # Has to match alembic ini section name
```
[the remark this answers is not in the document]
Panaetius: I have no idea what this is doing here, I moved it to the secrets/orm.py where it belongs, same as every other module that has this.

components/renku_data_services/app_config/config.py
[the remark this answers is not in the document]
olevski: Ok no problem we stick with what we have. No need to change.

components/renku_data_services/secrets/blueprints.py
```diff
@@ -0,0 +1,66 @@
...
+                )
+            decrypted_secrets = {}
+            for secret in secrets:
+                decryption_key = decrypt_rsa(self.secret_service_private_key, secret.encrypted_key)
+                decrypted_value = decrypt_string(decryption_key, user.id, secret.encrypted_value).encode()  # type: ignore
+                decrypted_secrets[secret.name] = b64encode(decrypted_value)
+
+            with contextlib.suppress(k8s_client.ApiException):
+                # try and delete the secret in case it already existed
+                self.core_client.delete_namespaced_secret(body.name, body.namespace)
```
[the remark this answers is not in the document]
Panaetius: I've changed it to just try and create and raise if it doesn't work (with just the exception name to not leak something). and it's up to the caller to make sure that the secret doesn't exist

olevski (approved):

olevski (approved):

Panaetius (approved):

lorenzo-cavazzi (approved):

PR #119, opened 2024-02-11
compute sessions

mana-alisafaee:


components/renku_data_services/session/api.spec.yaml
```diff
@@ -0,0 +1,267 @@
...
+        creation_date:
+          $ref: "#/components/schemas/CreationDate"
+        description:
+          $ref: "#/components/schemas/Description"
+        environment_id:
+          $ref: "#/components/schemas/EnvironmentId"
+        status:
+          $ref: "#/components/schemas/Status"
+        tags:
+          $ref: "#/components/schemas/TagsList"
```
leafty, line 142: I don't think we have anything yet in the pitch about tags. We would have to understand where to put them between sessions and environments, so I'm not sure we can use them now.

components/renku_data_services/session/api.spec.yaml
```diff
@@ -0,0 +1,267 @@
...
+        - sessions
+components:
+  schemas:
+    SessionsList:
+      description: A list of Renku sessions
+      type: array
+      items:
+        $ref: "#/components/schemas/Session"
+      minItems: 0
+    Session:
```
leafty, line 124: A session belongs to a project. I think there is a missing `project_id`.

components/renku_data_services/session/api.spec.yaml
mana-alisafaee: I'll remove them for now.

components/renku_data_services/session/api.spec.yaml
```diff
@@ -0,0 +1,250 @@
...
+info:
+  title: Renku 1.0 Environment and Session Management Service
+  description: |
+    Service that allows creating, deleting, and managing Renku 1.0 environments and sessions.
+  version: 0.1.0
+servers:
+  - url: /api/data
+  - url: /ui-server/api/data
+paths:
+  /sessions:
```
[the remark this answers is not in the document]
mana-alisafaee: As discussed, this will be implemented later in this pitch.

leafty (commented):
You need to add `run_migrations_for_app("sessions")` in `bases/renku_data_services/data_api/main.py` so that the new table is created.

leafty:
Should we merge this into a feature branch? Or do we consider this the feature branch?

components/renku_data_services/session/blueprints.py
```diff
@@ -0,0 +1,92 @@
...
+    authenticator: base_models.Authenticator
+
+    def get_all(self) -> BlueprintFactoryResponse:
+        """List all sessions."""
+
+        @authenticate(self.authenticator)
+        async def _get_all(_: Request, *, user: base_models.APIUser):
+            sessions = await self.session_repo.get_sessions(user=user)
+            return json(
+                [apispec.Session.model_validate(p).model_dump(exclude_none=True, mode="json") for p in sessions], 201
```
leafty, line 33: This should send back a 200:

```suggestion
                [apispec.Session.model_validate(p).model_dump(exclude_none=True, mode="json") for p in sessions]
```

mana-alisafaee:
> Should we merge this into a feature branch? Or do we consider this the feature branch?

We can consider this as the feature branch.

PR #127, opened 2024-02-26
Pitch/compute sessions start session

mana-alisafaee:


leafty:
Upon reflection, I am not sure we should start the sessions from `renku-data-services` for now. This adds API coupling between `renku-data-services` and `renku-notebooks`, as the service needs to understand and pass sessions options (e.g. session class, storage mounts, environment variables) to the notebooks. I think this would need some design and reflection to be done right.

For now we can launch sessions from `renku-ui` as is already the case and trim this down to be done at a later time.

PR #131, opened 2024-02-27
chore: apispec updates

olevski:
@ciyer note that this targets your branch.

Also the apispec will keep evolving by getting new endpoints and to a lesser extent getting new fields.

I have added some basic code to support this but it is not tested at all. Once I add some tests I will push the code as well so that you can start testing with the backend.

Also I have not added anything yet about the user namespaces. These are just the regular group namespaces not tied to a user. But we will need some additional endpoints for the namespaces that are automatically created for a user.

Let me know what you think.


olevski:
One thing to note that may seem weird but was kinda needed. I added "Group" in front of some of the schemas because we merge all apispecs into one and when we have duplicates we pick one of them and discard the other. So I was worried about having collisions with stuff in the projects and added "Group" to some of the schemas to avoid this.

components/renku_data_services/group/api.spec.yaml
```diff
@@ -36,7 +36,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: "#/components/schemas/GroupsList"
+                $ref: "#/components/schemas/GroupResponseList"
```
ciyer, line 39: I like this renaming -- makes the semantics clearer.

components/renku_data_services/group/api.spec.yaml
```diff
@@ -270,146 +273,134 @@ components:
...
       format: ulid
-    Name:
+    GroupName:
       description: Renku group name
       type: string
       minLength: 1
       maxLength: 99
       example: "My Renku Group :)"
-    Slug:
+    GroupSlug:
```
ciyer, line 315: Do we want to have a type name for each kind of slug we have (e.g., `ProjectSlug`, `GroupSlug`...)? If we can keep the structure of all slugs the same (and I think we should do this), then I would favor having just one `Slug` type, ideally with it defined only once in some shared location.

components/renku_data_services/group/api.spec.yaml
```diff
@@ -270,146 +273,134 @@ components:
...
       description: A command-line/url friendly name for a group
       type: string
       minLength: 1
       maxLength: 99
       pattern: "^[a-z0-9]+[a-z0-9._-]*$"  # Cannot contain consecutive special characters or end in .git and .atom
       example: "my-renku-group"
-    CreationDate:
-      description: The date and time the group was created (time is always in UTC)
+    GroupCreationDate:
```
ciyer, line 322: The same comment for "slug" applies here -- I think we should have only one format for creation dates and use that everywhere, rather than having potentially a different format for each entity.

components/renku_data_services/group/api.spec.yaml
```diff
@@ -270,146 +273,134 @@ components:
...
+      description: The date and time the group was created (in UTC and ISO-8601 format)
       type: string
       format: date-time
       example: "2023-11-01T17:32:28Z"
-    Description:
+    GroupDescription:
```
ciyer, line 327: See above comment about "slug" and "creation date".

ciyer (commented):
These are good improvements! For a few of the types, I have a question, thought about whether we want to have a global, shared e.g., "Slug" type or entity-specific "GroupSlug," etc. types.

ciyer (approved):
Looks great!

PR #137, opened 2024-03-04
OpenAPI spec proposal for user namespaces

ciyer:
OpenAPI spec proposal for user namespaces


olevski (approved):

ciyer:
Closing, since these changes are no longer needed. The creation of user namespaces will be handled differently

PR #140, opened 2024-03-07
Associate V2 projects with namespaces

ciyer:
* Namespace endpoint for listing all namespaces (group and user)
* Add namespace field to Project


olevski (approved):

components/renku_data_services/group/api.spec.yaml
```diff
@@ -241,6 +241,85 @@ paths:
...
+          description: Result's page number starting from 1
+          name: page
+          required: false
+          schema:
+            type: integer
+            minimum: 1
+            default: 1
+        - in: query
+          description: The number of results per page
+          name: per_page
```
ciyer, line 258: In the query parameters the field is `per_page`, but in the response it is `per-page`. This seems to be the case in all the services. Is there a reason that the name is not the same everywhere?

olevski: @ciyer we can change this to be the same everywhere as long as you make the corresponding change in the ui.

I would say lets just get this merged and then I can take care of this in the namespaces PR.

PR #144, opened 2024-03-11
chore: combine alembic version tables

olevski:
This PR does a few things:
- merges all alembic version tables into a single table
- assigns alembic branches to all existing alembic migrations that up till now were using separate version tables
- merges all alembic branches into a single migration so that going forward we do not have an explosion of branches
- updates the migrations and how they run 
- updates the docs
- applies migrations from older code that seems to have been missed - i.e. we modified the sqlalchemy models but we did not run migrations

@Panaetius unfortunately I had to run custom code to migrate the version tables. I could not find a way to avoid this - migrations are not meant to migrate the version table.

After this going back to separate version tables will require manual changes.


components/renku_data_services/migrations/utils.py
```diff
@@ -0,0 +1,139 @@
+"""Custom migrations env file to support modular migrations."""
```
olevski, line 1: This whole file is just the former content of env.py with minor changes.

For a true diff run git diff main:components/renku_data_services/migrations/env.py HEAD:components/renku_data_services/migrations/utils.py` after you have checked out this branch.

components/renku_data_services/migrations/versions/a06e2c511bea_run_missed_migrations.py
```diff
@@ -0,0 +1,69 @@
...
+down_revision = "e16782d6bf32"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.alter_column(
+        "projects", "creation_date", existing_type=postgresql.TIMESTAMP(timezone=True), nullable=True, schema="projects"
+    )
```
leafty, line 23: I think it would be nice to have this column not be nullable. Maybe the orm file needs to specify `nullable=False`?

components/renku_data_services/project_migrations/versions/7c08ed2fb79d_generate_tables.py
```diff
@@ -61,4 +61,5 @@ def downgrade() -> None:
     )
     op.drop_table("projects_repositories", schema="projects")
     op.drop_table("projects", schema="projects")
+    op.execute("DROP TYPE visibility CASCADE")
```
leafty, line 64: Note: another way to handle this is to have the following.

```python
  def downgrade() -> None:
    # ### commands auto generated by Alembic - please adjust! ###
    from renku_data_services.project.apispec import Visibility

    op.drop_index(
        op.f("ix_projects_projects_repositories_project_id"), table_name="projects_repositories", schema="projects"
    )
    op.drop_table("projects_repositories", schema="projects")
    op.drop_table("projects", schema="projects")
    sa.Enum(Visibility).drop(op.get_bind())
    # ### end Alembic commands ###
```

components/renku_data_services/migrations/versions/a06e2c511bea_run_missed_migrations.py
olevski: Fixed.

Panaetius (dismissed):

Panaetius (dismissed):

leafty (approved):

Panaetius (approved):

Issue #146, opened 2024-03-13
Namespace leftovers


ciyer:
The namespace build concentrated on the core functionality, but there are some loose ends that should be tied up.

* **Folder rename** The spec and and implementation for working with namespaces is in older `components/renku_data_services/group/`. This should be renamed to `components/renku_data_services/namespace/`
* **Pagination field unification** In the query parameters the field is called `per_page`, but in the response it is `per-page`. We should pick one canonical name and use it everywhere (also impacts other V2 endpoints with pagination, like projects)

PR #149, opened 2024-03-13
feat: initial namespace implementation

olevski:
The namespace slug is not shown in the project listing endpoints yet.

But the two endpoints that list namespaces work. And the list of users is used to create the namespaces.

/deploy renku-ui=ciyer/3023-namespace-projects renku=build/search-and-discovery-1.0 renku-gateway=master


components/renku_data_services/base_api/pagination.py
```diff
@@ -13,10 +13,17 @@ class PaginationRequest(NamedTuple):
     page: int
     per_page: int
 
+    def __post_init__(self):
+        if self.page > 2**63 - 1:
```
ciyer, line 17: I do not understand the need for this guard. 

In Python, I believe integers will always be promoted when necessary and can accommodate arbitrary size. If this limit is imposed elsewhere, shouldn't we use unsigned 64-bit int (i.e., limit is 2**64 -1) and not signed ints?

components/renku_data_services/base_api/pagination.py
```diff
@@ -13,10 +13,17 @@ class PaginationRequest(NamedTuple):
...
+            raise errors.ValidationError(message="Parameter 'page' is too large")
+
     @property
     def offset(self) -> int:
         """Calculate an item offset required for pagination."""
-        return (self.page - 1) * self.per_page
+        output = (self.page - 1) * self.per_page
+        if output > 2**63 - 1:
```
ciyer, line 24: See comment above.

components/renku_data_services/namespace/apispec.py
```diff
@@ -34,7 +34,7 @@ class NamespaceResponse(BaseAPISpec):
         example="my-renku-namespace",
         max_length=99,
         min_length=1,
-        pattern="^[a-z0-9]+[a-z0-9._-]*$",
+        pattern="^(?!.*\\.git$|.*\\.atom$|.*[\\-._][\\-._].*)[a-zA-Z0-9][a-zA-Z0-9\\-_.]*$",
```
ciyer, line 37: This same regex appears in many places. Would it be possible to make it a global or at least module-level constant?

components/renku_data_services/namespace/db.py
```diff
@@ -0,0 +1,374 @@
...
+
+
+class GroupRepository:
+    """Repository for groups."""
+
+    def __init__(self, session_maker: Callable[..., AsyncSession], group_authz=None):
+        self.session_maker = session_maker  # type: ignore[call-overload]
+        self.group_authz: None | Any = group_authz
+
+    async def generate_user_namespaces(self):
```
ciyer, line 32: I'm not sure this is working as expected. I am not seeing a namespace for my user in the test deployment.

components/renku_data_services/namespace/db.py
```diff
@@ -0,0 +1,374 @@
...
+                        group.members[new_member.id].role = role.value
+                    output.append(group.members[new_member.id].dump())
+                else:
+                    # The memeber does not exist in the group at all, make a new entry and add it to the group
+                    member_orm = schemas.GroupMemberORM(user_id=new_member.id, role=role)
+                    group.members[new_member.id] = member_orm
+                    output.append(member_orm.dump())
+            return output
+
+    async def delete_group(self, user: base_models.APIUser, slug: str):
```
ciyer, line 243: I created group A, created a project in that group, and then  deleted the group. The group is gone, but the project is still associated with the group.

components/renku_data_services/namespace/blueprints.py
```diff
@@ -141,3 +142,47 @@ async def _delete_member(_: Request, *, user: base_models.APIUser, slug: str, us
             return HTTPResponse(status=204)
 
         return "/groups/<slug>/members/<user_id>", ["DELETE"], _delete_member
+
+    def get_namespaces(self) -> BlueprintFactoryResponse:
```
ciyer, line 146: After changing the slug of a group, I am seeing it twice in the namespace listing (though it only appears once in the group listing).

ciyer (changes requested):
The core functionality added here works well. It is possible to create groups and associate projects with namespaces. 🚀 

There are some areas, though, where things to not work exactly as I would expect them to, and I have highlighted them in the review.

components/renku_data_services/base_api/pagination.py
leafty: This is a postgres limitation on the size of indexes, and we cannot control this type.

olevski: This is for postgres. Out automated tests on the api frequently pick very large number to test for edge cases and hit the limits on postgres. But I think we may have got the limit wrong. And it is as you suggest. I will try to get that error to show up again and see.

components/renku_data_services/namespace/apispec.py
olevski: This is automatically generated code from the apispec.yaml files that we merge from several components. In the apispec.yaml it appears in fewer places.

components/renku_data_services/base_api/pagination.py
ciyer: The limit is correct. I looked it up, Postgres is using signed integers for this type. You can ignore my comment.

components/renku_data_services/namespace/blueprints.py
ciyer: Re-tested. This seems to now cascade delete any projects in that group. Not 100% this is the behavior we want, but it is at least consistent; I will note in the dialog box that this action will delete all projects in that namespace.

components/renku_data_services/namespace/db.py
ciyer: Re-tested. This seems to now cascade delete any projects in that group. Not 100% this is the behavior we want, but it is at least consistent; I will note in the dialog box that this action will delete all projects in that namespace.

components/renku_data_services/namespace/blueprints.py
ciyer: Re-tested. This works correctly now!

components/renku_data_services/namespace/db.py
ciyer: Re-tested, this works now.

ciyer (approved):
I have tested the integration, and I think the core functionality that we intended to deliver is all here. I have created an issue with some things that we should follow-up: https://github.com/SwissDataScienceCenter/renku-data-services/issues/146

PR #150, opened 2024-03-18
feat: refactor messages for cleaner code and reliability

Panaetius:


components/renku_data_services/message_queue/orm.py
```diff
@@ -31,6 +32,8 @@ class EventORM(BaseORM):
     id: Mapped[int] = mapped_column(primary_key=True, default=None, init=False)
     """Unique id of the event."""
 
+    timestamp: Mapped[datetime] = mapped_column("timestamp", DateTime())
```
olevski, line 35: Is this UTC or not? I think it is probably better to just use UTC. Then you are sure that a different timezone did not get applied on a different deployment or in a different db/image.

components/renku_data_services/users/db.py
```diff
@@ -101,6 +106,21 @@ async def _get_users(self, email: str | None = None) -> List[UserInfo]:
...
+
+
+def create_user_updated_message(result: UserInfo, **_) -> UserUpdated:
+    """Transform user to message queue message."""
+    return UserUpdated(id=result.id, firstName=result.first_name, lastName=result.last_name, email=result.email)
+
+
+def create_user_removed_message(result, user_id: str) -> UserRemoved:
+    """Transform user removal to message queue message."""
+    return UserRemoved(id=user_id)
```
olevski, line 121: Does it cause problems (with circular imports) if you move these to the `message_queue` module? I.e. at `components/renku_data_services/message_queue`. I think that would be even better.

If we can move these, then it applies to all the other functions being passed to the decorators in the different modules.

Also if you want you can go a step further and make the specific decorators for each case in the `message_queue` module so then the specific decorator is imported and used. Rather than importing the general decorator and the specific event factory and then combining them.

It seems these were already part of the message_queue module and this PR moved them out. I think it is better if they stay in the message_queue module.

olevski (changes requested):
Thank you Ralf this looks really nice. 

I have 2 requests. See below.

olevski (approved):

olevski (approved):

components/renku_data_services/message_queue/db.py
```diff
@@ -41,7 +41,9 @@ async def send_pending_events(self) -> None:
 
         async with self.session_maker() as session:
             # we only consider events older than 5 seconds so we don't accidentally interfere with an ongoing operation
-            stmt = select(schemas.EventORM).where(schemas.EventORM.timestamp < datetime.utcnow() - timedelta(seconds=5))
+            stmt = select(schemas.EventORM).where(
+                schemas.EventORM.timestamp_utc < datetime.utcnow() - timedelta(seconds=5)
+            )
             result = await session.execute(stmt)
             events_orm = result.scalars().all()
```
leafty, line 48: Side question: why not use `scalars()` instead of `execute()`? It directly provides scalars results.

```python
result = await session.scalars(stmt)
events_orm = result.all()
```

PR #152, opened 2024-03-20
fix: keycloak user sync

olevski:
This fixes the syncing of keycloak users in the data service.

I setup iterators to do paginated requests to Keycloak and then just completely forgot to add a while loop to keep going through all possible pages. So currently only the first 20 users were imported.

I tested a job with this code on dev and it works. I use `asyncio.gather` for the user updates so if I use a nullpool then I open too many connections to postgres. That is why I added a pool with 2. I know this slows things down but this is just a job that runs periodically and it does not need to be fast. It is better for this to be slow than to nuke the db.

Once this is approved and merged. I will make a bugfix release from 0.5.0 that includes this change and then make a renku bugfix release. Currently on main in this repo we have a lot of stuff that will go out with renku 0.50.0 but this fix will be added to 0.49.x so that it can be deployed sooner than 0.50.0.

This is preventing admins from adding users to different resource pools.

### DIFF
--- a/components/renku_data_services/users/kc_api.py
+++ b/components/renku_data_services/users/kc_api.py
@@ -1,4 +1,5 @@
 """Keycloak API."""
+from copy import deepcopy
 from dataclasses import dataclass, field
 from datetime import date
 from typing import Any, Dict, Iterable, List, Protocol, cast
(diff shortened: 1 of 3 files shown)


components/renku_data_services/users/kc_api.py
```diff
@@ -76,15 +76,17 @@ def get_users(self) -> Iterable[Dict[str, Any]]:
...
-        if len(output) < self.result_per_request_limit + 1:
-            return
-        first += self.result_per_request_limit
+        while True:
+            res = self._http_client.get(url, params={**query_args, "first": first})
+            output = res.json()
+            if not isinstance(output, list):
+                raise ValueError("Received unexpected response from Keycloak for users")
+            output = cast(List[Dict[str, Any]], output)
+            yield from output[:-1]
```
Panaetius, line 85: wouldn't this ignore/not return the last entry in the list if the page isn't the last? it's not really clear from the code why this is necessary

Panaetius: Also we now have three pieces of code that are exactly the same (the while loop), just with different query args. Could be a function that is called with `yield from`

olevski: So I request one more than what I use on every page. The extra result is used to see whether I need to request the following page or I have reached the end. Because if I receive less than max_requests + 1 then I know there are no more pages available.

And I do yield the last element further below when I know that I am on the last page.

When I am iterating over pages the first element on the current page was the "+1" from the previous page.

Panaetius (approved):

PR #153, opened 2024-03-21
fix: handle no users available from keycloak during sync

olevski:
Thanks @leafty for spotting this.


leafty (approved):
At least now it deploys 😆

mana-alisafaee (approved):

PR #158, opened 2024-03-26
enable pyupgrade and fix violations

Panaetius:
a lot of these things I ignored in the past, like I thought `typing.Iterator` is a nicer import than `collections.abc.Iterator` since, well, we want to do type stuff, not use some abstract class.

But these things are actually deprecated and just aliases to collections. Not that I agree with the decision, but we shouldn't use deprecated things.

As this PR has a high potential for merge conflicts, we might want to hold off on merging it until the build stuff is merged. Though it's easy enough to fix with `ruff check --fix`, it handled like 580/582 issues automatically.


test/conftest.py
```diff
@@ -110,9 +110,7 @@ def only(iterable, default=None, too_long=None):
     except StopIteration:
         pass
     else:
-        msg = "Expected exactly one item in iterable, but got {!r}, {!r}, " "and perhaps more.".format(
-            first_value, second_value
-        )
+        msg = f"Expected exactly one item in iterable, but got {first_value!r}, {second_value!r}, " "and perhaps more."
```
olevski, line 113:
```suggestion
        msg = f"Expected exactly one item in iterable, but got {first_value!r}, {second_value!r}, and perhaps more."
```

olevski (changes requested):
Just one minor thing

olevski (approved):

PR #163, opened 2024-03-28
fix: printing of namespace outside of session

olevski:
This was causing printing the namespace orm when keycloak users are synced to crash.

I did not know we can set the repr field in sqlalchemy. I will open a followup PR to set repr to false on most relationships so that things do not blow up simply by trying to print something... as they often do now.

/deploy renku-ui=ciyer/3023-namespaces


olevski:
This was causing the data services to fail with this error:
```File "/app/env/lib/python3.11/site-packages/renku_data_services/namespace/db.py", line 46, in generate_user_namespaces
    logging.info(f"Creating user namespace {ns}")
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Because simply printing outside of a session was causing this. But now that the relationship fields are not printed we should have no issues.

Panaetius (approved):

PR #175, opened 2024-04-11
feat: add keywords to project

m-alisafaee:
Add a `keywords` field to the project entity.


components/renku_data_services/project/apispec.py
```diff
@@ -12,13 +12,13 @@
...
+class Keyword(RootModel[str]):
     root: str = Field(
         ...,
-        description="Member's KeyCloak ID",
-        example="123-keycloak-user-id-456",
+        description="A single keyword",
+        max_length=99,
         min_length=1,
-        pattern="^[A-Za-z0-9-]+$",
+        pattern="^[A-Za-z0-9\\s\\-_.]*$",
```
olevski, line 21: Why do we have to validate the keywords? We can just allow any character, right? Or do we break the search if we allow any character?

components/renku_data_services/message_queue/avro_models/io/renku/events/v1/project_updated.py
```diff
@@ -19,6 +19,7 @@ class ProjectUpdated(AvroModel):
     repositories: List[str]
     visibility: Visibility
     description: Optional[str]
+    keywords: List[str]
```
olevski, line 22:
```suggestion
    keywords: List[str] = field(default_factory=list)
```

But also you have to import field from dataclasses above.

Because the avro schema says that the default value of the field should be `[]` but the code generation tool misses this.

components/renku_data_services/message_queue/avro_models/io/renku/events/v1/project_created.py
```diff
@@ -20,6 +20,7 @@ class ProjectCreated(AvroModel):
     repositories: List[str]
     visibility: Visibility
     description: Optional[str]
+    keywords: List[str]
```
olevski, line 23:
```suggestion
    keywords: List[str] = field(default_factory=list)
```

See the other comment.

components/renku_data_services/project/api.spec.yaml
```diff
@@ -335,6 +340,8 @@ components:
           default: "private"
         description:
           $ref: "#/components/schemas/Description"
+        keywords:
+          $ref: "#/components/schemas/KeywordsList"
```
olevski, line 344:
```suggestion
          $ref: "#/components/schemas/KeywordsList"
          default: []
```

If you add this the dataclass generated from the apispec will have a deafult value of [] rather than None. But up to you if you prefer this or not. I find it easier if a list is initiated empty as default rather than None. It is easier to keep mypy happy with a single type rather than with `None | list[str]`. But it is a matter of taste.

olevski (changes requested):

components/renku_data_services/project/apispec.py
mana-alisafaee: I thought it might be good to not allow any characters here. There's no limitation on the search side. I'll remove this constraint.

components/renku_data_services/message_queue/avro_models/io/renku/events/v1/project_updated.py
mana-alisafaee: This gets overwritten we update the avro schemas (via `make update_avro`) and we need to make this change every time. Is this what you mean? I already take care of this case in the code and send an empty list when `keywords` is `null`.

olevski: Yes sorry. Lets keep it as is. I forgot it was part of the auto-generated code.

components/renku_data_services/project/apispec.py
olevski: Hey @m-alisafaee, I am sorry for going back and forth on this. I think it makes sense to make this more constrained as you originally had it. So lets keep it like you originally had it and relax it later if needed. Because we want the search to easily handle the keywords and not have to add logic for emojis or other random characters.

olevski (approved):

olevski (approved):

PR #176, opened 2024-04-12
feat: find session launchers by project namespace/slug

ciyer:
Endpoint to list project session launchers by project namespace/slug

/deploy renku-ui=ciyer/3096-merge-urls


ciyer:
This is not actually needed. Closing.

PR #178, opened 2024-04-19
feat: support for cloud storage in V2

m-alisafaee:
Adds a `/storages_v2` endpoint to works with cloud storages in Renku 2.


components/renku_data_services/storage/db.py
```diff
@@ -137,9 +138,57 @@ async def delete_storage(self, storage_id: str, user: base_models.APIUser) -> No
...
+
+    def __init__(
+        self,
+        project_authz: IProjectAuthorizer,
+        session_maker: Callable[..., AsyncSession],
+    ):
+        super().__init__(session_maker)
+        self.project_authz: IProjectAuthorizer = project_authz
+
+    async def filter_projects_by_access_level(
```
Panaetius, line 182: I think in v2 we can be smarter about this and don't have to get all projects a user has access to and match against the storage, and just get auth directly for the storages?
Though that probably depends on the authz work @olevski is doing, so maybe we need to improve this in a follow-up issue?
The main reason to get all projects instead of checking individually in v1 is that the gitlab api is slow and I'd rather just send 1 request than lots of requests, but for v2 we shouldn't have that limitation

Panaetius (commented):
this looks great, glad it was as minimal changes needed as I'd hoped.

components/renku_data_services/storage/db.py
```diff
@@ -183,12 +182,15 @@ async def filter_projects_by_access_level(
...
 
         scope = authz_models.Scope.WRITE if minimum_access_level == authz_models.Role.OWNER else authz_models.Scope.READ
+        user_project_ids = []
 
-        user_id = cast(str, user.id)
-        user_project_ids = await self.project_authz.get_user_projects(requested_by=user, user_id=user_id, scope=scope)
+        for project_id in project_ids:
+            has_permission = await self.project_authz.has_permission(user=user, project_id=project_id, scope=scope)
+            if has_permission:
+                user_project_ids.append(project_id)
```
Panaetius, line 194: this still does one query per project. Instead I'd add a new method on project_authz that does an [in](https://docs.sqlalchemy.org/en/20/core/operators.html#in-comparisons) query for all of them at once.

m-alisafaee: Thanks Ralf! We always call the `filter_projects_by_access_level` method with one project ID except in the [get_storage](https://github.com/SwissDataScienceCenter/renku-data-services/blob/1a6044abeb39cf7bf0f1d578bff7a25c38c3df24/components/renku_data_services/storage/db.py#L57) method. `get_storage` itself is always called with a single project ID although this can change in future. I've added a new method to `authz` as you've suggested in https://github.com/SwissDataScienceCenter/renku-data-services/pull/178/commits/9f1ddde2359b5637c362317f50ae1ffe6cf04c80.

Panaetius (approved):

PR #187, opened 2024-05-07
feat: send group events to message queue

m-alisafaee:


components/renku_data_services/message_queue/converters.py
```diff
@@ -280,52 +281,42 @@ class EventConverter:
...
-            return _GroupEventConverter.to_events(input, event_type)
-        else:
-            raise errors.EventError(
-                message=f"Trying to convert an unknown model of type {type(input)} to an event type {event_type}"
-            )
+
+        match event_type:
+            case v2.ProjectCreated | v2.ProjectRemoved:
+                project = cast(project_models.Project, input)
+                return _ProjectEventConverter.to_events(project, event_type)
```
olevski, line 292:
```suggestion
            case v2.ProjectCreated | v2.ProjectRemoved if isinstance(input, project_models.Project):
                return _ProjectEventConverter.to_events(input, event_type)
```
You can use a guard like that in every case so that you do not have to cast and you are sure that you have the right type.

components/renku_data_services/message_queue/converters.py
```diff
@@ -98,10 +88,11 @@ class _UserEventConverter:
     @staticmethod
     def to_events(
         user: user_models.UserInfo | user_models.UserWithNamespace | user_models.UserWithNamespaceUpdate,
-        event_type: type[AvroModel] | AmbiguousEvent,
+        event_type: type[AvroModel] | type[events.AmbiguousEvent],
     ) -> list[Event]:
         match event_type:
-            case v2.UserAdded | AmbiguousEvent.INSERT_USER_NAMESPACE if isinstance(user, user_models.UserWithNamespace):
+            case v2.UserAdded | events.InsertUserNamespace:
+                user = cast(user_models.UserWithNamespace, user)
```
olevski, line 95: I know the `if isinstance(...)` guards are ugly but with casting we completely bypass the type checks. So we could be getting something completely random but the value will still be what we cast it to according to the type checker.

olevski (changes requested):
Thanks Mohammad, with the match case statement things look much cleaner. But we should put back all the type checks because without them and with casts the type checker is essentially disabled for those variables.

m-alisafaee:
> Thanks Mohammad, with the match case statement things look much cleaner. But we should put back all the type checks because without them and with casts the type checker is essentially disabled for those variables.

Thanks for the review! The good thing with `cast` is that it doesn't have any runtime overhead. Why do we need to have type checks enabled in the runtime? We are sure that we pass the correct input type based on the event type and we have tests to cover those cases. I don't see how we can pass a wrong object type at the runtime.

olevski:
@m-alisafaee resolve the conflicts and I will approve this.

Keep in mind that the PR from Ralf was just merged which made it so that messages are sent to the queue only in a background job and the event message decorator always writes to postgres now. So in the tests if you check whether something is in redis it will not be there. You can see how the tests were modified by Ralf here: https://github.com/SwissDataScienceCenter/renku-data-services/pull/217

olevski (approved):

PR #188, opened 2024-05-08
feat: support GitHub

leafty:
WIP

/deploy #notest renku-ui=build/connected-services-final renku-notebooks=leafty/build-connected-services

### DIFF
--- a/components/renku_data_services/connected_services/orm.py
+++ b/components/renku_data_services/connected_services/orm.py
@@ -68,6 +68,8 @@ class OAuth2ClientORM(BaseORM):
         """The authorization URL for the OAuth2 protocol."""
         if not self.url:
             raise errors.ValidationError(message=f"URL not defined for provider {self.id}.")
+        if self.kind == ProviderKind.github:
+            return urljoin(self.url, "login/oauth/authorize")
         return urljoin(self.url, "oauth/authorize")
 
     @property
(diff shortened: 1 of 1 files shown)


components/renku_data_services/connected_services/db.py
```diff
@@ -272,14 +272,25 @@ async def get_oauth2_connected_account(
...
-            response = await oauth2_client.get(request_url)
+            request_url = urljoin(client.api_url, "user")
+            headers = (
+                {
+                    "Accept": "application/vnd.github+json",
+                    "X-GitHub-Api-Version": "2022-11-28",
+                }
+                if client.kind == ProviderKind.github
+                else None
+            )
```
Panaetius, line 283: it'd be really nice if we had different subclasses of client for the different providers and here we could have `headers=client.headers`, and below (line 289) like `account=client.get_connected_account(...)`. That way we can keep provider specific logic more in one place per provider. Also for methods on `OAuth2ClientORM` directly.

It is an ORM and polymorphism on the database level sucks, so I wouldn't go for something like https://docs.sqlalchemy.org/en/20/orm/inheritance.html#single-table-inheritance. Maybe just have classes that wrap the ORM? Or maybe there's another way to conditionally instantiate query results in sqlalchemy.

It would prevent a lot of if-else statements that probably will grow in the future.

leafty: Actually done in 14fdf725e1503177a7e4d7585219911ca072a8be.

Panaetius (approved):
nice!

PR #198, opened 2024-05-13
refactor: use type state pattern for DB ids

Panaetius:


components/renku_data_services/base_models/core.py
```diff
@@ -21,6 +21,44 @@ async def authenticate(self, access_token: str, request: Request) -> "APIUser":
...
+#     """A descriptor for wrapping non-existing ids in class attributes.
+
+#     Fails on access.
+#     """
+
+#     def __get__(self, obj, objtype=None):
+#         raise NotImplementedError("Cannot get an id of NoId")
+
+#     def __set__(self, obj, value):
+#         raise NotImplementedError("Cant set NoId, use dedicated 'set_id' method instead")
```
olevski, line 53: I think all of this can be removed, right @Panaetius?

Panaetius: ah yes, that was an earlier attempt using descriptors, but mypy doesn't understand them properly and it's simple with typevar anyways

components/renku_data_services/project/models.py
```diff
@@ -35,7 +34,19 @@ def etag(self) -> str | None:
         return compute_etag_from_timestamp(self.updated_at)
 
 
-ProjectsType: TypeAlias = list[Project]
+@dataclass(frozen=True, eq=True, kw_only=True)
+class Project(BaseProject):
+    """Base Project model."""
+
+    id: base_models.Id
```
olevski, line 41: I think now this can be only a string. We dont really need the Id type.

olevski (changes requested):
I am not sure how I feel about the distinction we have now just to sidestep having the Namespace in a project. But we can leave it in for now. I will try to think of a way where we can maybe have one class where the actual `Namespace` object can be fetched from the db if required.

There is just one minor change now.

PR #216, opened 2024-05-22
feat: support PKCE and store next_url in the database

### LINKED ISSUE #193: Add support for PKCE for connected services
Currently PKCE is not supported but the authlib library we use should easily support PKCE.

This is the library docs page: https://docs.authlib.org/en/latest/index.html

And I think that the way to use PKCE is described here: https://docs.authlib.org/en/latest/client/oauth2.html#add-pkce-for-authorization-code

### LINKED ISSUE #210: Store `next_url` in the database during the OAuth2 authorization flow
GitHub wants an exact match between the configured redirect URL and the request parameter used during the authorization flow. This means that we should avoid sending a redirect URL which contains query parameters and instead store `next_url` in the database.

Example configured callback URL:
```
https://ci-renku-3620.dev.renku.ch/api/data/oauth2/callback?next_url=https%3A%2F%2Fci-renku-3620.dev.renku.ch%2Fv2%2Fconnected-services
```

Solving this may also make PKCE work (as we need to save the PKCE code as well). #193

leafty:
(Stacked PR)

Fixes #193, fixes #210.

Add PKCE support and store `next_url` in the database.

* PKCE support is indicated by a new field, `use_pkce`, added to OAuth2 Clients. Note that only GitLab providers can support PKCE for now.
* Stores the `next_url` parameter in the database instead of passing it around as a query parameter. This allows final redirects to be customized for GitHub which does not allow extra query parameters.

/deploy #notest renku=release-0.52.x


components/renku_data_services/connected_services/orm.py
```diff
@@ -77,6 +79,7 @@ class OAuth2ConnectionORM(BaseORM):
     token: Mapped[dict[str, Any] | None] = mapped_column("token", JSONVariant)
     state: Mapped[str | None] = mapped_column("state", String(), index=True, unique=True)
     status: Mapped[ConnectionStatus]
+    extra_data: Mapped[dict[str, Any] | None] = mapped_column("extra_data", JSONVariant)
```
olevski, line 82: I think it is better to not have such general fields in the DB and just have a field specifically for the next_url. Especially when it comes to OAuth. Because it increases the risk that someone would insert something in this dictionary that should not be shared or used in a specific way but we would not know. And the name `extra_data` does not really communicate what the purpose of this data is. 

Also we store the code verifier in this `extra_data` field. But the code verifier is supposed to be used only for a single authentication request and with this setup it could happen that 2 different requests conflict with one another or one request overwrites an existing code verifier of a authentication that is in progress. In a lot of cases the code verifier is just stored in an encypted cookie. And I was hoping that the auth library will handle this for you.

components/renku_data_services/connected_services/db.py
```diff
@@ -213,33 +225,36 @@ async def authorize_callback(
...
+                )
 
                 logger.info(f"Token for client {client.id} has keys: {", ".join(token.keys())}")
 
                 connection.token = self._encrypt_token_set(token=token, user_id=connection.user_id)
                 connection.state = None
                 connection.status = schemas.ConnectionStatus.connected
 
-                return token
+                extra_data = connection.extra_data or dict()
```
olevski, line 256: I think you should remove the authentication-request specific data from the database here. I.e. the next_url and the code_verifier. What happens if the current request did not set a value for `next_url`? You will end up reusing the `next_url` value from the previous authentication request, right?

olevski (changes requested):
Thanks for adding this Flora.

My biggest concern is that we are saving things in the DB (like the code verifier and next_url) that are supposed to be unique for each authentication request.

I think here we have 2 options:
1. put these things in an encrypted cookie and do not store them in the DB
2. make a separate table keyed by the state parameter from the oauth flow and store these things there, but you have to make sure you delete things that are old and also delete things when the oauth flow completes
3. store things in the same table you have now but only allow one single authentication request to be in progress at a time, for this you can store the state parameter and fail if you get a request with a different state parameter, also in this case too you will have to cleanup/reset fields on a successful completion of the oauth flow

components/renku_data_services/connected_services/api.spec.yaml
```diff
@@ -329,6 +337,11 @@ components:
         The base URL of the OAuth2 Resource Server, e.g. "https://gitlab.com".
       type: string
       example: "https://example.org"
+    UsePKCE:
+      description: |
+        Whether or not to use PKCE during authorization flows.
+      type: boolean
+      example: false
```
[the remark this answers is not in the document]
leafty: Is this really necessary? There are no clients yet, so an API break will not change anything.

leafty:
> Thanks for adding this Flora.
> 
> My biggest concern is that we are saving things in the DB (like the code verifier and next_url) that are supposed to be unique for each authentication request.
> 
> I think here we have several options:
> 
> 1. put these things in an encrypted cookie and do not store them in the DB
> 2. make a separate table keyed by the state parameter from the oauth flow and store these things there, but you have to make sure you delete things that are old and also delete things when the oauth flow completes
> 3. store things in the same table you have now but only allow one single authentication request to be in progress at a time, for this you can store the state parameter and fail if you get a request with a different state parameter, also in this case too you will have to cleanup/reset fields on a successful completion of the oauth flow

Number 3 is what already happens now. There is a uniqueness constraint on having a single connection per user per provider. The state is also used as an indexed key for the callback.

I can change the db fields and store the PKCE code with encryption.

olevski:
> Number 3 is what already happens now. There is a uniqueness constraint on having a single connection per user per provider. The state is also used as an indexed key for the callback.

Ah I see. I missed this earlier. 

> store the PKCE code with encryption.

You mean store the `code_verifier` value encrypted? I dont think there is a need for this.

> I  can change the db fields

Yes it would be nice to have separate fields for `next_url` and the pkce `code verifier` in the db.

components/renku_data_services/connected_services/api.spec.yaml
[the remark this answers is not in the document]
leafty: Done in 306d369fe58d254cab9c4ae3d5604bcac6750cd4.

components/renku_data_services/connected_services/orm.py
leafty: Note that `authlib` does not handle client sessions, so it will not save cookies or the like.

Done in a10a435f31628c9df8758ff1fed407bcb5788858.

olevski (approved):

PR #222, opened 2024-05-23
chore: build secrets storage image with github actions

olevski:
I think for all other releases we published the image manually. I will also publish it manually for 0.12.0 which I tagged an hour ago.


Panaetius (approved):
It wasn't manual, I think it just happened through chartpress

PR #225, opened 2024-05-24
feat: configure trusted reverse proxies

### LINKED ISSUE #224: Reverse proxy config is missing in production and CI/CD deployments
See: https://sanic.dev/en/guide/advanced/proxy-headers.md

`request.scheme` returns `"http"` instead of `"https"`, most likely because the server is not configured to trust reverse proxies.

leafty:
Fixes #224.

WIP

### DIFF
--- a/bases/renku_data_services/data_api/main.py
+++ b/bases/renku_data_services/data_api/main.py
@@ -95,6 +95,12 @@ def create_app() -> Sanic:
         validator = RCloneValidator()
         app.ext.dependency(validator)
 
+    @app.after_server_start
+    async def after_server_start(app: Sanic):
+        logger.info("Printing all routes by name")
+        for key in app.router.name_index.keys():
+            logger.info(f"Route: {key}")
+
     async def send_pending_events(app):
         """Send pending messages in case sending in a handler failed."""
         while True:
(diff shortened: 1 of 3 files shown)


olevski:
Also shouldn't the proxy count be 2? Because in most cases the ui will call this through the ui-server, which will then go through the gateway and finally reach the data service.

bases/renku_data_services/data_api/main.py
```diff
@@ -66,6 +66,15 @@ async def setup_sentry(_):
         app.signal("http.lifecycle.request")(_hub_enter)
         app.signal("http.lifecycle.response")(_hub_exit)
         app.signal("http.routing.after")(_set_transaction)
+    if config.trusted_proxies.forwarded_secret:
+        app.config.FORWARDED_SECRET = config.trusted_proxies.forwarded_secret
+        logger.info(f"FORWARDED_SECRET = {app.config.FORWARDED_SECRET}")
```
olevski, line 71: Lets not print the secret in the logs. You can replace it with just a message that a secret was found and will be used but don't print the secret.

olevski (changes requested):

bases/renku_data_services/data_api/main.py
leafty: I will remove this setting for now.

leafty: Done in fd7a7abbda27c0c6e6b44b83bc999fe92ec4fe9c.

PR #228, opened 2024-05-28
chore: fix formatting and add types everywhere

Panaetius:
also sets mypy to be stricter and makes CI actually check ruff formatting.

I applied mypy typing rules incrementally in different commits, so we can also undo some of the stricter changes if needed. This enables most of the checks [here](https://mypy.readthedocs.io/en/stable/config_file.html#disallow-dynamic-typing), except `disallow_any_decorated` which causes around 250 errors, as e.g. `def x(cls, ...)` counts `cls` as unannotated and so decorating it fails.

note: in some places I annotate kwargs with dict, sometimes Any. I try to use the former unless mypy complains.

I realized halfway through this that we can just use ParamSpec and do like `kwargs['x']=x` instead of making arguments positional as with APIUser but I don't want to go through everything again and change it, and having APIUser as second parameter after Request isn't that bad.


olevski (dismissed):

olevski:
You will have to resolve conflicts. But let me know and I will re-approve.

olevski (approved):

PR #230, opened 2024-05-29
chore: add more tests for migrations

leafty:
NOTE: waiting until the `authzed` migrations are fixed before revisiting this.

Details:

* Add a test to check if the Alembic migrations have a unique head
* Add one upgrade/downgrade cycle to the test

Note that the additional upgrade/downgrade cycle catches this issue at the moment:
```
grpc._channel._InactiveRpcError: <_InactiveRpcError of RPC that terminated with:
        status = StatusCode.INVALID_ARGUMENT
        details = "cannot delete object definition `user_namespace`, as a relationship exists under it"
        debug_error_string = "UNKNOWN:Error received from peer ipv4:127.0.0.1:50051 {grpc_message:"cannot delete object definition `user_namespace`, as a relationship exists under it", grpc_status:3, created_time:"2024-05-29T09:05:15.715134445+00:00"}"
```


components/renku_data_services/migrations/versions/ce1a111d6694_authz_add_intial_schema.py
```diff
@@ -23,6 +24,15 @@
...
     config = AuthzConfig.from_env()
     client = config.authz_client()
+
+    try:
+        existing_schema = client.ReadSchema(ReadSchemaRequest())
+        if "definition user" in existing_schema.schema_text:
+            logging.info(f"Skipping {revision}, as Authz is likely already populated.")
+            return
+    except RpcError:
+        pass
```
olevski, line 34: Is this to catch the issue you mention in the PR description?

The problem is that this will also catch other issues that are unrelated but potentially important.

The proper way to address the problem in the PR description is to delete all relationships that reference user_namespace.

This can hide errors that are important.

olevski (changes requested):

components/renku_data_services/migrations/versions/ce1a111d6694_authz_add_intial_schema.py
leafty: I you would like, we can delay this until the migrations have a proper fix. I do not know how upgrade/downgrades of the authzed schema need to be handled.

leafty:
Obsoleted by #240.

PR #234, opened 2024-05-29
fix: allow re-adding existing users to resource pools

### LINKED ISSUE #202: Re-adding an existing user to a resource pool should not result in 500 status code
image

m-alisafaee:
Fixes #202


components/renku_data_services/crc/db.py
```diff
@@ -695,6 +695,8 @@ async def update_user_resource_pools(
                         message=f"User with keycloak id {keycloak_id} cannot access the default resource pool"
                     )
             if append:
+                user_rp_ids = [rp.id for rp in user.resource_pools]
```
Panaetius, line 698: use a set for better time complexity

components/renku_data_services/crc/db.py
```diff
@@ -760,7 +762,9 @@ async def update_resource_pool_users(
                 schemas.RPUserORM(keycloak_id=user_id) for user_id in user_ids if user_id not in user_ids_to_add_exist
             ]
             if append:
-                rp.users.extend(list(users_to_add_exist) + users_to_add_missing)
+                rp_user_ids = [rp.id for rp in rp.users]
```
Panaetius, line 765: also use a set

Panaetius (changes requested):
thanks for looking into this! Just have two minor performance considerations.

olevski (approved):

Panaetius (approved):

PR #236, opened 2024-05-30
fix: dont populate user namespaces in authz in migrations

olevski:
Fixes the long-running migrations by moving the data loading in the background jobs.


bases/renku_data_services/background_jobs/main.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+async def main() -> None:
+    """Synchronize data from Keycloak and the user database."""
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "tasks",
+        nargs="+",
+        choices=["keycloak_events_sync", "keycloak_users_sync", "keycloak_admins_sync", "user_namespaces_bootstrap"],
+    )
```
Panaetius, line 23: I think we should use subparsers for different actions. That allows to have different arguments if we ever need it, is also exhaustive and automatically handles errors. It also means we don't need a match statement but can just have it be functions that get called directly.

olevski:
Deployed over at https://github.com/SwissDataScienceCenter/renku/pull/3649

bases/renku_data_services/background_jobs/utils.py
Panaetius, line 1: The code in here is not really utility code but the core business logic. We still don't have a consistent place to put business logic in data services, I'm working on something else at the moment where I put it in `core.py` but am also not that happy with that naming (plus polylith creates core.py by default, so high chance of confusion...).
Not sure if you have a better idea? But `utils` feels wrong.

bases/renku_data_services/background_jobs/utils.py
```diff
@@ -0,0 +1,60 @@
...
+from renku_data_services.authz.authz import Authz, ResourceType, _Relation
+from renku_data_services.background_jobs.config import SyncConfig
+from renku_data_services.message_queue.avro_models.io.renku.events import v2
+from renku_data_services.message_queue.converters import EventConverter
+
+
+async def sync_user_namespaces(config: SyncConfig) -> None:
+    """Lists all user namespaces in the database and adds them to Authzed and the event queue."""
+    authz = Authz(config.authz_config)
+    user_namespaces = await config.group_repo._get_user_namespaces()
```
Panaetius, line 16: I think it'd be good to get namespaces in batches (You can have an AsyncGenerator that yields batches of namespaces and use it with `async for ...`) otherwise this might have a pretty big result set as time goes on.

bases/renku_data_services/background_jobs/utils.py
```diff
@@ -0,0 +1,60 @@
...
+        session = config.session_maker()
+        tx = session.begin()
+        await tx.start()
+        try:
+            await authz.client.WriteRelationships(authz_change.apply)
+            for event in events:
+                await config.event_repo.store_event(session, event)
+        except Exception as err:
+            logging.error(f"Failed to sync user namespace {user_namespace} because {err}")
+            await tx.rollback()
```
Panaetius, line 30: do we also need to try to authz_change.undo here?

bases/renku_data_services/background_jobs/utils.py
```diff
@@ -0,0 +1,60 @@
...
+        else:
+            await tx.commit()
+        finally:
+            await session.close()
+    logging.info(f"Wrote authorization changes and events for {len(user_namespaces)} user namespaces")
+
+
+async def bootstrap_user_namespaces(config: SyncConfig) -> None:
+    """Sycnhornize user namespaces to the authorization database only if none are already present."""
```
Panaetius, line 39: spelling

components/renku_data_services/authz/authz.py
```diff
@@ -620,6 +616,10 @@ async def _remove_project(
         zed_token: ZedToken | None = None,
     ) -> _AuthzChange:
         """Remove the relationships associated with the project."""
+        if not project.id:
```
Panaetius, line 619: A bit confused by this (and the other occurrences below), this can't be None with the changes I did?

components/renku_data_services/authz/authz.py
```diff
@@ -1055,6 +1063,8 @@ async def _remove_group(
         zed_token: ZedToken | None = None,
     ) -> _AuthzChange:
         """Remove the group from the authorization database."""
+        if not group.id:
+            raise errors.ProgrammingError(message="Cannot a group in the authorization database if the group has no ID")
```
Panaetius, line 1067: You accidentally a verb :slightly_smiling_face:

Panaetius (changes requested):

components/renku_data_services/authz/authz.py
olevski: I started the PR from before the change you made was merged. Now they are not needed. I removed them.

bases/renku_data_services/background_jobs/utils.py
olevski: I do not want to do that because if the authz update succeeded I will leave it in. It is acceptable to have something in the authz DB but not in the message queue but not vice versa.

bases/renku_data_services/background_jobs/utils.py
olevski: So I used a stream from sqlalchemy which is supposed to gradually stream the results from the server. And I used async for. So there is no need to batch this - the results will come gradually.

bases/renku_data_services/background_jobs/utils.py
olevski: As discussed I renamed to `core.py`.

Panaetius (approved):

Panaetius (approved):

Issue #227, opened 2024-05-27
Prevent the last owner from leaving a resource


leafty:
The last owner of a group is allowed to change their role within the group to be `editor` or `viewer`, leaving the group without any owner.

Any owned resource should always have at least one owner and invalid updates should be rejected. Owned resources include for now `Group` and `Project` but may include more or new entity types in the future.

PR #237, opened 2024-05-31
fix: refuse to leave resources without an owner when changing roles

mohammad-alisafaee:
Fixes: #227


components/renku_data_services/authz/authz.py
```diff
@@ -831,6 +851,19 @@ async def upsert_project_members(
                 # The existing relationships should be deleted if all goes well and added back in if we have to undo
                 existing_rel = existing_rels[0]
                 if existing_rel.relationship != rel:
+                    if existing_rel.relationship.relation == _Relation.owner.value:
+                        if existing_owners_rels is None:
+                            existing_owners_rels = await self._get_resource_owners(
+                                resource_type, resource_id, consistency
+                            )
+                        if len(existing_owners_rels) == 1:
```
Panaetius, line 859: I think there's still an edge case here if there's two or more owners and all of them get unset in the same request. This code wouldn't fail but you'd still end up with no owners.

components/renku_data_services/authz/authz.py
```diff
@@ -1110,16 +1131,44 @@ async def upsert_group_members(
...
-                        RelationshipUpdate(
-                            operation=RelationshipUpdate.OPERATION_TOUCH,
-                            relationship=rel,
-                        ),
+                    if existing_rel.relationship.relation == _Relation.owner.value:
+                        if existing_owners_rels is None:
+                            existing_owners_rels = await self._get_resource_owners(
+                                resource_type, resource_id, consistency
+                            )
+                        if len(existing_owners_rels) == 1:
```
Panaetius, line 1139: this has the same edge case as mentioned above

Panaetius (changes requested):
thank you for looking into this!
But I think there's still another edge case  (albeit a very rare one) when more than one owner is changed at a time.

Panaetius (approved):

PR #242, opened 2024-06-05
fix: quote sentry types

Panaetius:
/deploy extra-values=dataService.sentry.enabled=true,dataService.sentry.dsn=https://e47c95d665684bce89984bd4595268df@sentry.dev.renku.ch/13,dataService.sentry.environment=ci-renku-3548,dataService.sentry.sampleRate=0.5

### DIFF
--- a/bases/renku_data_services/data_api/main.py
+++ b/bases/renku_data_services/data_api/main.py
@@ -71,8 +71,8 @@ def create_app() -> Sanic:
         logger.info("enabling sentry")
 
         def filter_error(
-            event: sentry_sdk._types.Event, hint: sentry_sdk._types.Hint
-        ) -> sentry_sdk._types.Event | None:
+            event: "sentry_sdk._types.Event", hint: "sentry_sdk._types.Hint"
+        ) -> "sentry_sdk._types.Event | None":
             if "exc_info" in hint:
                 exc_type, exc_value, tb = hint["exc_info"]
                 if isinstance(


bases/renku_data_services/data_api/main.py
```diff
@@ -71,8 +71,8 @@ def create_app() -> Sanic:
         logger.info("enabling sentry")
 
         def filter_error(
-            event: sentry_sdk._types.Event, hint: sentry_sdk._types.Hint
-        ) -> sentry_sdk._types.Event | None:
+            event: "sentry_sdk._types.Event", hint: "sentry_sdk._types.Hint"
+        ) -> "sentry_sdk._types.Event | None":
```
leafty, line 75: No need to quote `None`:

```suggestion
        ) -> "sentry_sdk._types.Event" | None:
```

Panaetius: that is not valid code. 
[The | operand cannot be used at runtime to define unions where one or more members is a forward reference. For example, int | "Foo", where "Foo" is a reference to a class not yet defined, will fail at runtime. For unions which include forward references, present the whole expression as a string, e.g. "int | Foo". ](https://docs.python.org/3/library/stdtypes.html#union-type)

leafty (approved):
Service is running with Sentry enabled. :+1:

olevski (approved):

PR #247, opened 2024-06-10
feat: add SecretKind

m-alisafaee:


components/renku_data_services/users/api.spec.yaml
```diff
@@ -356,6 +369,12 @@ components:
       type: string
       minLength: 1
       maxLength: 5000
+    SecretKind:
+      description: Kind of secret
+      type: string
+      enum:
+        - "general"
+        - "storage"
```
ciyer, line 377: In my experience, enums are usually in all caps:

```suggestion
        - "GENERAL"
        - "STORAGE"
```

components/renku_data_services/users/apispec.py
```diff
@@ -22,6 +23,11 @@ class Version(BaseAPISpec):
     version: str
 
 
+class SecretKind(Enum):
+    general = "general"
+    storage = "storage"
```
ciyer, line 28: In my experience, enums are usually in all caps:

```suggestion
    GENERAL = "GENERAL"
    STORAGE = "STORAGE"
```

ciyer (commented):
This looks good! I have a small question about the convention for writing enums.

components/renku_data_services/users/api.spec.yaml
ciyer: Clarified -- in this repo lowercase is used for enums.

components/renku_data_services/users/apispec.py
ciyer: Clarified -- in this repo lowercase is used for enums.

components/renku_data_services/secrets/models.py
```diff
@@ -15,6 +17,7 @@ class Secret(BaseModel):
     encrypted_key: bytes = Field(repr=False)
     id: str | None = Field(default=None, init=False)
     modification_date: datetime = Field(default_factory=lambda: datetime.now(UTC).replace(microsecond=0), init=False)
+    kind: SecretKind
```
olevski, line 20: suggestion(blocking): The base model should not use/depend on auto-generated code from the apispec. We should define an identical enum here and use it in the model. Also we should use this model enum in all other components like the blueprints. Let the autogenerated enum just be for serialization/deserialization. The problem with the autogenerated enum is that when it changes it will be automatically regenerated so you may get surprised with new fields or more drastic changes you dont expect. Also since the apispec stuff and migrations are autogenerated it can happen that you autogenerate a new enum from an apispec and also autogenerate a migration that is not fully expected/intended.

I know we do this in a few places already. I plan to go in and change all occurrences like what I described above during cooldown.

components/renku_data_services/secrets/orm.py
```diff
@@ -34,14 +35,17 @@ class SecretORM(BaseORM):
     encrypted_value: Mapped[bytes] = mapped_column(LargeBinary())
     encrypted_key: Mapped[bytes] = mapped_column(LargeBinary())
     modification_date: Mapped[datetime] = mapped_column("modification_date", DateTime(timezone=True))
+    kind: Mapped[SecretKind]
```
olevski, line 38: Same as the other comment, use the enum you define from the model.

components/renku_data_services/users/blueprints.py
```diff
@@ -168,7 +183,12 @@ def post(self) -> BlueprintFactoryResponse:
         @validate(json=apispec.SecretPost)
         async def _post(_: Request, user: base_models.APIUser, body: apispec.SecretPost) -> JSONResponse:
             encrypted_value, encrypted_key = await self._encrypt_user_secret(requested_by=user, secret_value=body.value)
-            secret = Secret(name=body.name, encrypted_value=encrypted_value, encrypted_key=encrypted_key)
+            secret = Secret(
+                name=body.name,
+                encrypted_value=encrypted_value,
+                encrypted_key=encrypted_key,
+                kind=body.kind,
+            )
```
olevski, line 191: Use model enum here, not enum from apispec.

olevski (changes requested):
Thanks Mohammad. Just a few minor changes about using the auto-generated apispec enums in models/db.

components/renku_data_services/secrets/models.py
m-alisafaee: I've added an enum to models. However, we cannot pass the `Secret` model to apispec `SecretWithId` model's `model_validate` anymore when returning results from various endpoints (since now they use different `SecretKind` enums). To address this, I call `Secret.model_dump_json` on the secret model. I then don't call the `apispec.SecretWithId.model_validate` on the apispec model since imo we already have a valid json result and don't need to validate it (and dump it to json) again. I can put back this in, if you think it's required.

olevski (dismissed):
Thank you Mohammad! I will re-approve because due to me delaying things on approving in the first place now you have conflicts.

Please ping me to reapprove when you resolve the conflicts.

olevski (approved):

PR #249, opened 2024-06-10
feat: add resource class id in session launcher

andre-code:
PR to support resource class in session launcher.

/deploy


components/renku_data_services/session/orm.py
```diff
@@ -113,6 +114,15 @@ class SessionLauncherORM(BaseORM):
     )
     """Id of the session environment."""
 
+    resource_class_id: Mapped[int | None] = mapped_column(
+        "resource_class_id",
+        ForeignKey(ResourceClassORM.id, ondelete="SET NULL"),
+        default=None,
+        nullable=True,
+        index=True,
```
leafty, line 122: Do not index this field. It is not used and consumes resources.

leafty (changes requested):

components/renku_data_services/session/db.py
```diff
@@ -208,6 +209,17 @@ async def insert_launcher(
...
+            if resource_class_id is not None:
+                res = await session.scalars(
+                    select(schemas.ResourceClassORM).where(schemas.ResourceClassORM.id == resource_class_id)
+                )
+                resource_class = res.one_or_none()
+                if resource_class is None:
+                    raise errors.MissingResourceError(
+                        message=f"Resource class with id '{resource_class_id}' does not exist or you do not have access to it."  # noqa: E501
+                    )
+
```
Panaetius, line 222: issue: we should use `get_classes` with the `user` here. doing DB queries in the session db on things owned by the crc DB doesn't separate concerns properly and despite the error saying `or you do not have access to it`, this does not actually check permissions. 
So here we should call ResourcePoolRepository.get_classes. For this to work, we need to inject the ResourcePoolRepository into the SessionRepository. similar to line 47 for project_authz.
To be clear, this is not something you did wrong, but something this whole file is guilty of and that we need to address in the future. But at least for new code we should do it properly.

components/renku_data_services/session/db.py
```diff
@@ -249,14 +261,27 @@ async def update_launcher(
...
+            resource_class_id = kwargs.get("resource_class_id")
+            if resource_class_id is not None:
+                res = await session.scalars(
+                    select(schemas.ResourceClassORM).where(schemas.ResourceClassORM.id == resource_class_id)
+                )
+                resource_class = res.one_or_none()
+                if resource_class is None:
+                    raise errors.MissingResourceError(
+                        message=f"Resource class with id '{resource_class_id}' does not exist or you do not have access to it."  # noqa: E501
+                    )
```
Panaetius, line 273: issue: same here, we need to properly validate that a user actually has access to the class.

components/renku_data_services/session/orm.py
```diff
@@ -139,6 +150,7 @@ def dump(self) -> models.SessionLauncher:
             description=self.description,
             environment_kind=self.environment_kind,
             environment_id=self.environment_id if self.environment_id is not None else None,
+            resource_class_id=self.resource_class_id if self.resource_class_id is not None else None,
```
Panaetius, line 153: nitpick: I don't think this if actually does something, same for the line above. we can just assign directly.

test/bases/renku_data_services/data_api/test_sessions.py
```diff
@@ -277,6 +277,7 @@ async def test_post_session_launcher(sanic_client: SanicASGITestClient, user_hea
     assert res.json.get("environment_kind") == "container_image"
     assert res.json.get("container_image") == "some_image:some_tag"
     assert res.json.get("environment_id") is None
+    assert res.json.get("resource_class_id") is None
```
Panaetius, line 280: it would be nice to test that setting this to a value also works, though that's a bit more difficult as we don't have a fixture to have a pool/class available.

It'd be nice to have something like https://github.com/SwissDataScienceCenter/renku-data-services/blob/main/test/bases/renku_data_services/data_api/test_resource_pools.py#L68-L74 as a fixture that you can depend on. Let me know if you need help with that.

Panaetius (changes requested):
Thank you, this is really nice. I have some comments, but that's mostly things that were already wrong before you started, so nothing you did wrong and we need to clean this up during cooldown. But we should still address them in the newly added code at least. Ping me if there's questions of I can help :slightly_smiling_face:

components/renku_data_services/crc/blueprints.py
```diff
@@ -318,7 +318,7 @@ def get_no_pool(self) -> BlueprintFactoryResponse:
         @authenticate(self.authenticator)
         @validate_db_ids
         async def _get_no_pool(_: Request, user: base_models.APIUser, class_id: int) -> HTTPResponse:
-            res = await self.repo.get_classes(api_user=user, id=class_id)
+            res = await self.repo.get_classes(api_user=None, id=class_id)
```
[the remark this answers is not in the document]
andre-code: Sure, included in the last commit

components/renku_data_services/session/db.py
andre-code: Sure, included in the latest commit.

test/bases/renku_data_services/data_api/test_sessions.py
andre-code: sure, I added the fixture to create a resource pool and the test to validate if the user has access to the resource class when creating or updating a launcher.

andre-code:
Thank you, @Panaetius and @leafty, for your review. I have incorporated your suggestions, and the changes are ready for your review again.

Panaetius (changes requested):

components/renku_data_services/migrations/versions/57dfd69ea814_add_resource_class_in_session_launcher.py
```diff
@@ -0,0 +1,41 @@
...
+down_revision = "c0631477aea4"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column("launchers", sa.Column("resource_class_id", sa.Integer(), nullable=True), schema="sessions")
+
+    # Recreate the foreign key with ON DELETE SET NULL
```
leafty, line 23: non-blocking: I think this comment can be removed here.

```suggestion
```

Panaetius (approved):

leafty (approved):

PR #252, opened 2024-06-11
fix: ignore empty previous secret key in secrets storage

Panaetius:
I had previously tested this (it's how I got a test deployment in the first place), but then later refactored parts of it. Turns out that when you mount a secret subpath that doesn't exist, k8s will just create an empty file for you.
The code should error if a secret can't be parsed, but this is a special edge case so an empty secret is ok. We can't really check if the secret is empty and handle this in the chart since it secrets get created by an init job and we don't know in the deploy stage.


Panaetius:
not actually needed

PR #263, opened 2024-06-18
chore: set up dependabot depedency updates

Panaetius:
note: multi-directory support is beta and not in the documentation other than https://github.blog/changelog/2024-04-29-dependabot-multi-directory-configuration-public-beta-now-available/

this updates poetry dependencies for all pyproject.toml's grouped together, on every sunday (when CI is unlikely to be running).

We could also have separate PRs for dev and prod dependencies.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,16 @@
+version: 2
+updates:
+- package-ecosystem: "pip"
+  directories:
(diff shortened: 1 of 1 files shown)


m-alisafaee (approved):
Thank you!

m-alisafaee (approved):

Issue #267, opened 2024-06-19
Logging is broken in `background_jobs`


leafty:
The `background_jobs` containers produce no logs.

Root cause: The containers are initialized with `logging.basicConfig(level=logging.INFO)` but this is overridden by Alembic when we call `run_migrations_for_app()`.

See configuration in `alembic.ini`:
https://github.com/SwissDataScienceCenter/renku-data-services/blob/8634341630624a75e22d4ff60bc63b3a98e3dab9/components/renku_data_services/migrations/alembic.ini#L89-L92

This means after migrations are run, only `WARN` and above messages are logged.

olevski:
Fixed in #277 by setting the level in alembic.ini to `INFO`.

PR #277, opened 2024-06-21
fix: properly cascade group permissions to all projects in the group

### LINKED ISSUE #274: Group membership does not cascade permissions to projects
To replicate:
1. create a group as user1
2. create a private project as user1 in the group from step 1
3. Add user2 to group from step1
4. Login as user2
5. Try to access the project from step 2 and you will not be able to, but you should

olevski:
Fixes #274 

/deploy

### DIFF
--- a/bases/renku_data_services/background_jobs/core.py
+++ b/bases/renku_data_services/background_jobs/core.py
@@ -2,12 +2,22 @@
 
 import logging
 
-from authzed.api.v1.permission_service_pb2 import ReadRelationshipsRequest, RelationshipFilter
+from authzed.api.v1.core_pb2 import ObjectReference, Relationship, RelationshipUpdate, SubjectReference
+from authzed.api.v1.permission_service_pb2 import (
+    Consistency,
+    ReadRelationshipsRequest,
+    RelationshipFilter,
+    SubjectFilter,
+    WriteRelationshipsRequest,
+)
 
 from renku_data_services.authz.authz import Authz, ResourceType, _Relation
 from renku_data_services.background_jobs.config import SyncConfig
+from renku_data_services.base_api.pagination import PaginationRequest
+from renku_data_services.base_models.core import InternalServiceAdmin, ServiceAdminId
 from renku_data_services.message_queue.avro_models.io.renku.events import v2
 from renku_data_services.message_queue.converters import EventConverter
+from renku_data_services.namespace.models import NamespaceKind
 
 
 async def sync_user_namespaces(config: SyncConfig) -> None:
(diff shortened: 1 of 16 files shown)


bases/renku_data_services/background_jobs/core.py
```diff
@@ -67,3 +76,54 @@ async def bootstrap_user_namespaces(config: SyncConfig) -> None:
...
+        project = await config.project_repo.get_project(api_user, project_id)
+        if project.namespace.kind != NamespaceKind.group:
+            continue
+        correct_group_id = project.namespace.underlying_resource_id
+        authzed_group_id = rel.relationship.subject.object.object_id
+        if authzed_group_id != correct_group_id:
+            logging.info(
+                f"The project namespace ID in Authzed {authzed_group_id} "
+                f"does not match the expected group ID {correct_group_id}, correcting it..."
+            )
```
leafty, line 107: Logging doesn't work here btw (see #267).

components/renku_data_services/namespace/blueprints.py
```diff
@@ -163,7 +163,7 @@ async def _get_namespaces(
                     name=ns.name,
                     slug=ns.latest_slug if ns.latest_slug else ns.slug,
                     created_by=ns.created_by,
-                    creation_date=ns.creation_date,
+                    creation_date=None,  # NOTE: we do not save creation date in the DB
```
leafty, line 166: Maybe we should keep an issue to fix the API spec then?

bases/renku_data_services/background_jobs/core.py
olevski: I know I hope that it will when we fix the logging setup.

olevski: I will try swapping the alembic config level to info and see what happens. Hopefully alembic does not flood the console with lots of random messages.

olevski: Logs look normal with the level set to INFO in alembic. So I will change that.

components/renku_data_services/namespace/orm.py
```diff
@@ -71,15 +71,37 @@ class NamespaceORM(BaseORM):
...
+        name: str | None = None
         if self.group:
             created_by = self.group.created_by
+            name = self.group.name
+        elif self.user:
+            if self.user.first_name and self.user.last_name:
+                name = f"{self.user.first_name} {self.user.last_name}"
+            else:
+                name = self.user.first_name or self.user.last_name
+        # NOTE: Mypy should have inferred the 2 types below beacuse at the top we check that both values are not None
```
leafty, line 94: To avoid this, you can have:

```python
    def dump(self) -> models.Namespace:
        """Create a namespace model from the ORM."""
        if self.group_id and self.group:
            created_by = self.group.created_by
            name = self.group.name
            underlying_resource_id = self.group_id
            return models.Namespace(
                id=self.id,
                slug=self.slug,
                latest_slug=self.slug,
                created_by=created_by,
                kind=models.NamespaceKind.group,
                underlying_resource_id=underlying_resource_id,
                name=name,
            )

        if not self.user or not self.user_id:
            raise errors.ProgrammingError(message="Found a namespace that has no group or user associated with it.")

        created_by = self.user_id
        name = (
            f"{self.user.first_name} {self.user.last_name}"
            if self.user.first_name and self.user.last_name
            else self.user.first_name or self.user.last_name
        )
        underlying_resource_id = self.user_id
        return models.Namespace(
            id=self.id,
            slug=self.slug,
            latest_slug=self.slug,
            created_by=created_by,
            kind=models.NamespaceKind.user,
            underlying_resource_id=underlying_resource_id,
            name=name,
        )
```

bases/renku_data_services/background_jobs/main.py
```diff
@@ -20,6 +20,7 @@ async def short_period_sync() -> None:
     await bootstrap_user_namespaces(config)
     await config.syncer.events_sync(config.kc_api)
     await sync_admins_from_keycloak(config.kc_api, Authz(config.authz_config))
+    await fix_mismatched_project_namespace_ids(config)
```
Panaetius, line 23: question: does this need to be done regularly if it's just a one-time fix?

bases/renku_data_services/background_jobs/core.py
leafty: An easy way to have the logging work in the background job is to use `logger = logging.getLogger("background_jobs")` and use that.

leafty:
```python
logger = logging.getLogger("background_jobs")
logger.setLevel(logging.INFO)
```

bases/renku_data_services/background_jobs/main.py
olevski: It is just a one time fix.

See https://github.com/SwissDataScienceCenter/renku-data-services/issues/278 on how we can address the problem of constantly re-running such tasks.

components/renku_data_services/namespace/orm.py
olevski: Yes this is nicer. I will add it.

components/renku_data_services/namespace/blueprints.py
olevski: https://github.com/SwissDataScienceCenter/renku-data-services/issues/283

bases/renku_data_services/background_jobs/core.py
olevski: With logging level set to info in alembic it works, so I will keep it as is.

leafty (approved):

PR #281, opened 2024-06-21
Save storage credentials api

m-alisafaee:


components/renku_data_services/secrets/db.py
```diff
@@ -53,16 +53,35 @@ async def get_secrets_by_ids(self, requested_by: APIUser, secret_ids: list[str])
...
+            kind=kind,
+        )
+
+    @staticmethod
+    def update_secret_orm(secret: SecretORM, encrypted_value: bytes, encrypted_key: bytes) -> None:
+        """Update an existing secret."""
+        secret.encrypted_value = encrypted_value
+        secret.encrypted_key = encrypted_key
+        secret.modification_date = datetime.now(UTC).replace(microsecond=0)
+
```
Panaetius, line 76: Suggestion: I would put these on SecretORM directly, the first as a class method, the latter just regular method.

components/renku_data_services/storage/db.py
```diff
@@ -142,6 +154,95 @@ async def delete_storage(self, storage_id: str, user: base_models.APIUser) -> No
...
+            raise errors.ValidationError(message="APIUser has no id")
+
+        user_secret_key = await self.user_repo.get_or_create_user_secret_key(requested_by=requested_by)
+
+        return encrypt_user_secret(
+            user_id=requested_by.id,
+            user_secret_key=user_secret_key,
+            secret_service_public_key=self.secret_service_public_key,
+            secret_value=secret_value,
+        )
```
Panaetius, line 168: issue: how a secret is encrypted is business logic to me and not really part of the db layer (the db layer just stores/loads things, and doesn't really do any transformations except minor ones). I know in the user code I opted for having it in blueprints.py, which is equivalent to the presentation layer, but that's also the wrong place. 

I discussed with Tasko a few weeks back that we should be more consistent with where things go and we'd settled on putting business logic in a `core.py` in the respective modules, though we haven't had a chance to write this up and discuss it with the team.

For secrets specifically though, I wonder if it'd make sense to have a single module or class that deals with all of the secret business logic so it can be reused where needed (the above method is almost identical to the one in users/blueprints.py

components/renku_data_services/storage/db.py
```diff
@@ -142,6 +154,95 @@ async def delete_storage(self, storage_id: str, user: base_models.APIUser) -> No
...
+            existing_storage_secrets_orm = result.scalars().all()
+
+            existing_secrets = {s.name: s for s in existing_storage_secrets_orm}
+            stored_secrets = []
+
+            for name, value in secret_names_values.items():
+                encrypted_value, encrypted_key = await self._encrypt_user_secret(requested_by=user, secret_value=value)
+
+                if name in existing_secrets:
+                    storage_secret_orm = existing_secrets[name]
```
Panaetius, line 195: suggestion:
```suggestion
                if storage_secret_orm := existing_secrets.get(name):
```

components/renku_data_services/storage/models.py
```diff
@@ -222,3 +222,28 @@ def _from_ambiguous_url(
...
+    def from_dict(cls, data: dict) -> "CloudStorageSecret":
+        """Create the model from a plain dictionary."""
+        if "user_id" not in data:
+            raise errors.ValidationError(message="'user_id' not set")
+        if "storage_id" not in data:
+            raise errors.ValidationError(message="'storage_id' not set")
+        if "name" not in data:
+            raise errors.ValidationError(message="'name' not set")
+        if "secret_id" not in data:
+            raise errors.ValidationError(message="'secret_id' not set")
```
Panaetius, line 245: question: doesn't pydantic already validate this? also I don't see this method used anywhere in this PR.

Panaetius (changes requested):

components/renku_data_services/storage/models.py
m-alisafaee: Most of these classes had validation and I'd also did the same here. Removed.

olevski (dismissed):

Panaetius (approved):

Panaetius (dismissed):

Panaetius (approved):

Panaetius (approved):

PR #288, opened 2024-06-28
feat: add new notebooks endpoints for new amalthea version

olevski:
@andre-code let me know what you think.

### DIFF
--- a/components/renku_data_services/notebooks/__init__.py
+++ b/components/renku_data_services/notebooks/__init__.py
@@ -1,0 +1,1 @@
+"""Endpoints for Jupyter and AmaltheaSessions."""
--- a/components/renku_data_services/notebooks/api.spec.yaml
+++ b/components/renku_data_services/notebooks/api.spec.yaml
@@ -1,0 +1,408 @@
+openapi: 3.0.2
+info:
+  title: Renku Data Services API
+  description: |
+    Service that allows creating, updating, deleting, and managing Renku user sessions.
+    All errors have the same format as the schema called ErrorResponse.
+  version: v1
+servers:
+  - url: /api/data
+  - url: /ui-server/api/data
+paths:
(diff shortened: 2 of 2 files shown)


andre-code (commented):
Thanks @olevski  for including those changes. I checked what we currently use and found the following:

- Does "running degraded" mean the session is stopping? What are the statuses for starting and stopping?

-  We use `details`and `message` for the start session progress bar and the session status popover,  the rest isn't used in v2.:
  ```json
  {
      "details": [
          {"status": "waiting", "step": "Initialization"},
          {"status": "waiting", "step": "Downloading server image"},
          {"status": "waiting", "step": "Cloning and configuring the repository"},
          {"status": "waiting", "step": "Git credentials services"},
          {"status": "waiting", "step": "Authentication and proxying services"},
          {"status": "waiting", "step": "Auxiliary session services"},
          {"status": "waiting", "step": "Starting session"}
      ],
      "message": "Steps with non-ready statuses: initialization, downloading server image, cloning and configuring the repository, git credentials services, authentication and proxying services, auxiliary session services, starting session.",
      "readyNumContainers": 0,
      "state": "starting",
      "totalNumContainers": 7,
      "warnings": []
  }
  ```


- Can `session_launcher_id` be null for orphan sessions?

- We show an alert icon when `annotations.default_image_used` is set and the session is running. Do you know if this is needed, and why?

- Is `created_at` the same as `started`?

- We use `session.url` to offer an "open in new tab" option, but it seems missing.

- We show `session.resources.request` for resource changes when hibernating. The `resource_class_id` could no exist anymore and we lose that info

- The schema for launching a session isn’t clear on sending more info than cloud storage IDs:
  - Can we send credentials?
  - Can we overwrite `environment_variables`?
  - What are the new values? the ones included in the jupyter free pitch

PR #292, opened 2024-07-01
test: make schemathesis request time checks more consistent

### LINKED ISSUE #231: Flaky schemathesis tests on API endpoints with `authzed` checks
Schemathesis tests can be flaky for API endpoints which query permissions. I suggest listing them as slow endpoints.

Panaetius:
closes #231 

this collects execution times using a session scoped fixture and then checks that 95% of them took no longer than 100ms to complete (in absolute terms, around 15 of our 269 tests can be slower than this.)


olevski (approved):
Thanks! It will be great to not rerun this all the time.

PR #296, opened 2024-07-07
feat: save storage secrets

m-alisafaee:


components/renku_data_services/storage/db.py
```diff
@@ -71,25 +73,38 @@ async def get_storage(
...
             )
 
-            return [p.dump() for p in orms if p.project_id in accessible_projects]
+            storages = [s.dump() for s in storage_orms if s.project_id in accessible_projects]
+
+            if include_secrets:
+                for storage in storages:
+                    storage_id = cast(ULID, storage.storage_id)
+                    saved_secrets = await self.get_storage_secrets(storage_id, user)
+                    storage.secrets = saved_secrets
```
Panaetius, line 87: suggestion: I think this can be more efficiently handled above around line 69 by conditionally adding something like `.options(selectinload(schemas.StorageORM.secrets))` if we add it on CloudStorageORM.

m-alisafaee: The issue here is that, `CloudStorageORM` doesn't include `user_id`, so, we cannot use it to filter secrets for a specific user. A `secrets` relationship in `CloudStorageORM` would return all secrets from `CloudStorageSecretsORM` that are saved for a storage. I can add it and then filter it based on the calling user, but we always need to remember that this is the case. Wdyt?

Panaetius (approved):

PR #299, opened 2024-07-10
feat: add custom openBIS type

olloz26:


Panaetius (changes requested):
Failing due to not passing style checks. If you install https://pre-commit.com/ it'd check this before every commit. Alternatively, run `make style_checks` to run the checks locally.
In this case, a simple `ruff check --fix` should fix it.

components/renku_data_services/storage/rclone.py
```diff
@@ -363,7 +416,7 @@ class RCloneProviderSchema(BaseModel):
     @property
     def required_options(self) -> list[RCloneOption]:
         """Returns all required options for this provider."""
-        return [o for o in self.options if o.required]
+        return [o for o in self.options if o.required and not o.sensitive]
```
Panaetius, line 419: issue: could you write a comment on why sensitive required options should not be returned here?

components/renku_data_services/utils/core.py
```diff
@@ -90,3 +91,71 @@ async def transaction_wrapper(self: _WithSessionMaker, *args: _P.args, **kwargs:
...
+async def get_openbis_session_token(
+    host: str,
+    username: str,
+    password: str,
+    timeout: int = 12,
+) -> str:
+    """Requests an openBIS session token with the user's login credentials."""
+    login = {"method": "login", "params": [username, password], "id": "2", "jsonrpc": "2.0"}
+    async with httpx.AsyncClient(verify=get_ssl_context()) as client:
+        response = await client.post(_get_url(host), json=login, timeout=timeout)
```
Panaetius, line 109: issue: I think we have to check the `response.status_code` and raise an error if the request fails (see e.g. in `gitlab.py` in `_query_gitlab_graphql`)

components/renku_data_services/utils/core.py
```diff
@@ -90,3 +91,71 @@ async def transaction_wrapper(self: _WithSessionMaker, *args: _P.args, **kwargs:
...
+    minimum_validity_in_days: int = 2,
+    timeout: int = 12,
+) -> tuple[str, datetime]:
+    """Requests an openBIS PAT with an openBIS session ID."""
+    url = _get_url(host)
+
+    async with httpx.AsyncClient(verify=get_ssl_context()) as client:
+        get_server_information = {"method": "getServerInformation", "params": [session_id], "id": "2", "jsonrpc": "2.0"}
+        response = await client.post(url, json=get_server_information, timeout=timeout)
+        if response.status_code == 200:
```
Panaetius, line 127: suggestion: I'd flip the if-statement to `!= 200` with error raising to reduce nesting of this code block and keep the raised error closer to the thing that causes it, same for the if statements below, with more tailored error messages to help tell apart what went wrong if something goes wrong. it's more code to write but if something ever goes wrong it's much easier to debug.

Panaetius (changes requested):

components/renku_data_services/storage/db.py
```diff
@@ -185,6 +186,7 @@ async def upsert_storage_secrets(
             existing_secrets = {s.name: s for s in existing_storage_secrets_orm}
             stored_secrets = []
 
+            secret_names_values = {s.name: s.value for s in secrets}
             for name, value in secret_names_values.items():
```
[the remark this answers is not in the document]
olloz26: I know. This is not my code. My goal was to keep existing code as unchanged as possible.

components/renku_data_services/storage/rclone.py
```diff
@@ -126,6 +126,59 @@ def __patch_schema_add_switch_provider(spec: list[dict[str, Any]]) -> None:
...
+    def __patch_schema_add_openbis_type(spec: list[dict[str, Any]]) -> None:
+        """Adds a fake type to help with setting up openBIS storage."""
+        spec.append(
+            {
+                "Name": "openbis",
+                "Description": "openBIS",
+                "Prefix": "openbis",
+                "Options": [
+                    {
+                        "Name": "host",
```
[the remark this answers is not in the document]
olloz26: Works as you described.

components/renku_data_services/utils/core.py
olloz26: True. I will add a unit test for it.

components/renku_data_services/utils/core.py
olloz26: I prefer "whitelist" ifs because they make the code more robust. And robust code does not need to be debugged 😉  Anyway, I can apply your suggestions if you want.

Panaetius: it's up to you (that's why marked as suggestion). i prefer less nesting but it's a matter of choice :)

components/renku_data_services/storage/rclone.py
olloz26: I improved the code so that this is no longer necessary.

PR #310, opened 2024-07-16
chore: add kind to devcontainer

olevski:
Also installs both amalthea operators in the kind cluster.


.github/workflows/save_cache.yml
leafty, line 1: Why not use caching from [devcontainers/ci](https://github.com/devcontainers/ci)? This would cache everything we use in the devcontainer steps and is much easier to setup (add a few args to existing tasks).

olevski: The python environment is not part of the image. So with that the python package installation will not be cached because it is not part of the image since it happens in a postCreate command after the container starts running.

And I wanted to cache the python packages.

I can add the caching of the image with devcontainer ci too.

leafty: I think that is not exactly working the same as in a local machine. If you look at the action logs, the time spent in the `postCreate` phase is approx. ~15 seconds while the image build take 3-4 minutes. Caching the image will yield a much nicer benefit compared to doing poetry cache.

Also, it seems that poetry already caches packages in the Dockerfile.

olevski: Ok now both the poetry cache and the devcontainer image cache should work.

olevski: But they will come into effect once we merge this to main. I wanted to only update the caches on merges to main.

.github/workflows/save_cache.yml
```diff
@@ -15,12 +15,20 @@ jobs:
...
       - name: Install python deps
         uses: devcontainers/ci@v0.3
         with:
           runCmd: poetry install --with dev
-          push: never
+          push: always
           skipContainerUserIdUpdate: false
+          imageName: renku/renku-data-service
+          imageTag: devcontainer-main
+          cacheFrom: renku/renku-data-service:devcontainer-main
```
leafty, line 31: I think using GHCR for this would be much faster (near storage in the same infrastructure vs remote registry).

.github/workflows/save_cache.yml
```diff
@@ -18,17 +19,18 @@ jobs:
...
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
       - name: Install python deps
         uses: devcontainers/ci@v0.3
         with:
           runCmd: poetry install --with dev
           push: always
           skipContainerUserIdUpdate: false
-          imageName: renku/renku-data-service
+          imageName: ghcr.io/${{ github.repository }}
```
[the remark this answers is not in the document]
olevski: oh wow ok thanks... I was trying to understand wth is going on

.github/workflows/save_cache.yml
leafty: Great!

.devcontainer/.poetry_cache/.gitignore
```diff
@@ -0,0 +1,2 @@
+*
```
leafty, line 1: non-blocking: would it also work to add `.devcontainer/.poetry_cache` to the repo `.gitignore` and force-commit a `.keep` file? This feels more conventional.

leafty (commented):
The `.github` changes look good 👍.

olevski:
@Panaetius as discussed will wait for your review here. I am already using this setup to run tests in my PR for the new operator. Currently the kind stuff is not being used.

.github/workflows/save_cache.yml
```diff
@@ -0,0 +1,41 @@
...
+          runCmd: poetry install --with dev
+          push: always
+          skipContainerUserIdUpdate: false
+          imageName: ${{ env.DEVCONTAINER_IMAGE_CACHE }}
+          cacheFrom: ${{ env.DEVCONTAINER_IMAGE_CACHE }}
+      - uses: actions/cache/save@v3
+        name: Create cache
+        with:
+          path: ${{ env.CACHE_PATH }}
+          key: ${{ env.CACHE_KEY }}
```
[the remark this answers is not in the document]
olevski: This is what I wanted. As in I dont want perfect caching but as long as we can reuse the cache from main but not change it that is fine. I am a bit scared to keep updating the cache all the time.

Also my understanding of the github caches is that they cannot be used across branches. So you can only use the cache from the current branch or from a parent branch. With this in mind having this more ideal caching seemed less important because in a fresh PR you can only use the cache from main. And I worried that if we add the hash of poetry.lock then we have even less chances of hits. But we can try this.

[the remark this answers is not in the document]
Panaetius: the bullet point right below leafty's in the link she posted mentions that feature branches can share the cache of the main branch, just not from one feature branch in another.

Panaetius (dismissed):
looks good other than the comments that are already here. Let me know once those are addressed and I can reapprove

.github/workflows/save_cache.yml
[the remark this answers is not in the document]
leafty: No, the point is that as is, you will create a single cache archive FOREVER. It will not update, which means that updates to the dependencies contained in poetry will not be cached after being merged to `main`.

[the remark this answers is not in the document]
leafty: > I am a bit scared to keep updating the cache all the time.

This will not happen if the cache key reflects the `poetry.lock` contents.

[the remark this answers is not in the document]
olevski: Ok I removed all the poetry/python caching. We can add that later.

.github/workflows/test_publish.yml
```diff
@@ -9,6 +9,9 @@ on:
   pull_request:
     branches:
 
+env:
+  DEVCONTAINER_IMAGE_CACHE: ghcr.io/swissdatasciencecenter/renku-data-services/devcontainer
```
leafty, line 13: non-blocking: the cache image is never created.

leafty (approved):

Panaetius (approved):

PR #341, opened 2024-08-06
refactor: add validation to connected services

Panaetius:

PR #340, opened 2024-08-06
test: fix schemathesis query parameters test

Panaetius:
Validates query parameters, getting schemathesis negative tests to work.

PR #326, opened 2024-07-26
refactor: add validation to crc, user and secrets endpoints

Panaetius:
adds validation of ids and responses to crc, user and secret endpoints. 
handles response validation errors correctly
part of #304

PR #325, opened 2024-07-26
refactor: unify user and user_preferences modules

Panaetius:
moves `user_preferences` over to `users` namespace.
Also moves table to `users` schema, migrating old preferences and deleting the `user_preferences` db schema.

part of addressing #304

PR #342, opened 2024-08-06
refactor: add validation to namespaces blueprint

Panaetius:
also adds an extra method to allow proper validation errors in pagination returns

### DIFF
--- a/components/renku_data_services/base_models/validation.py
+++ b/components/renku_data_services/base_models/validation.py
@@ -11,6 +11,23 @@ from sanic.response import JSONResponse
 from renku_data_services import errors
 
 
+def validate_and_dump(
+    model: type[BaseModel],
+    data: Any,
+    exclude_none: bool = True,
+) -> Any:
+    """Validate and dump with a pydantic model, ensuring proper validation errors."""
+    try:
+        body = model.model_validate(data).model_dump(exclude_none=exclude_none, mode="json")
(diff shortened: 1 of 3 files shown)


Panaetius:

- #340
    - #325
        - #326
            - #341
                - #342 ◀

components/renku_data_services/base_models/validation.py
```diff
@@ -11,6 +11,23 @@
 from renku_data_services import errors
 
 
+def validate_and_dump(
+    model: type[BaseModel],
+    data: Any,
+    exclude_none: bool = True,
+) -> Any:
```
olevski, line 18: suggestion(non-blocking): Should we make the return type `dict | list`?

olevski (approved):

components/renku_data_services/base_models/validation.py
Panaetius: I don't know if that's actually better.
- pydantic has as type annotation for model_dump `dict[str, Any]` which is wrong (as it'll produce a list when supplied a list)
- I could of course change the type and just `cast(...)` in the function here. But then I also need to `cast` to `list` at the callsite (otherwise mypy complains that `_get_all` is supposed to return a list but now we get `dict|list`)
- so it's two casts or just setting it to `Any` and there's no validation. I don't think there's a difference in safety in this case, but the two casts may signify intent better?

PR #343, opened 2024-08-06
fix: query args validation of /user/secrets GET

### LINKED ISSUE #312: Potential high failure rate of `test_api_schemathesis[GET /api/data/user/secrets]`
See for example: https://github.com/SwissDataScienceCenter/renku-data-services/actions/runs/9970890833/job/27552572648?pr=285.

olevski:
fixes #312

### DIFF
--- a/components/renku_data_services/base_api/error_handler.py
+++ b/components/renku_data_services/base_api/error_handler.py
@@ -1,5 +1,8 @@
 """The error handler for the application."""
 
+import os
+import sys
+import traceback
 from collections.abc import Mapping, Set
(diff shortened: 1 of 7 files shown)


Panaetius (approved):

olevski:
@Panaetius I thought I could handle the query param validation in a simpler way simply by passing the args to the pydantic schema but I could not. Because `request.args` is a dictionary of query name to list of values - even if there is one value. And sanic already deals with this logic correctly so it was easier to just wrap the sanic validator and catch the `KeyError`.

Panaetius (approved):

PR #345, opened 2024-08-06
refactor: add validation to project blueprints

Panaetius:


Panaetius:

- #340
    - #325
        - #326
            - #341
                - #342
                    - #345 ◀

components/renku_data_services/session/models.py
```diff
@@ -17,10 +26,9 @@ class Member(BaseModel):
 
 
 @dataclass(frozen=True, eq=True, kw_only=True)
-class Environment(BaseModel):
-    """Session environment model."""
+class UnsavedEnvironment(BaseModel):
+    """Session environment model that isn't in the db yet."""
```
olevski, line 30: suggestion: Since we are changing these, there is no reason whatsoever to use pydantic here. These can be just dataclasses. I already did this change in the PR that expands the environments and sessions.

components/renku_data_services/session/models.py
```diff
@@ -30,11 +38,17 @@ class Environment(BaseModel):
...
+class Environment(UnsavedEnvironment):  # type: ignore[misc]
+    """Session environment model."""
+
+    id: str
+
 
-    id: str | None
-    project_id: str
+@dataclass(frozen=True, eq=True, kw_only=True)
+class UnsavedSessionLauncher(BaseModel):
```
olevski, line 48: suggestion: Same as above there is no reason to use pydantic here. The model validator below can just be a dataclass `__post_init__` method.

olevski: I already did this in this PR: https://github.com/SwissDataScienceCenter/renku-data-services/pull/338/files#diff-f3319353302e4b6f60f44bc85e62fb3130e542a9b71c91f9ed6803060dc57c54R34

olevski (changes requested):

PR #347, opened 2024-08-09
refactor: add validation to storage, repo and session blueprints

Panaetius:
I had to change the ULID regex for the schemathesis tests. a 26 char ulid can theoretically store 130 bits, but ulid are limited to 128 bit. so they can't start with `8` or higher (that would be a too high timestamp). As per the [spec](https://github.com/ulid/spec?tab=readme-ov-file#overflow-errors-when-parsing-base32-strings)

EDIT: A breaking a change that requires changes on the UI. -> https://github.com/SwissDataScienceCenter/renku-ui/pull/3299

/deploy renku=feat-jupyter-free-sessions amalthea-sessions=main renku-notebooks=feat-add-amalthea-session-cache renku-gateway=master renku-ui=andrea/use-data-service-sessions-2.0 extra-values=amalthea-sessions.deployCrd=false #notest

PR #346, opened 2024-08-08
refactor: add custom ULID type for sqlalchemy

Panaetius:
this allows us to have ULID on ORMs. In the DB it will still be a string
/deploy

PR #348, opened 2024-08-09
refactor: add sqlalchemy ulid type to all relevant models

Panaetius:


Panaetius:

- #326
    - #341
        - #342
            - #345
                - #346
                    - #347
                        - #348 ◀

olevski (approved):

components/renku_data_services/authz/authz.py
```diff
@@ -580,7 +580,7 @@ def _add_project(self, project: Project) -> _AuthzChange:
             object=(
                 _AuthzConverter.user_namespace(project.namespace.id)
                 if project.namespace.kind == NamespaceKind.user
-                else _AuthzConverter.group(project.namespace.underlying_resource_id)
+                else _AuthzConverter.group(cast(ULID, project.namespace.underlying_resource_id))
```
olevski, line 583: Why do we need this cast?

Panaetius: because `undelying_resource_id` is an ULID for groups but a str keycloak id for user namespaces. so we know it is ULID here due to the if/else on `kind`, but mypy does not know this.
somewhat related, we should discuss at some point on if we want to use our own user ids or if we want to keep using keycloak ones directly.

PR #362, opened 2024-08-21
refactor: merge UserInfo and UserWithNamespace

Panaetius:
Combines UserInfo and UserWithNamespace classes into a single user object that always has to have a namespace (as there are no users without a namespace.

### DIFF
--- a/components/renku_data_services/users/db.py
+++ b/components/renku_data_services/users/db.py
@@ -195,7 +195,7 @@ class UsersSync:
         kwargs.pop("keycloak_id", None)
         kwargs.pop("id", None)
         slug = base_models.Slug.from_user(
-            kwargs["email"], kwargs["first_name"], kwargs["last_name"], user_id
+            kwargs.get("email"), kwargs.get("first_name"), kwargs.get("last_name"), user_id
         ).value.lower()
         namespace = await self.group_repo._create_user_namespace_slug(
             session, user_slug=slug, retry_enumerate=5, retry_random=True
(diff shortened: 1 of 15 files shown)


Panaetius:

- #362 ◀

components/renku_data_services/users/db.py
```diff
@@ -176,36 +177,45 @@ async def _get_user(self, id: str) -> UserInfo | None:
...
 
-    async def _insert_user(self, session: AsyncSession, user_id: str, **kwargs: Any) -> UserWithNamespaceUpdate:
+    async def _insert_user(self, session: AsyncSession, user_id: str, **kwargs: Any) -> UserInfoUpdate:
         """Insert a user."""
         kwargs.pop("keycloak_id", None)
         kwargs.pop("id", None)
-        new_user = UserORM(keycloak_id=user_id, **kwargs)
+        slug = base_models.Slug.from_user(
+            kwargs.get("email"), kwargs.get("first_name"), kwargs.get("last_name"), user_id
+        ).value.lower()
```
olevski, line 199: suggestion: This is not necessary since the `__init__` method of `Slug` already calls `lower` on the value when it sets it.

components/renku_data_services/users/db.py
```diff
@@ -176,36 +177,45 @@ async def _get_user(self, id: str) -> UserInfo | None:
...
         self, user_id: str, payload: dict[str, Any], *, session: AsyncSession | None = None
-    ) -> UserWithNamespaceUpdate:
+    ) -> UserInfoUpdate:
         """Update a user or insert it if it does not exist."""
         if not session:
             raise errors.ProgrammingError(message="A database session is required")
-        res = await session.execute(select(UserORM).where(UserORM.keycloak_id == user_id))
+        res = await session.execute(
+            select(UserORM).where(UserORM.keycloak_id == user_id).options(selectinload(UserORM.namespace))
+        )
```
olevski, line 186: suggestion (non-blocking): This is really nit-picky and really not a problem. But it may be nicer to add this on the schema so that you dont have to repeat `options(selectinload(UserORM.namespace))`. But it is not that big of a deal. So ignore this if you wish.

olevski (changes requested):
Thanks for doing this Ralf. Just one minor change.

olevski (approved):

PR #366, opened 2024-08-23
fix: do not validate user emails

olevski:
This causes problems because invalid emails can show up in Keycloak either from other other identity providers.

/deploy


components/renku_data_services/namespace/api.spec.yaml
```diff
@@ -519,7 +519,10 @@ components:
       maxLength: 256
     UserEmail:
       type: string
-      format: email
+      # We can get invalid emails when people log in via Github or other serivces
```
Panaetius, line 522:
```suggestion
      # We can get invalid emails when people log in via Github or other services
```
also in the other docstrings.
We should create an issue with the Sentry contents as sentry errors are not durable and this link will likely be dead in a couple of weeks

Panaetius (approved):

PR #371, opened 2024-08-26
refactor: add typestate pattern for namespace, secrets and storage

Panaetius:


Panaetius:

- #371 ◀

components/renku_data_services/authz/models.py
```diff
@@ -53,11 +53,21 @@ class Scope(Enum):
...
+
+    def with_group(self, group_id: ULID) -> "Member":
+        """Turn to member with group."""
+        return Member(role=self.role, user_id=self.user_id, resource_id=group_id)
+
+
+@dataclass
+class Member(UnsavedMember):
+    """Member stored in the database."""
+
```
olevski, line 70: suggestion: On the line below this we have `resource_id: str | ULID` but it can be converted to just `ULID` by changing `components/renku_data_services/project/blueprints.py:264` to use `ULID.from_str` for the project ID in the body of the request.

I know this is not part of the PR but it would be nice to be more restrictive with the types if we can and in this case we can. I ran mypy when I was trying this and that is the only thing it complained about.

olevski (changes requested):
Thanks Ralf just one minor change.

components/renku_data_services/authz/models.py
Panaetius: I've done that in https://github.com/SwissDataScienceCenter/renku-data-services/pull/347 where I added ULID for all project endpoints

olevski (approved):

PR #360, opened 2024-08-20
feat: add new apispec for the new amalthea sessions

olevski:
This will be used only in Renku v2 with the new Amalthea operator.

Issue #303, opened 2024-07-15
Followup work from moving renku notebooks


olevski:
Leftover from renku notebooks code move:

- components/renku_data_services/notebooks/api/classes/data_service.py: I guess you'll do that in a followup PR but all the http calls in this file are superfluous and can be replaced with direct calls
- components/renku_data_services/notebooks/api/notebooks.py L677: the error message could be more descriptive, if I saw this in a log, I wouldn't know what the cause was.
- components/renku_data_services/notebooks/api/notebooks.py L683 and 685: isn't this what empty is for?
- components/renku_data_services/notebooks/config/init.py: TODO for later, move out of init.py
- components/renku_data_services/notebooks/util/authn.py: We need to refactor this code. we create a RegisteredUser no matter what, check if authenticated is True (which it is if all required headers are present, but we don't actually check the content of the headers in the RegisteredUser init) and if not create an AnonymousUser. This seems pretty brittle and is a weird way for this code to be written. and if I read it correctly, I could put whatever I want in those headers and it would consider it a registered user?

Related to #301

PR #375, opened 2024-08-27
feat: run notebooks in data service

olevski:
This includes major edits to the notebooks code to work with the data service.

/deploy renku=feat-jupyter-free-sessions amalthea-sessions=feat-add-cloud-storage renku-notebooks=feat-jupyter-free-sessions renku-gateway=feat-add-renku-refresh-token-for-data-service renku-ui=andrea/custom-launcher-jupyter-free extra-values=amalthea-sessions.deployCrd=false #notest


olevski:

- #360
    - #338
        - #375 ◀

components/renku_data_services/notebooks/api/amalthea_patches/init_containers.py
```diff
@@ -79,21 +80,170 @@ def git_clone(server: "UserServer") -> list[dict[str, Any]]:
...
+            if full_name == "":
+                full_name = server.user.last_name or ""
+            else:
+                full_name += " " + (server.user.last_name or "")
+            env.append(
+                {
+                    "name": f"{prefix}USER__FULL_NAME",
+                    "value": server.user.full_name,
+                },
+            )
```
Panaetius, line 243: nitpic: same here, if this was a method on user, it'd simplify things a bit

components/renku_data_services/base_api/auth.py
```diff
@@ -153,3 +244,30 @@ async def decorated_function(*args: _P.args, **kwargs: _P.kwargs) -> _T:
         return response
 
     return decorated_function
+
+
+def internal_gitlab_authenticate(
```
Panaetius, line 249: suggestion: I think we could rewrite the decorator to accept chained authenticators, and the authenticators can wrap the user objects with dedicated objects (Like GitUser(APIUser()) or similar, if their `authenticate` method accepted an optional `user` object from previous calls) to be more flexible. That would allow adding additional authenticators more easily without an explosion of custom decorators, but I'm a bit torn on if it actually makes the code simpler/easier to understand.

components/renku_data_services/notebooks/api/classes/data_service.py
Panaetius, line 1: question: is there an issue for getting rid of this file in favor of direct calls?

components/renku_data_services/notebooks/api/classes/k8s_client.py
```diff
@@ -2,240 +2,215 @@
...
+from renku_data_services.notebooks.util.kubernetes_ import find_env_var
+from renku_data_services.notebooks.util.retries import (
+    retry_with_exponential_backoff_async,
+)
+
+sanitize_for_serialization = ApiClient().sanitize_for_serialization
+
+
+# NOTE The type ignore below is because the kr8s library has no type stubs, they claim pyright better handles type hints
+class AmaltheaSessionV1Alpha1Kr8s(APIObject):  # type: ignore
```
Panaetius, line 36: question: I'm a bit confused why this uses `Kr8s` instead of the more common `k8s`? even the doc string uses k8s. that applies to this whole file

components/renku_data_services/notebooks/api/classes/k8s_client.py
```diff
@@ -259,26 +234,17 @@ def patch_image_pull_secret(self, server_name: str, gitlab_token: GitlabToken) -
...
-                return
-            raise
+            sts = cast(StatefulSet, await StatefulSet.get(name=name, namespace=self.namespace))
+        except NotFoundError:
+            return None
 
-        containers: list[V1Container] = sts.spec.template.spec.containers
-        init_containers: list[V1Container] = sts.spec.template.spec.init_containers
+        containers: list[V1Container] = [V1Container(**i) for i in sts.spec.template.spec.containers]
+        init_containers: list[V1Container] = [V1Container(**i) for i in sts.spec.template.spec.init_containers]
```
Panaetius, line 247: nitpick: I wouldn't use `i` as a loop variable for something that isn't an index

Panaetius (changes requested):
we should probably make all the api/amalthea_patches/* return typed objects instead of dicts.

components/renku_data_services/notebooks/blueprints.py
```diff
@@ -669,18 +742,257 @@ async def _server_logs(
...
+            work_dir = Path("/home/jovyan/work")
+            user_secrets: K8sUserSecrets | None = None
+            # if body.user_secrets:
+            #     user_secrets = K8sUserSecrets(
+            #         name=server_name,
+            #         user_secret_ids=body.user_secrets.user_secret_ids,
+            #         mount_path=body.user_secrets.mount_path,
+            #     )
+            cloud_storage: list[RCloneStorage] = []
+            # repositories = [Repository(i.url, branch=i.branch, commit_sha=i.commit_sha) for i in body.repositories]
```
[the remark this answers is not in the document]
sgaist: Will remove that with the commit(s) that should make the start endpoint working

components/renku_data_services/notebooks/api/classes/data_service.py
sgaist: Do you mean a GitHub issue associated to this task or if there's a problem with going with direct calls ?
For the former, #303 lists it.
For the latter, AFAIK, the plan is to get notebooks working and then refactor it to remove the indirection.

components/renku_data_services/notebooks/api/classes/k8s_client.py
sgaist: Agreed, I already changed a similar loop in another part.
Should this be done in this PR or left for the following cleanup PR ?

components/renku_data_services/base_api/auth.py
sgaist: Are you thinking of something like:
```
@authenticate([APIUser, GitUser])
```
Passing an ordered list of authenticator my be more readable

components/renku_data_services/notebooks/api/classes/data_service.py
Panaetius: yes a github issue

components/renku_data_services/notebooks/api/classes/k8s_client.py
Panaetius: i think small things like that should be done in this PR otherwise they'll get lost

components/renku_data_services/notebooks/blueprints.py
```diff
@@ -212,17 +265,18 @@ async def _launch_notebook_old(
...
                 repositories=None,
+                internal_gitlab_user=internal_gitlab_user,
             )
             return json(NotebookResponse().dump(server), status_code)
 
         return "/notebooks/old/servers", ["POST"], _launch_notebook_old
 
     @staticmethod
-    def launch_notebook_helper(
+    async def launch_notebook_helper(
```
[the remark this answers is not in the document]
Panaetius: yes doing this in a follow up is fine (as long as we collect it in an issue so it doesn't get lost).

components/renku_data_services/notebooks/blueprints.py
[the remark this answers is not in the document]
Panaetius: `server_name` is a parameter in the apispec for a couple of endpoints, e.g. line 167 `return "/notebooks/servers/<server_name>", ["GET"], _user_server` and in the apispec it's just `string`, and my suggestion was to do some basic validation on this, since we know what it's supposed to look like. That makes it clearer to callers what a server name is supposed to be and we could reject bad requests without checking for a server.
It's not strictly necessary, the code works as is, hence marked `suggestion`, but I think we at least know how short or long a server name should be and could add max\min length restrictions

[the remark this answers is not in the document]
Panaetius: well in general `blueprints.py` should only take care of presentation logic (converting from/to json, things like pagination etc.) and all logic on how things work should be seperate. right now this file has a LOT of logic on constructing a server, which should be elsewhere (`core.py` by convention) also also not inside of http handler methods. that's already the case with some of the methods, like `launch_notebook_helper`, but e.g. `patch_server()` contains quite a bit of business logic. So e.g. if I ever wanted to patch a server from some other code, I couldn't reuse this (Or I'd have to fake a request object).
So ideally we'd just have methods for `do_the_thing` in core.py and in blueprints.py it's just wiring to hook those methods up to the REST api.

components/renku_data_services/notebooks/config/dynamic.py
```diff
@@ -389,18 +406,46 @@ def from_env(cls) -> Self:
...
 
-    renku_namespace: str
+    renku_namespace: str = "default"
     sessions_namespace: Optional[str] = None
 
     @classmethod
     def from_env(cls) -> Self:
         return cls(
-            renku_namespace=os.environ["KUBERNETES_NAMESPACE"],
+            renku_namespace=os.environ.get("KUBERNETES_NAMESPACE", "default"),
```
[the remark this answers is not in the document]
Panaetius: yes I know how environ works.
I don't know what K8s does with a namespace of `default`. Because the `default` namespace in any of our clusters is empty and not used, so `default` is a bad default for us. 
I was just wondering if `default` is some magic value that automatically picks e.g. `renku` in our cluster. I don't think that's the case, but then I don't really know why we would use `default` as the default, as that won't work in any of our deployments and erroring here might be a better choice.

components/renku_data_services/base_api/auth.py
Panaetius: well I guess it'd be `@authenticate([self.authenticator, self.gitlab_authenticator])`, the `GitUser(ApiUser())` was more of an example of what the return value of that might be. Either using a decorator pattern (classes with the same base api wrapping each other) or just composition in general? Could also just be adding to some context object on APIUser. as an end-result, it'd just be nice to have one object that can carry all the user authentication related information (especially since typing decorators is such a pain and having so do stuff like "the api user is always in position 2 and the gitlab user is always in position 3" is pretty annoying and conflict prone)

components/renku_data_services/notebooks/api/amalthea_patches/init_containers.py
```diff
@@ -79,21 +80,170 @@ def git_clone(server: "UserServer") -> list[dict[str, Any]]:
...
```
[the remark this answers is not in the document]
Panaetius: no more like `server.user.get_full_name()` and use that instead of duplicating the logic in the 2-3 places this is used.
But maybe an get_environ method would be nice as well in addition to that, if the environ is the same for the different containers.

components/renku_data_services/notebooks/api/classes/k8s_client.py
sgaist: Done for all the loops within the notebook volume.

components/renku_data_services/notebooks/blueprints.py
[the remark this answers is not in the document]
sgaist: I updated #303 so it has now a task list and this part is tracked as #396

components/renku_data_services/notebooks/blueprints.py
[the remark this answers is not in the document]
sgaist: My bad, I was looking at the wrong thing.

I agree, the earlier the validation the better. Task added to #303 and tracked as #398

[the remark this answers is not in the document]
sgaist: This will also help makes thing more testable.

Tracked as #399

components/renku_data_services/notebooks/config/dynamic.py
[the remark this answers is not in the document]
sgaist: Somehow I knew I was answering the wrong question, sorry.

I think there are two aspects here:
- The namespace where Renku itself is deployed
- The namespace where the sessions are created

Are they always the same ?

components/renku_data_services/notebooks/config/dynamic.py
```diff
@@ -203,7 +203,7 @@ class _CustomCaCertsConfig:
     @classmethod
     def from_env(cls) -> Self:
         return cls(
-            image=os.environ.get("NB_SESSIONS__CA_CERTS__IMAGE", "renku-jupyterserver"),
+            image=os.environ.get("NB_SESSIONS__CA_CERTS__IMAGE", "renku/certificates:0.0.2"),
```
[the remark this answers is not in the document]
sgaist: Tracked at #400

components/renku_data_services/notebooks/config/dynamic.py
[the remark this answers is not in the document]
Panaetius: they can be different but in practice they are not, we don't use that setting anywhere to my knowledge.

olevski:
> we should probably make all the api/amalthea_patches/* return typed objects instead of dicts.

@Panaetius I dont see the point in this. The stuff that the new amalthea uses is typed. And the rest is not typed and I dont think it is worth adding this. Since we will at some point retire the code. What we can do is call `dict[str, Any]` a specific custom type for each patch and use that. Either way I will open an issue for this. https://github.com/SwissDataScienceCenter/renku-data-services/issues/414

components/renku_data_services/base_api/auth.py
```diff
@@ -44,9 +45,99 @@ async def decorated_function(request: Request, *args: _P.args, **kwargs: _P.kwar
...
+
+            response = await f(request, user, *args, **kwargs)
+            return response
+
+        return decorated_function
+
+    return decorator
+
+
+def only_authenticated_new(
```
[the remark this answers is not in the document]
olevski: This is not used at all. So I will remove it.

components/renku_data_services/notebooks/api/amalthea_patches/init_containers.py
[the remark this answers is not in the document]
olevski: I added a `get_full_name` method on the APIUser that uses the code you suggested.

components/renku_data_services/notebooks/api/classes/k8s_client.py
olevski: I reworded the docstring. Normal python k8s has zero async support - the main reason for using kr8s is that they support async.

olevski: Or are you talking about renaming the title of the class to be `AmaltheaSessionV1Alpha1K8s` instead of `AmaltheaSessionV1Alpha1Kr8s` i.e. just removing the `r` from the name?

components/renku_data_services/notebooks/blueprints.py
```diff
@@ -602,35 +671,39 @@ async def _patch_server(
...
                 renku_tokens = RenkuTokens(access_token=user.access_token, refresh_token=user.refresh_token)
-                gitlab_token = GitlabToken(access_token=user.git_token, expires_at=user.git_token_expires_at)
-                self.nb_config.k8s_client.patch_tokens(server_name, renku_tokens, gitlab_token)
-                new_server = self.nb_config.k8s_client.patch_server(
-                    server_name=server_name, safe_username=user.safe_username, patch=patch
+                gitlab_token = GitlabToken(
+                    access_token=internal_gitlab_user.access_token,
+                    expires_at=floor(user.access_token_expires_at.timestamp())
+                    if user.access_token_expires_at is not None
+                    else -1,
```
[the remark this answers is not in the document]
olevski: You are talking about formatting it like this, right?
```
                    expires_at=(
                        floor(user.access_token_expires_at.timestamp())
                        if user.access_token_expires_at is not None
                        else -1
                    ),
```

Panaetius:
> > we should probably make all the api/amalthea_patches/* return typed objects instead of dicts.
> 
> @Panaetius I dont see the point in this. The stuff that the new amalthea uses is typed. And the rest is not typed and I dont think it is worth adding this. Since we will at some point retire the code. What we can do is call `dict[str, Any]` a specific custom type for each patch and use that. Either way I will open an issue for this. #414

fair enough. Having the nicely typed returns in some places just made the `dict[str,Any]` stick out like a sore thumb more :slightly_smiling_face: Looking forward to getting rid of the old code!

components/renku_data_services/notebooks/blueprints.py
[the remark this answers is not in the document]
olevski: I added server name validation in the apispec.

components/renku_data_services/notebooks/config/dynamic.py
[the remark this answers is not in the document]
olevski: I dont think we tag `latest` for this at all. Or if we do I am not sure if we are careful enough to not tag the non-release merges to `master` with `latest`. Or even to tag a random commit from a PR with `latest`. So as Samuel mentioned we should address this in a followup.

components/renku_data_services/notebooks/config/dynamic.py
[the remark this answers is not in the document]
olevski: So this is leftover from a feature that we partly or fully implemented on notebooks. Where we wanted Renku services to run in one namespace and sessions to run in a different namespaces. But we never tested it properly and we never deployed. So we can fully remove `sessions_namespace`. I will do that.

components/renku_data_services/notebooks/api/amalthea_patches/init_containers.py
olevski: done

components/renku_data_services/base_api/blueprint.py
```diff
@@ -21,7 +21,7 @@ class CustomBlueprint:
     """
 
     name: str
-    url_prefix: Optional[str] = None
+    url_prefix: str | None = None
```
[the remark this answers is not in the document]
olevski: Done. This is now required and it has to be a string.

components/renku_data_services/base_api/auth.py
```diff
@@ -44,9 +45,99 @@ async def decorated_function(request: Request, *args: _P.args, **kwargs: _P.kwar
...
+        full_name=user.full_name,
+        first_name=user.first_name,
+        last_name=user.last_name,
+        email=user.email,
+        is_admin_init=user.is_admin,
+        refresh_token=user.refresh_token,
+    )
+
+
+def authenticated_or_anonymous(
```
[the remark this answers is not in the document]
olevski: Done.

components/renku_data_services/base_api/auth.py
olevski: > well I guess it'd be @authenticate([self.authenticator, self.gitlab_authenticator])

The problem with this getting mypy and the typing right on such a decorator. And this is possible but it will probably mean that you have to inject the APIUser instances in the decorated functions as a list. And then you lose information about how many there are and what they are.

That is why I went for this:

```
def authenticate_2(
    authenticator1: Authenticator,
    authenticator2: Authenticator,
) -> Callable[
    [Callable[Concatenate[Request, AnyAPIUser, AnyAPIUser, _P], Coroutine[Any, Any, _T]]],
    Callable[Concatenate[Request, _P], Coroutine[Any, Any, _T]],
]:
    """Decorator for a Sanic handler that adds the APIUser when another authentication has already been done."""

    def decorator(
        f: Callable[Concatenate[Request, AnyAPIUser, AnyAPIUser, _P], Coroutine[Any, Any, _T]],
    ) -> Callable[Concatenate[Request, _P], Coroutine[Any, Any, _T]]:
        @wraps(f)
        async def decorated_function(request: Request, *args: _P.args, **kwargs: _P.kwargs) -> _T:
            token1 = request.headers.get(authenticator1.token_field)
            token2 = request.headers.get(authenticator2.token_field)
            user1: AnyAPIUser
            user2: AnyAPIUser
            [user1, user2] = await asyncio.gather(
                authenticator1.authenticate(token1 or "", request),
                authenticator2.authenticate(token2 or "", request),
            )
            response = await f(request, user1, user2, *args, **kwargs)
            return response

        return decorated_function

    return decorator
```

This way we retain all the useful info about the stuff being injected in the function. And we insure that if we want 2 authenticator then we get 2 APIUsers injected.

I dont think we will need that many but if really needed we can make `_3` and similar variants. But I think the most we will need is 2.

olevski: I already spent 3-4 hours trying to get the typing right on the above example which is simpler IMO (when it comes to mypy) than what you propose. I really dont want to spend more time wrestling with mypy.

bases/renku_data_services/data_api/app.py
```diff
@@ -134,6 +135,24 @@ def register_all_handlers(app: Sanic, config: Config) -> Sanic:
...
     )
+    notebooks = NotebooksBP(
+        name="notebooks_old",
+        url_prefix=url_prefix,
+        authenticator=config.authenticator,
+        nb_config=config.nb_config,
+        internal_gitlab_authenticator=config.gitlab_authenticator,
+        git_repo=config.git_repositories_repo,
+    )
+    notebooks_new = NotebooksNewBP(
```
sgaist, line 146: refactor suggestion: rename to sessions.
Even if the session can start notebooks as the old implementation, it will make the code clearer.
I would suggest adding a task to #303 about that.

components/renku_data_services/base_models/core.py
```diff
@@ -3,41 +3,67 @@
...
+    first_name: str | None = None
+    last_name: str | None = None
+
+
+@dataclass(kw_only=True, frozen=True)
+class AnonymousAPIUser(APIUser):
+    """The model for an anonymous user of the API."""
+
+    id: str
+    is_admin: bool = field(init=False, default=False)
```
sgaist, line 60: Shouldn't repr also be set to false ?

```suggestion
    is_admin: bool = field(init=False, default=False, repr=False)
```

components/renku_data_services/crc/db.py
```diff
@@ -183,6 +184,21 @@ async def get_resource_pools(
                 output.append(rp.dump(quota))
             return output
 
+    async def get_default_resource_class(self) -> models.ResourceClass:
+        """Get the default reosurce class in the default resource pool."""
```
sgaist, line 188:
```suggestion
        """Get the default resource class in the default resource pool."""
```

components/renku_data_services/notebooks/api/amalthea_patches/init_containers.py
```diff
@@ -189,35 +332,28 @@ def certificates(config: _NotebooksConfig) -> list[dict[str, Any]]:
...
+                        "value": api_client.sanitize_for_serialization(vol),
+                    },
+                ],
+            },
+        )
     return patches
 
 
-def download_image(server: "UserServer") -> list[dict[str, Any]]:
+def download_image_container(server: "UserServer") -> client.V1Container:
```
sgaist, line 356: Just to be sure: it's an image puller ?

components/renku_data_services/notebooks/api/classes/data_service.py
```diff
@@ -4,22 +4,27 @@
...
     InvalidComputeResourceError,
     MissingResourceError,
 )
 
-from ..schemas.server_options import ServerOptions
-from .repository import INTERNAL_GITLAB_PROVIDER, GitProvider, OAuth2Connection, OAuth2Provider
-from .user import User
-
 
 class CloudStorageConfig(NamedTuple):
```
sgaist, line 29: Just a question: any reason to use a NamedTuple rather than a Dataclass ?

components/renku_data_services/notebooks/api/classes/k8s_client.py
```diff
@@ -1,241 +1,216 @@
...
             if err.status not in [400, 404]:
-                logger.exception(f"Cannot list servers because of {err}")
+                logging.exception(f"Cannot list servers because of {err}")
                 raise IntermittentError(f"Cannot list servers from the k8s API with selector {label_selector}.")
             return []
-        return cast(list[dict[str, Any]], jss.get("items", []))
+        output: list[_SessionType] = (
+            [self.server_type.model_validate(servers.to_dict())]
+            if isinstance(servers, APIObject)
+            else [self.server_type.model_validate(i.to_dict()) for i in servers]
```
sgaist, line 201:
```suggestion
            else [self.server_type.model_validate(server.to_dict()) for server in servers]
```

components/renku_data_services/notebooks/api/classes/k8s_client.py
Panaetius: yes the title as well. it's really minor, just thought we should stick to one naming convention

Panaetius (approved):

sgaist (commented):
It's taking a good shape !

Makefile
```diff
@@ -153,7 +153,13 @@ kind_cluster:  ## Creates a kind cluster for testing
...
 	kubectl wait --namespace ingress-nginx --for=condition=ready pod --selector=app.kubernetes.io/component=controller --timeout=90s
 
-install_amaltheas:  ## Installs both version of amalthea in the currently active k8s context.
+install_amaltheas:  ## Installs both version of amalthea in the. NOTE: It uses the currently active k8s context.
 	helm repo add renku https://swissdatasciencecenter.github.io/helm-charts
-	helm install amalthea-js renku/amalthea --version $(AMALTHEA_JS_VERSION)
-	helm install amalthea-sessions renku/amalthea-sessions --version $(AMALTHEA_SESSIONS_VERSION)
+	helm repo update
+	helm upgrade --install amalthea-js renku/amalthea --version $(AMALTHEA_JS_VERSION)
+	helm upgrade --install amalthea-sessions amalthea-sessions-0.0.9-new-operator-chart.tgz --version $(AMALTHEA_SESSIONS_VERSION)
```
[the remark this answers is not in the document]
olevski: Yeah the CI and dev flow are still far from perfect. And I will improve this in followup PRs.

bases/renku_data_services/data_api/app.py
olevski: I will just add a task for this in #303 I dont want to rename this now because it may cause additional conflicts in the many PRs that branch off of this.

Makefile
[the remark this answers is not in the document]
olevski: I added a followup task in this in #303

components/renku_data_services/base_models/core.py
olevski: I dont see why we need to hide this? It is nice to see whether the user is or isnt an admin. There is nothing secret here.

components/renku_data_services/notebooks/api/classes/data_service.py
olevski: This is just what was used in the old notebooks code that we moved over. So I did not touch it. And considering that we want to eventually just fully remove this code I dont want to touch it now.

bases/renku_data_services/data_api/app.py
sgaist: Sure thing, that's why I suggested to add it to #303.
All good from my side.

components/renku_data_services/base_models/core.py
sgaist: My thinking is that as an anonymous users, only the bare minimum of information should be returned.

components/renku_data_services/notebooks/api/amalthea_patches/init_containers.py
olevski: Yes this is a container used in the old code that just runs first (or among the first) and just downloads the image. The status updates for sessions (in v1 for the old amalthea) are tied to each container in the session. So having such a container allows us to surface in our status how long it takes to just download the session image. This is not used with the new amalthea btw.

components/renku_data_services/notebooks/api/classes/data_service.py
sgaist: Fair enough

components/renku_data_services/base_models/core.py
olevski: But this is only what will show up in the logs. When we use the class we can still access that. I would just leave this as is. The most important thing is that with `init=False` and the `frozen=True` it is very hard for someone in the code to make an anonymous user where is_admin is True. I would rather be able to see in the logs that this anonymous user really had the field set to False in the logs.

sgaist: Fair enough

sgaist (approved):
LGTM

PR #377, opened 2024-08-28
chore: Send all search synchronisation events to a single queue

eikek:
Instead of using multiple queues/streams all the events are now sent to a single stream `search.sync`. They are only intended to synchronize data between data-services and search-service.

This requires a change to the `Event.queue` field. It was used for two things: the redis qeue/stream name and the message type. Now these two are different now and so the event must carry both separately. Events are first stored to the postgres database as single value (its avro payload) and the qeue name is stored alongside. When re-creating the event object from the database, we don't have the `(header, payload)` tuple, but only a `dict[str, Any]` (the raw message). This is different to when creating the event from a `message-type` and structured avro object. Since we never need to read the structured avro object of an event, the `Event` class is now simplified to only carry the raw message. This makes the generic type param obsolete and removes the early return in the `serialize` method which checked the payloads runtime type (as when read from the db).

The `serialize` method is still there to keep the interface, allowing perhaps later to add more structure to the class.

Additional to the canonical constructor of the event, there is a second one, `create`, for creating an event from a structured avro object and message type.

### DIFF
--- a/components/renku_data_services/message_queue/converters.py
+++ b/components/renku_data_services/message_queue/converters.py
@@ -1,6 +1,6 @@
 """Converter of models to Avro schemas for events."""
 
-from typing import TypeVar, cast
+from typing import Final, TypeVar, cast
 
 from dataclasses_avroschema.schema_generator import AvroModel
 
@@ -12,6 +12,12 @@ from renku_data_services.message_queue.models import Event
 from renku_data_services.namespace import models as group_models
 from renku_data_services.project import models as project_models
 from renku_data_services.users import models as user_models
+
+QUEUE_NAME: Final[str] = "search.sync"
+
+
+def _make_event(message_type: str, payload: AvroModel) -> Event:
+    return Event.create(QUEUE_NAME, message_type, payload)
 
 
 class _ProjectEventConverter:
@@ -39,7 +45,7 @@ class _ProjectEventConverter:
         match event_type:
             case v2.ProjectCreated:
                 return [
-                    Event(
+                    _make_event(
                         "project.created",
                         v2.ProjectCreated(
                             id=project_id_str,
@@ -54,7 +60,7 @@ class _ProjectEventConverter:
                             keywords=project.keywords or [],
                         ),
                     ),
-                    Event(
+                    _make_event(
                         "projectAuth.added",
                         v2.ProjectMemberAdded(
                             projectId=project_id_str,
@@ -65,7 +71,7 @@ class _ProjectEventConverter:
                 ]
             case v2.ProjectUpdated:
                 return [
-                    Event(
+                    _make_event(
                         "project.updated",
                         v2.ProjectUpdated(
                             id=project_id_str,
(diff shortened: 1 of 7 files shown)


components/renku_data_services/message_queue/converters.py
```diff
@@ -13,6 +13,12 @@
 from renku_data_services.project import models as project_models
 from renku_data_services.users import models as user_models
 
+QUEUE_NAME: Final[str] = "search.sync"
```
olevski, line 16: suggestion: I think we probably want to rename this. Because for now it is only the search but in the future it may be more services that use this same stream. So I think the name should be based on the producer and event types rather than the current consumer. I.e. something like `data_service.all_events`? Hopefully it is not too complicated to make this change now.

components/renku_data_services/message_queue/orm.py
```diff
@@ -54,9 +55,15 @@ def load(cls, event: Event) -> "EventORM":
...
 
-    def dump(self) -> Event[dict[str, Any]]:
+    def dump(self) -> Event:
         """Create an event from the ORM object."""
         message = deepcopy(self.payload)
         if "payload" in message and isinstance(message["payload"], str):
             message["payload"] = base64.b64decode(message["payload"])
         return Event(self.queue, message)
+
+    def get_message_type(self) -> str:
```
olevski, line 65: nit-pick: It may be slightly nicer to return `None` when the message type is not present. But it is really not a big deal. So please feel free to ignore this.

test/components/renku_data_services/message_queue/test_queue.py
```diff
@@ -42,21 +42,21 @@ async def fake_db_method(self, *, session, some_arg):
     fakerepo = FakeRepo()
     await fakerepo.fake_db_method(some_arg="test")
 
-    events = await app_config.redis.redis_connection.xrange("project.removed")
+    events = await app_config.redis.redis_connection.xrange("search.sync")
```
olevski, line 45: suggestion(non-blocking): You can insert `QUEUE_NAME` from `components/renku_data_services/message_queue/converters.py` here if you wish rather than using a hardcoded value.

olevski (changes requested):
Thank you Eike this looks great! 

If we can I think it would be worth to change the queue name now to something more generic that would make sense if more consumers jump on the same stream in addition to the search service.

components/renku_data_services/message_queue/converters.py
eikek: No, that is no problem! I'll change it to your suggested name. I thought we decided to use this stream now only to sync data between data-services and search. But obviously, if other are intended to read these events, the name is not good! Perhaps we then should also mark the "reprovisioning events" so consumers can distinguish them from "live events".

components/renku_data_services/message_queue/orm.py
eikek: Oh sure, that is much better, thank you!

test/components/renku_data_services/message_queue/test_queue.py
eikek: Thank you!

olevski (approved):

PR #383, opened 2024-09-02
feat: add support for oauth storage providers

Panaetius:
This PR adds three new OAuth2 providers, Google Drive, Microsoft OneDrive and Dropbox.
I tested these with some hacked together code in this PR and after confirming they worked, undid the hacky changes. As such, this PR doesn't do anything. Or rather, it'd allow admins to set up OAuth2 apps and for users to connect to them, but once connected, the users can't use them anywhere.
There will be a follow-up build including UI changes that uses these changes to automatically generate tokens for cloud storage for users.


m-alisafaee (dismissed):
LGTM, Thanks!

components/renku_data_services/errors/errors.py
```diff
@@ -45,6 +45,16 @@ class UnauthorizedError(BaseError):
     quiet: bool = True
 
 
+@dataclass
+class InvalidOAuth2Token(BaseError):
```
leafty, line 49: Why not have `class InvalidOAuth2Token(UnauthorizedError):` and only override the `message` field?

components/renku_data_services/migrations/versions/b8cbd62e85b9_add_google_drive_provider_kind.py
leafty, line 1: This migration should be collapsed with the next one.

leafty (approved):

components/renku_data_services/errors/errors.py
```diff
@@ -45,6 +45,13 @@ class UnauthorizedError(BaseError):
     quiet: bool = True
 
 
+@dataclass
+class InvalidOAuth2Token(UnauthorizedError):
+    """Raised when a token for a connected service is invalid."""
+
+    message: str = "OAuth2 token for connected service invalid or expired."
```
leafty, line 52: Non-blocking: does it make sense to add a subclass here when this seems to be used only once?

leafty (approved):
LGTM

leafty (dismissed):

leafty (dismissed):

Panaetius:
I'd missed the new migration in the recent merge and had to consolidate migration heads

leafty (approved):

PR #386, opened 2024-09-04
feat!: namespaced data connectors

leafty:
Merging into `leafty/build-data-connectors` (feature branch).

Introduce namespaced data connectors for Renku 2.0.

Details:
* Remove API routes for `/storages_v2`
* Add API routes for `/data_connectors`
* Add `data_connectors` table
* Drop the `storage_secrets` table
* Add `data_connectors_secrets` table

Out of scope for this PR:
* authorization on data connectors
* secrets for data connectors

/deploy renku=release-0.57.0


components/renku_data_services/data_connectors/api.spec.yaml
```diff
@@ -0,0 +1,635 @@
+openapi: 3.0.2
+info:
+  title: Renku Data Services API
+  description: |
+    This service is the main backend for Renku. It provides information about users, projects,
+    cloud storage, access to compute resources and many other things.
```
ciyer, line 6:
```suggestion
    data connectors, access to compute resources and many other things.
```

components/renku_data_services/data_connectors/orm.py
```diff
@@ -0,0 +1,136 @@
...
+from sqlalchemy import JSON, Boolean, DateTime, ForeignKey, MetaData, String, func
+from sqlalchemy.dialects.postgresql import ARRAY, JSONB
+from sqlalchemy.orm import DeclarativeBase, Mapped, MappedAsDataclass, mapped_column, relationship
+from sqlalchemy.schema import Index
+from ulid import ULID
+
+from renku_data_services.authz import models as authz_models
+from renku_data_services.base_orm.registry import COMMON_ORM_REGISTRY
+from renku_data_services.data_connectors import models
+from renku_data_services.data_connectors.apispec import Visibility
```
Panaetius, line 15: nitpick: I think using `from renku_data_services.authz.models import Visibility` is cleaner here, then there's no dependency here nor in db.py on the apispec. (I think there's other places in the code base that have this issue as well).

components/renku_data_services/data_connectors/db.py
```diff
@@ -0,0 +1,225 @@
...
+        async with self.session_maker() as session:
+            stmt = select(schemas.DataConnectorORM)
+            if namespace:
+                stmt = _filter_by_namespace_slug(stmt, namespace)
+            stmt = stmt.limit(pagination.per_page).offset(pagination.offset)
+            stmt = stmt.order_by(schemas.DataConnectorORM.id.desc())
+            stmt_count = select(func.count()).select_from(schemas.DataConnectorORM)
+            if namespace:
+                stmt_count = _filter_by_namespace_slug(stmt_count, namespace)
+            results = await session.scalars(stmt), await session.scalar(stmt_count)
```
Panaetius, line 40: suggestion: [asyncio.gather](https://docs.python.org/3/library/asyncio-task.html#asyncio.gather) would allow these to be run concurrently

components/renku_data_services/data_connectors/db.py
```diff
@@ -0,0 +1,225 @@
...
+        )
+        data_connector_slug = ns_schemas.EntitySlugORM.create_data_connector_slug(
+            slug, data_connector_id=data_connector_orm.id, namespace_id=ns.id
+        )
+
+        session.add(data_connector_orm)
+        await session.flush()
+        session.add(data_connector_slug)
+        await session.flush()
+        await session.refresh(data_connector_orm)
```
Panaetius, line 135: question: I overlooked this change in the project db in the previous pr https://github.com/SwissDataScienceCenter/renku-data-services/pull/384 but I thought you could do a single flush and the orm would take care of doing it in the correct order (as long as the foreign key relationships are the right way around)? Did having it with a single flush break something?

components/renku_data_services/data_connectors/db.py
```diff
@@ -0,0 +1,225 @@
...
+        result = await session.scalars(
+            select(schemas.DataConnectorORM).where(schemas.DataConnectorORM.id == data_connector_id)
+        )
+        data_connector = result.one_or_none()
+        if data_connector is None:
+            return None
+
+        await session.delete(data_connector)
+
+        return data_connector.dump()
```
Panaetius, line 213: issue: The only reason to return something here is if there's a decorator that uses the return value to send a delete message to the message queue, which isn't the case here. So a return value isn't needed.
Additionally, calling `dump()` after `delete()` can cause issues if there's lazy loaded child entities, as dump will try to load the child entities that were just deleted. This isn't an issue here, I think, but in general I started just returning the entity id in delete methods to prevent that problem and since delete messages only contain an id anyways.
So I'd either remove the return or return just the id.

Panaetius (changes requested):

components/renku_data_services/data_connectors/db.py
leafty: See https://github.com/SwissDataScienceCenter/renku-data-services/issues/404, this would be invalid use of concurrency.

components/renku_data_services/data_connectors/db.py
leafty: Yes, doing it in a single flush saves the `data_connector_slug` first for some reason, which surfaces as a foreign key constraint violation. My intuition on this is that alembic may not guess the correct order because our `ORM` classes are split across different bases. Alembic was issuing the correct order when the two tables where in the same `schema` and the `ORM` classes derived from the same base.

One potential solution may be to enforce using a single `ORM` base (though it may be annoying to define different schemas for different tables).

Panaetius: well if it causes issues then that's fine, having to talk to the db twice isn't a big enough downside to worry about it. Though we might want to explain this in a comment in code.

components/renku_data_services/data_connectors/db.py
leafty: I think messages will have to be implemented in the future (post-build), but I'll remove the `dump()` and simply return `None` or the ID.

components/renku_data_services/data_connectors/db.py
Panaetius: actually I think it might be due to using `relationship()` both on `EntitySlugORM`  and `DataConnectorORM`, without a `back_populates` on the `DataConnectorORM`. That might screw up the sqlalchemy depencency checker (as in, it just sees two unrelated `relationship`'s)
from the [docs](https://docs.sqlalchemy.org/en/20/orm/relationship_api.html#sqlalchemy.orm.relationship.params.back_populates):
> This allows objects on both sides of each [relationship()](https://docs.sqlalchemy.org/en/20/orm/relationship_api.html#sqlalchemy.orm.relationship) to synchronize in-Python state changes and also provides directives to the [unit of work](https://docs.sqlalchemy.org/en/20/glossary.html#term-unit-of-work) flush process how changes along these relationships should be persisted.

components/renku_data_services/data_connectors/orm.py
leafty: I will defer this to later because using the `Visibility` from `renku_data_services.authz.models` causes the following error:

```
asyncpg.exceptions.InvalidTextRepresentationError: invalid input value for enum visibility: "PUBLIC"
```

Panaetius: ah because that one is upper case. It's a bit weird that it takes the enum entry name instead of the string value. Maybe if that one was `StrEnum` like we use in other places it'd work.
Doing it later is fine, I think we need to clean this up in general.

components/renku_data_services/data_connectors/db.py
leafty: OK, so this was a big head-scratcher because `back_populates` is defined on both sides. The issue was that `viewonly=True` deactivates persistence related features, so `viewonly=True` had to be removed from the slug ORM to make this work.

leafty: Done in 2e4f53b7c156c74455c476d79380ddcc837e429f.

components/renku_data_services/data_connectors/db.py
leafty: Done in ceb40df67d2a0a519b334089c541d2a72453ef56.

Panaetius (approved):

PR #387, opened 2024-09-04
feat: use cloud storage from amalthea

olevski:
Adds cloud storage for the new amalthea operator and with the new csi rclone version.

/deploy renku=feat-jupyter-free-sessions amalthea-sessions=feat-add-cloud-storage renku-notebooks=feat-jupyter-free-sessions renku-gateway=feat-add-renku-refresh-token-for-data-service renku-ui=andrea/custom-launcher-jupyter-free extra-values=amalthea-sessions.deployCrd=false,amalthea-sessions.rcloneStorageClass=csi-rclone-387-secret-annotation #notest


olevski:

- #360
    - #338
        - #375
            - #387 ◀

components/renku_data_services/notebooks/blueprints.py
```diff
@@ -804,17 +809,49 @@ async def _handler(
...
+                s.storage_id: RCloneStorage(
+                    source_path=s.source_path,
+                    mount_folder=(work_dir / s.target_path).as_posix(),
+                    configuration=s.configuration,
+                    readonly=s.readonly,
+                    config=self.nb_config,
+                    name=None,
+                )
+                for s in body.cloudstorage or []
+                if s.storage_id is not None
```
Panaetius, line 844: question: historically, users could mount arbitrary, unsaved cloud storage and the UI would just post a body here, in which case it wouldn't have a storage_id. I don't think this is around in v2 but v1 might still use this. So I'm not entirely sure if this if statement will break things and we'd have to generate ad-hoc ids or if this is fine?

components/renku_data_services/notebooks/blueprints.py
```diff
@@ -834,6 +871,14 @@ async def _handler(
                 is_image_private=False,
                 internal_gitlab_user=internal_gitlab_user,
             )
+            # Generate the cloud starge secrets
```
Panaetius, line 874: nitpick: spelling

components/renku_data_services/notebooks/crs.py
```diff
@@ -32,6 +32,8 @@
 from renku_data_services.notebooks.cr_amalthea_session import Model as _ASModel
 from renku_data_services.notebooks.cr_amalthea_session import Resources3 as Resources
 from renku_data_services.notebooks.cr_amalthea_session import Secret1 as SecretAsVolume
+from renku_data_services.notebooks.cr_amalthea_session import SecretRef as SecretRefKey
+from renku_data_services.notebooks.cr_amalthea_session import SecretRef1 as SecretRefWhole
```
Panaetius, line 36: question: `SecretRef1` seems like a pretty bad name in the originating file, could this be changed or is this something autogenerated that we can't do anything about?

Panaetius (changes requested):
General Question: there's still https://github.com/SwissDataScienceCenter/renku-data-services/pull/347 that does a bit of cleanup in cloud storage and probably will cause quite a bit of conflict with this, so I'm wondering what the best way would be for us to bring this together with your changes. There's also lots of changes @leafty is working on in regards to cloud storage, and 3 people working on the same bits of code at the same time is bound to cause issues. maybe we should sync?

components/renku_data_services/notebooks/crs.py
olevski: It is auto-generated. Everything in `cr_jupyter_server.py` and `cr_amalthea_session.py` is autogenerated. That is why I import from these two files into `crs.py` and I rename there. There is no way to change the names in the autogenerated code.

Panaetius: can it be changed on the amalthea side where it comes from?

olevski: > can it be changed on the amalthea side where it comes from?

No I mean I would have to change the CRD spec for amalthea sessions just so I have a nice name when I autogenerate code. That does not make sense. And the reason there is multiple is that the code generation does not deduplicate all things. And there are multiple `SecretRef` fields in the CRD spec. So TLDR No.

Panaetius: ok, makes sense

olevski:
> General Question: there's still https://github.com/SwissDataScienceCenter/renku-data-services/pull/347 that does a bit of cleanup in cloud storage and probably will cause quite a bit of conflict with this, so I'm wondering what the best way would be for us to bring this together with your changes. There's also lots of changes @leafty is working on in regards to cloud storage, and 3 people working on the same bits of code at the same time is bound to cause issues. maybe we should sync?

When it comes to #347 I think we should split out the API breaking change and merge that at some other time. There is still a lot of testing and minor fixes that are needed on the UI as well. So if we can not have the API breaking change then we dont have to distract the UI with these changes (in addition to many already introduced with the new amalthea). You opened #347 @Panaetius do you think you can split out the API breaking change?

As for the cloud storage changes that @leafty will introduce. I think she is standing by until we merge more things. So after we get this PR and the one furtehr in the stack (#411) I think that @leafty can start targeting the same release branch we have been targeting. 

But yeah just to be safe lets have a quick chat in the afternoon tomorrow.

components/renku_data_services/notebooks/blueprints.py
olevski: For this first implementation of the cloud storage in V2 lets keep things simple and not allow people to mount cloud storage that is unsaved. Then if needed we can add this functionality. This also needs some design because in theory we could allow this in many other places in the API where we currently expect something that exists in the DB. So I will make the storage ID required in the post request to launch a session. If a payload is provided then it just overrides the cloud storage saved in our DB.

leafty:
> As for the cloud storage changes that @leafty will introduce. I think she is standing by until we merge more things. So after we get this PR and the one furtehr in the stack (#411) I think that @leafty can start targeting the same release branch we have been targeting.

If I understand things correctly, the new sessions API will not support Renku 1.0. In which case the changes should be reasonable.

leafty:
Note: waiting for a rebase, so I can try to work from this branch.

Panaetius:
> When it comes to https://github.com/SwissDataScienceCenter/renku-data-services/pull/347 I think we should split out the API breaking change and merge that at some other time.

but the API is currently broken and that it works is mostly an accident, any update to sanic or one of its components could break it. So yes, it's a breaking API change but at the same time it's more of a bugfix and unbreaking the API. I wouldn't hold off on this. And it is needed for that PR. Also https://github.com/SwissDataScienceCenter/renku-ui/pull/3299 addressed those changes on the UI AND addresses the changes to storage, so there's a bit of a triangle dependency here.

Panaetius (approved):

PR #392, opened 2024-09-10
feat: support for search reprovisioning

m-alisafaee:
Adds a Sanic background task that reads data from the database and sends them as events.

Search reprovisioning.

/deploy


components/renku_data_services/message_queue/blueprints.py
```diff
@@ -0,0 +1,80 @@
...
+
+        @authenticate(self.authenticator)
+        async def _get_status(_: Request, __: base_models.APIUser) -> JSONResponse | HTTPResponse:
+            active_reprovisioning = await self.reprovisioning_repo.get_active_reprovisioning()
+            return json({"active": bool(active_reprovisioning)}, 200)
+
+        return "/search/reprovision", ["GET"], _get_status
+
+    def delete(self) -> BlueprintFactoryResponse:
+        """Stop reprovisioning (if any)."""
```
olevski, line 72: question: Do we really want this at all? Because if you start a reprovisioning but then stop it you leave the event queue in a weird inconsistent state. So maybe once you start a reprovisioning you have to wait for it to finish or error out.

components/renku_data_services/message_queue/orm.py
```diff
@@ -71,3 +73,20 @@ def get_message_type(self) -> Optional[str]:
...
+    """Reprovisioning table.
+
+    This table is used to make sure that only one instance of reprovisioning is run at any given time.
+    It gets updated with the reprovisioning progress.
+    """
+
+    __tablename__ = "reprovisioning"
+
+    id: Mapped[ULID] = mapped_column("id", ULIDType, primary_key=True, default_factory=lambda: str(ULID()), init=False)
+    start_date: Mapped[datetime] = mapped_column("start_date", DateTime(timezone=True), nullable=False)
```
olevski, line 88: question: How do we handle errors during reprovisioning? And we have 2 kind here:
1. Errors that fully crash the whole reprovisioning process
2. Errors that may result from trying to reprovision a specific entity only

At least for no 1 we should save something in the database and report it back to the admin. For no 2 the question is whether we want to keep going if we encounter errors with some entities or do we fully stop?

And related to this is another question. What do we do if we encounter an error? Should we try to delete the / cleanup the events that were added so that we can effectively rollback to the pre-reprovisioining request state?

olevski (changes requested):
Thanks this looks really good Mohammad. I have a few questions and requests for changes. For some of the questions it may be easier to have a live discussion - but you tell me what you prefer.

components/renku_data_services/message_queue/api.spec.yaml
```diff
@@ -0,0 +1,100 @@
...
+  - url: /api/data
+  - url: /ui-server/api/data
+paths:
+  /search/reprovision:
+    post:
+      summary: Start a new reprovisioning
+      description: Only a single reprovisioning is active at any time
+      responses:
+        "201":
+          description: The reprovisioning is/will be started
```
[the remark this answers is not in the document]
m-alisafaee: I return `409` in the code to note that a reprovisioning is in progress and another one cannot be started meanwhile.

pyproject.toml
```diff
@@ -201,6 +201,7 @@ module = [
   "renku_data_services.namespace.apispec",
   "renku_data_services.notebooks.apispec",
   "renku_data_services.platform.apispec",
+  "renku_data_services.message_queue.apispec",
```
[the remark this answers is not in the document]
m-alisafaee: No, I didn't find a similar entry in `projects/renku_data_service/pyproject.toml`. There's one for poly components which already includes `message_queue`.

test/bases/renku_data_services/data_api/test_message_queue.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+@pytest.mark.asyncio
+async def test_search_get_reprovisioning_status(sanic_client, admin_headers, user_headers, monkeypatch):
+    monkeypatch.setattr(renku_data_services.message_queue.blueprints, "reprovision", long_reprovisioning_mock)
+
+    _, response = await sanic_client.get("/api/data/search/reprovision", headers=user_headers)
+
+    assert response.status_code == 200, response.text
+    assert response.json["active"] is False
```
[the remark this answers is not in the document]
m-alisafaee: Done. 404 is returned when there's no active reprovisioning.

components/renku_data_services/message_queue/blueprints.py
leafty: Some 2 cents: while a `delete` action may not be desirable, there should be a way to force a reprovisioning to be marked as errored. You cannot always ensure that a job will report back it has errored, so an admin needs to be able to mark it that way (and attempt a new one).

components/renku_data_services/message_queue/core.py
```diff
@@ -0,0 +1,69 @@
...
+    group_repo: GroupRepository,
+    project_repo: ProjectRepository,
+    authz: Authz,
+) -> None:
+    """Create and send various data service events required for reprovisioning the search index."""
+    async with session_maker() as session, session.begin():
+        start_event = make_event(
+            message_type="reprovisioning.started", payload=v2.ReprovisioningStarted(id=str(reprovisioning.id))
+        )
+        await event_repo.store_event(session, start_event)
```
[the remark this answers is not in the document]
m-alisafaee: We check if a reprovisioning is active, in the blueprint and before calling this function. So, the `reprovisioning.started` message isn't sent twice.

I'll add the DB constraint in the follow up PR.

components/renku_data_services/message_queue/api.spec.yaml
```diff
@@ -0,0 +1,115 @@
...
+          description: The date and time the reprovisioning was started (in UTC and ISO-8601 format)
+          type: string
+          format: date-time
+          example: "2023-11-01T17:32:28Z"
+      required:
+        - id
+        - start_date
+      example:
+        - id: 01BX5ZZK2KAC4AV9WEV3EMM0S0
+          start_date: "2023-11-01T17:32:28Z"
```
olevski, line 76:
```suggestion
```
This is not needed. You will get the correct example from the definition of each field for free.

Or if you want to keep it then remove the `-` and indent 2 spaces to the left. Because the response is not a list.

olevski (changes requested):
That is great just one minor thing I noticed in the apispec.

components/renku_data_services/message_queue/blueprints.py
olevski: Yeah the DELETE endpoint to stop a reprovisioning will stay.

olevski (approved):

PR #393, opened 2024-09-10
feat: do not get disabled keycloak users

wesjdj:
Enables removing disabled Keycloak user accounts from data services and Renku 2.0's search.

1. Modifies background jobs not to sync any user accounts disabled in Keycloak to data services.
2. Adds an API endpoint to allow for the deletion of user accounts from data services by providing the associated Keycloak user ID. 

When a user account is deleted using the new API endpoint, the user no longer appears in Renku 2.0's search, and references of the user are removed from the `authz` database.

/deploy


test/bases/renku_data_services/data_api/test_users.py
```diff
@@ -205,3 +205,92 @@ async def test_logged_in_user_check_adds_user_if_missing(sanic_client, users, ad
...
+            created_by="admin-id",
+        ),
+    )
+    admin_token = {
+        "id": admin.id,
+        "is_admin": True,
+        "first_name": admin.first_name,
+        "last_name": admin.last_name,
+        "email": admin.email,
+    }
```
leafty, line 232: Question: is there a specific reason to create an admin user here when `admin_user` and `admin_headers` are available test fixtures?

leafty (changes requested):
Thanks for taking care of this.

Important change needed: fix `authz` transaction change (see below).

test/bases/renku_data_services/data_api/test_users.py
wesjdj: I will modify it to use admin_user or admin_headers instead. I was following what test_get_all_users_as_admin on line 12 does for reference.

components/renku_data_services/users/db.py
```diff
@@ -124,10 +124,35 @@ async def _get_users(self, email: str | None = None) -> list[UserInfo]:
...
+    @with_db_transaction
+    @dispatch_message(avro_schema_v2.UserRemoved)
+    async def _remove_user(self, user_id: str, *, session: AsyncSession | None = None) -> str | None:
+        """Remove a user from the database."""
+        if not session:
+            raise errors.ProgrammingError(message="A database session is required")
+        logger.info(f"Trying to remove user with ID {user_id}")
+        stmt = delete(UserORM).where(UserORM.keycloak_id == user_id).returning(UserORM)
+        user = await session.scalar(stmt)
+        await self.authz._remove_user_namespace(user_id)
```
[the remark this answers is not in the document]
leafty: The original function may have been written before `@Authz.authz_change` was available.

It should be placed just below `@with_db_transaction`, for example:
```python
@with_db_transaction
@Authz.authz_change(AuthzOperation.delete, ResourceType.user)
@dispatch_message(avro_schema_v2.UserRemoved)
async def _remove_user(self, user_id: str, *, session: AsyncSession | None = None) -> str | None:
    [...]
```

Then the corresponding changes will need to exist in `components/renku_data_services/authz/authz.py`:
* Add cases in `_get_authz_change()`

Potentially open for discussion is the fact that the current `authz` changes only remove the user namespace but do not remove any permission associated with that user. I think this should probably be done, otherwise the user will still appear in membership lists.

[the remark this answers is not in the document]
leafty: I confirmed with @olevski that the current code does not remove all references of that user in the `authz` database, but we should do that (since the user ID would still be listed in groups and projects).

@wesjdj I can help update the code in `authz.py` if needed.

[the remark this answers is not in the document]
wesjdj: @leafty Thanks! I've done the initial change requested now. We'd need to look up the references in the spicedb first, I assume?

components/renku_data_services/authz/authz.py
```diff
@@ -51,7 +51,7 @@ def authz(self) -> "Authz":
 
 _AuthzChangeFuncResult = TypeVar(
     "_AuthzChangeFuncResult",
-    bound=Project | ProjectUpdate | Group | UserInfoUpdate | list[UserInfo] | None,
+    bound=Project | ProjectUpdate | Group | UserInfoUpdate | list[UserInfo] | str | None,
```
leafty, line 54: Would it be possible to use `UserInfo` as a type here? `str` could represent anything.

wesjdj: I was looking at this with @Panaetius on Monday. Iirc, because we delete the namespace as part of [`_remove_user`](https://github.com/SwissDataScienceCenter/renku-data-services/blob/0b55c7bb989ca8d29644741e64f88288ca73f856/components/renku_data_services/users/db.py#L135), of course now using the decorator, we then couldn't dump it into a new `UserInfo` object. Therefore, we modified `_remove_user` to return a string instead of a UserInfo. Maybe we could initialise the UserInfo object before actually deleting the user, but I don't know if we could make that work nicely with the Authz decorator. What do you think?

components/renku_data_services/users/db.py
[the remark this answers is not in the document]
wesjdj: Makes sense! Should we wait on merging this PR then until #425 has been merged?

components/renku_data_services/authz/authz.py
leafty: All other deletion, e.g. `Project` deletion manage to do that. I think first, the object is fetched and dumped then deleted so that the method in `db.py` can return a representation of the deleted object.

components/renku_data_services/users/db.py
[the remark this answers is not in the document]
leafty: I can also write a `_remove_user()` for this PR.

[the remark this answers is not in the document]
leafty: @wesjdj Code is available in https://github.com/SwissDataScienceCenter/renku-data-services/pull/427.

components/renku_data_services/authz/authz.py
wesjdj: Afaict, this is possible for a `Project` because in it's `dump` function, it uses `namespace=self.slug.namespace.dump()`, whereas the `User` `dump` function uses `namespace=self.namespace.dump(),`. Have there been any ideas or discussions about creating a UserSlug?

leafty: No, it is not an issue, see #427. The code which needed additional changes was to deal with the event queue (also updated there).

Panaetius: the main reason is that all the deletion messages only contain the ID and imo we should send UserInfo objects around that don't actually exist anymore, especially since the `UserInfo` objects will refer to a `Namespace` object that has its `id` set, even though it's not in the database.
To me this just sounds like a bug waiting to happen.
I agree that having `str` as a type is also very ambiguous, but the solution to me is not to reintroduce returning the object, but rather to have a [NewType](https://docs.python.org/3/library/typing.html#newtype) for the `str` and return that (e.g. `UserId` as per the example in the linked docs)

leafty: @Panaetius Let's not change how ALL the deletion methods work in the authz database NOW. You can file an issue with the suggested change, but I do not think this PR is where this needs to happen.

Panaetius: I didn't say we should do this everywhere right now...

But we shouldn't make it worse again in this PR for code that we're already touching

leafty: To me this just sounds like a bug waiting to happen.

> No it's not, it is absolutely not likely to happen that a bug would happen. Returning the deleted object is common convention in a lot of codebases and is not causing bugs.

leafty: Well, using the `id` is why we want the `Saved*` kind for Projects and Groups. I am really confused here as to why you want to insist on breaking a convention on this PR. We can create `DeletedProject`, `DeletedGroup`, etc. types and update the code in another PR. Using `str` in the middle of `model` types just for users is a big break of convention which I do not see as beneficial.

Panaetius: > Well, using the id is why we want the Saved* kind for Projects and Groups.

For me it's not about using the ID, though that certainly figures into it, but about the state of the object (is this or is this not in the DB). Which it isn't after a delete.

> I am really confused here as to why you want to insist on breaking a convention on this PR.

I'm not insisting on breaking convention in this PR. This is already in [main](https://github.com/SwissDataScienceCenter/renku-data-services/blob/main/components/renku_data_services/users/db.py#L244) and Wes is merely moving it to the proper class. And there is no convention right now imo (the two examples you gave are like the only places where it's done this way and I want to change those in the future):
- dumping the object:
   - resource pool 
   - group
   - project
- returning `None`: 
   - resource class
   - tolerations
   - affinities
   - resource pool users
   - user preferences
   - storage
   - storage secrets
   - environments
   - launchers
   - secrets
   - events
   - oauth2 clients
- returning id (in main after recent refactoring and this PR)
   - user info  
   
plus some methods that delete relationships, like membership change methods that return MembershipChange events (so not the entity that was deleted itself), and removing pinned projects that returns user preferences.

In addition, on the blueprints side there's a single DELETE endpoint that returns anything (remove pinned projects), all others return an empty body. So none of this is actually returned to a user or used elsewhere.

The only reason any of the db methods return an object on delete is to update the auth db and/or send a messages to search about the deletion, which only needs the ID (or `None` to signify no change/no need to send a message, though it's debatable if that's the best approach). 

So the dumping of the whole object is pretty pointless in my opinion, we transform data that just isn't used anywhere. And doing it that way and making sure the semantics of the types hold would dumping to different, new types like `DeletedProject` that would just exist to hold semantic meaning but whose properties wouldn't be used other than the id (like `DeletedProject`) or just returning the ID/None (which we could typealias to e.g. `DeletedProjectId` so it's not just str). Or `DeletedProject` would probably just contain the ID anyways as we don't use any other properties of deleted entities, so there's no meaningful difference between that and a simple type alias imo.

And the way it's currently done can break, due to the lazy loading and the way a lot of the code currently does it. E.g. for [project](https://github.com/SwissDataScienceCenter/renku-data-services/blob/main/components/renku_data_services/project/db.py#L312) the dump here would fail if there was a lazy loaded child property that gets cascade-deleted as it wouldn't be there anymore when the dump is called. The workaround is to dump it before and make sure no saved object is leaked, but again, that's extra work for data that is ultimately not used.

So I'm arguing in favor of having it simpler and not do unnecessary extra work by just returning ID's in the cases where this is needed for authz/message sending. And in other cases that already just return `None`, I think it's fine keeping that but we could also return the IDs in all delete methods for consistency. 

But in this PR specifically, for this one method, I just don't see a reason to undo a change that's already in main and that follows an approach that I'd like to change other methods to in the future to make things more consistent, given that it's already not really consistent. For me it'd just mean that it needs to get changed back again in that future refactoring.

But I'm happy to discuss this further in person.

Panaetius: https://github.com/SwissDataScienceCenter/renku-data-services/issues/442

Panaetius (approved):

leafty (approved):

PR #394, opened 2024-09-10
feat: add support for filtering project and group listing by direct membership

andre-code:
Add support to filter the project and group list endpoints by using the query parameter `direct_member`.
When used, the group or project list will contain only items the current user is a direct member of.

/deploy


components/renku_data_services/project/db.py
```diff
@@ -55,15 +55,20 @@ async def get_projects(
...
+            stmt = select(schemas.ProjectORM).where(schemas.ProjectORM.id.in_(project_ids))
             if namespace:
                 stmt = _filter_by_namespace_slug(stmt, namespace)
+
+            stmt = stmt.order_by(
+                case(
+                    {schemas.ProjectORM.updated_at.is_not(None): schemas.ProjectORM.updated_at},
+                    else_=schemas.ProjectORM.creation_date,
+                ).desc()
+            )
```
Panaetius, line 68:
```suggestion
from sqlalchemy.sql.functions import coalesce
[...]
            stmt = stmt.order_by(
                coalesce(
                    schemas.ProjectORM.updated_at,
                    schemas.ProjectORM.creation_date
                ).desc()
            )
```

Panaetius (dismissed):
Looks great, just a minor simplification.

components/renku_data_services/authz/authz.py
```diff
@@ -358,6 +358,37 @@ async def resources_with_permission(
                 ids.append(response.resource_object_id)
         return ids
 
+    async def resource_ids_for_user_membership(
```
leafty, line 361: Why add this when `resources_with_permission` exists? (function just above)

You can see it used to filter namespaces by role here -> https://github.com/SwissDataScienceCenter/renku-data-services/blob/7ae9526c8f0ebcc78f23a2c390b6d07503d45053/components/renku_data_services/namespace/db.py#L298.

leafty:
@andre-code is this PR ready? If yes, please fix the PR title.

components/renku_data_services/authz/authz.py
andre-code: The function is needed to properly get the projects where the user is a member with only Viewer permissions. If I use resources_with_permission with `Scope.WRITE`, it leaves out projects where the user is just a Viewer. But if I use `Scope.READ`, it brings in public projects the user can access, even though they’re not an owner or member.

leafty (changes requested):
This change cannot be accepted as is, it hides groups and projects the user has the permission to view from the API.

Please add a query parameter to support this filtering of groups and projects. You can call it `direct_member` -> `GET /api/data/projects?direct_member=1`.

andre-code:
> This change cannot be accepted as is, it hides groups and projects the user has the permission to view from the API.
> 
> Please add a query parameter to support this filtering of groups and projects. You can call it `direct_member` -> `GET /api/data/projects?direct_member=1`.

Sure, that makes sense. I've incorporated the suggestions, and it's ready for your review. I'll continue working on getting the tests to pass. Thanks for your review.

components/renku_data_services/authz/authz.py
```diff
@@ -358,6 +358,35 @@ async def resources_with_permission(
...
 
+    async def resources_with_direct_membership(
+        self, user: base_models.APIUser, resource_type: ResourceType
+    ) -> list[str]:
+        """Get all the resource IDs (for a specific resource kind) that a specific user is a direct member of."""
+        resource_ids: list[str] = []
+        if user.id is None:
+            return resource_ids
+
+        sub = SubjectReference(object=(_AuthzConverter.to_object(ResourceType.user, user.id)))
```
leafty, line 369: This `sub` does not seem to be used:

```suggestion
```

components/renku_data_services/authz/authz.py
```diff
@@ -358,6 +358,35 @@ async def resources_with_permission(
...
+
+        rel_filter = RelationshipFilter(
+            resource_type=resource_type.value,
+            optional_subject_filter=SubjectFilter(
+                subject_type=ResourceType.user.value, optional_subject_id=sub.object.object_id
```
leafty, line 374:
```suggestion
                subject_type=ResourceType.user.value, optional_subject_id=user.id
```

components/renku_data_services/namespace/api.spec.yaml
```diff
@@ -525,6 +525,18 @@ components:
...
             description: A minimum role to filter results by.
             $ref: "#/components/schemas/GroupRole"
+    GroupsGetQuery:
+      description: Query params for namespace get request
+      allOf:
+      - $ref: "#/components/schemas/PaginationRequest"
+      - properties:
+          minimum_role:
+            description: A minimum role to filter results by.
+            $ref: "#/components/schemas/GroupRole"
```
leafty, line 535: This is not implemented, so it should not appear in the spec:

```suggestion
```

components/renku_data_services/namespace/db.py
```diff
@@ -74,14 +74,20 @@ async def generate_user_namespaces(self, *, session: AsyncSession | None = None)
...
         async with self.session_maker() as session, session.begin():
-            stmt = select(schemas.GroupORM).limit(pagination.per_page).offset(pagination.offset)
-            stmt = stmt.order_by(schemas.GroupORM.creation_date.asc(), schemas.GroupORM.id.asc(), schemas.GroupORM.name)
+            stmt = select(schemas.GroupORM)
+            if direct_member:
+                stmt = stmt.where(schemas.GroupORM.id.in_(group_ids))
+            stmt = stmt.limit(pagination.per_page).offset(pagination.offset)
+            stmt = stmt.order_by(
+                schemas.GroupORM.creation_date.desc(), schemas.GroupORM.id.desc(), schemas.GroupORM.name
+            )
```
leafty, line 90: You can sort by `id.desc()` here. IDs are unique and sorting them will correspond to the time the ID was created.

```suggestion
            stmt = stmt.order_by(schemas.GroupORM.id.desc())
```

leafty (changes requested):
Implementation seems to be OK for the most part.

Tests for the updated endpoints are missing and should be added, e.g. `test_get_groups_with_direct_membership` and `test_get_projects_with_direct_membership`.

leafty:
> Tests for the updated endpoints are missing and should be added, e.g. `test_get_groups_with_direct_membership` and `test_get_projects_with_direct_membership`.

For example, here is a test for projects:
```python
@pytest.mark.asyncio
async def test_get_projects_with_direct_membership(sanic_client, user_headers, member_1_headers, member_1_user) -> None:
    # Create a group
    namespace = "my-group"
    payload = {
        "name": "Group",
        "slug": namespace,
    }
    _, response = await sanic_client.post("/api/data/groups", headers=user_headers, json=payload)
    assert response.status_code == 201, response.text
    # Create some projects in the group
    payload = {
        "name": "Project 1",
        "namespace": namespace,
    }
    _, response = await sanic_client.post("/api/data/projects", headers=user_headers, json=payload)
    assert response.status_code == 201, response.text
    project_1 = response.json
    payload = {
        "name": "Project 2",
        "namespace": namespace,
    }
    _, response = await sanic_client.post("/api/data/projects", headers=user_headers, json=payload)
    assert response.status_code == 201, response.text
    project_2 = response.json
    # Add member_1 to the group
    roles = [{"id": member_1_user.id, "role": "editor"}]
    _, response = await sanic_client.patch(f"/api/data/groups/{namespace}/members", headers=user_headers, json=roles)
    assert response.status_code == 200, response.text
    # Add member_1 to Project 2
    roles = [{"id": member_1_user.id, "role": "editor"}]
    _, response = await sanic_client.patch(
        f"/api/data/projects/{project_2["id"]}/members", headers=user_headers, json=roles
    )
    assert response.status_code == 200, response.text

    parameters = {"direct_member": True}
    _, response = await sanic_client.get("/api/data/projects", headers=member_1_headers, params=parameters)

    assert response.status_code == 200, response.text
    projects = response.json
    assert len(projects) == 1
    project_ids = {p["id"] for p in projects}
    assert project_ids == {project_2["id"]}

    # Check that both projects can be seen without the filter
    _, response = await sanic_client.get("/api/data/projects", headers=member_1_headers)
    projects = response.json
    assert len(projects) == 2
    project_ids = {p["id"] for p in projects}
    assert project_ids == {project_1["id"], project_2["id"]}
```

andre-code:
> Implementation seems to be OK for the most part.
> 
> Tests for the updated endpoints are missing and should be added, e.g. `test_get_groups_with_direct_membership` and `test_get_projects_with_direct_membership`.

Thanks @leafty   for providing an example of the test! I’ve included the one for groups, along with your other suggestions.

leafty (approved):
Code looks good now 👍 

Before merging, please update the PR title to:
```
feat: add support for filtering project and group listing by direct membership
```
and the PR description to:
```
Add support to filter the project and group list endpoints by using the query parameter `direct_member`.
When used, the group or project list will contain only items the current user is a direct member of.
```

And use these when merging to `main`. These offer a good description of the contents of this PR.

PR #422, opened 2024-09-23
feat: migrate storages_v2 to data_connectors

leafty:
Merging into `leafty/build-data-connectors` (feature branch).

Add a background job to migrate "Data Sources" (`/storages_v2`) to "Data Connectors" (`/data_connectors`). Each storage item has a new data connector created and linked to the source project.

See for manual testing: https://github.com/SwissDataScienceCenter/renku/pull/3797
Logs from migration job:
```
INFO:background_jobs.migrate_storages_v2_to_data_connectors:Migrating 6 cloud storage v2 items to data connectors.
INFO:background_jobs.migrate_storages_v2_to_data_connectors:Migrated giab-1 to flora.thiebaut/giab-1.
INFO:background_jobs.migrate_storages_v2_to_data_connectors:Deleted storage_v2: 01J9KGAW3V27VERY7S7W71VX7D
INFO:background_jobs.migrate_storages_v2_to_data_connectors:Migrated giab-1 to flora.thiebaut/giab-1-uv38rbtp.
INFO:background_jobs.migrate_storages_v2_to_data_connectors:Deleted storage_v2: 01J9KGBPRVS1RWYMMYGKEZ8VVW
INFO:background_jobs.migrate_storages_v2_to_data_connectors:Migrated giab-2 to flora.thiebaut/giab-2.
INFO:background_jobs.migrate_storages_v2_to_data_connectors:Deleted storage_v2: 01J9KGC5549CHKFKKX3W7XYQVT
INFO:background_jobs.migrate_storages_v2_to_data_connectors:Migrated giab-group to group-test-1/giab-group.
INFO:background_jobs.migrate_storages_v2_to_data_connectors:Deleted storage_v2: 01J9KGDH5108975M95HXK228ED
INFO:background_jobs.migrate_storages_v2_to_data_connectors:Migrated giab-group to group-test-1/giab-group-5hph7nxd.
INFO:background_jobs.migrate_storages_v2_to_data_connectors:Deleted storage_v2: 01J9KGEF9XXBNPMP17NKZR0SD1
INFO:background_jobs.migrate_storages_v2_to_data_connectors:Migrated giab to group-test-2/giab.
INFO:background_jobs.migrate_storages_v2_to_data_connectors:Deleted storage_v2: 01J9KGHFAVT6KECE6WQE40XHR9
INFO:background_jobs.migrate_storages_v2_to_data_connectors:Migrated 6/6 data connectors.
```


bases/renku_data_services/background_jobs/core.py
```diff
@@ -264,3 +264,38 @@ async def migrate_user_namespaces_make_all_public(config: SyncConfig) -> None:
...
+            logger.error(err)
+            failed_storages.append(str(storage.storage_id))
+            if first_error is None:
+                first_error = err
+
+    logger.info(f"Migrated {len(storages_v2)-len(failed_storages)}/{len(storages_v2)} data connectors.")
+    if failed_storages:
+        logger.error(f"Migration failed for storages: {failed_storages}.")
+    if first_error is not None:
+        raise first_error
```
Panaetius, line 301: issue: I don't think we want to raise here. We had it in the past where some migration raised an error, which caused the job to fail without doing latter migrations. `fix_mismatched_project_namespace_ids` failed and so the migrating groups jobs were never executed which had caused issues until we rolled out a fix a week later. The ugly quick fix back then was to just not raise when that error happened, but the whole background jobs code is still kinda brittle in this regard.

This also gets a bit tricky to manage in cases where a latter migration should only be done if a previous one passed, though I don't think we have such dependencies at the moment?

In any case, I think we should collect exceptions for each of the tasks in the short_period_sync job and only raise at the end once all things have been done. This does not mean you need to do that in this PR for all the migration tasks, if you don't want to, a follow-up issue is fine, but until we have a proper solution we should not raise in a migration task directly.

And I know that right now it doesn't matter since this is the last migration in the list, but for sure the next person along will just add another one after this one without considering it.

components/renku_data_services/data_connectors/migration_utils.py
```diff
@@ -0,0 +1,214 @@
...
+
+        # Link the data connector to the project
+        unsaved_link = models.UnsavedDataConnectorToProjectLink(
+            data_connector_id=data_connector.id,
+            project_id=project_id,
+        )
+        await self.data_connector_project_link_repo.insert_link(user=data_connector_owner, link=unsaved_link)
+
+        # Remove the storage_v2 from the database
+        await self._delete_storage_v2(requested_by=requested_by, storage_id=storage.storage_id)
```
Panaetius, line 83: suddestion: should we do this in a db transaction? If this fails somewhere in the middle, it could leave the db in a weird state.

components/renku_data_services/data_connectors/migration_utils.py
```diff
@@ -0,0 +1,214 @@
...
+
+    async def _get_all_project_ids(self, requested_by: base_models.APIUser) -> list[ULID]:
+        """Get all Renku 2.0 projects."""
+        if requested_by.id is None:
+            raise errors.UnauthorizedError(message="You do not have the required permissions for this operation.")
+        if not requested_by.is_admin:
+            raise errors.ForbiddenError(message="Only admins can perform this operation.")
+
+        async with self.session_maker() as session:
+            stmt = select(projects_schemas.ProjectORM)
```
Panaetius, line 211: nitpick: you don't really need to load the whole objects and dump them, you can just do `select(projects_schemas.ProjectORM).options(load_only(projects_schemas.ProjectORM.id))` to only get a list of IDs and return those.

Panaetius (changes requested):

components/renku_data_services/data_connectors/migration_utils.py
leafty: This would not be easy to achieve. The code use methods from repositories when they are appropriate, which use their own session and transaction. Having the whole thing wrapped in a transaction means I would have to refactor all the methods used or to copy them here to be compatible with a parent transaction.

Panaetius: Maybe I'm missing something, but _insert_data_connector and _delete_storage_v2 are already in this class, so there's only `insert_link` that's elsewhere, which uses the `with_db_transaction` decorator which supports passing in a transaction.

leafty: OK, so I tried to wrap all the database changes inside a single transaction but this does not work because of the `Authz.authz_change()` decorator. The decorator performs `session.commit()` after the wrapped function returns, so that means that calling a method with the `authz_change` decorator has to be the last action on a db transaction. In this case, both `insert_data_connector` and `insert_link` have authz changes tied to them with the decorator, so it is not possible to call both of them within a single transaction.

bases/renku_data_services/background_jobs/core.py
leafty: Done in dcc3080584e3618b7b41e67a19a51402e6771b4b.

components/renku_data_services/data_connectors/migration_utils.py
leafty: Done in f45f3e4cf8dd80ed3f2cf232a03bdf706dbd6854.

components/renku_data_services/data_connectors/migration_utils.py
Panaetius: thanks for trying. I think then it's ok, it's a pretty low chance of being an issue anyways.

Panaetius (approved):

Panaetius (approved):

PR #423, opened 2024-09-23
refactor: introduce @validate_body_root_model

leafty:
Add `@validate_body_root_model` which is a drop-in replacement for Sanic's `@validate` when validating a `RootModel` or a model deriving from it.

The decorator should be replaced by `@validate` once https://github.com/sanic-org/sanic-ext/issues/198 is fixed.

/deploy


Panaetius (approved):
thanks for this!

I have two bugfix PRs open in sanic (one addresses this issue) for over a month now, without any reply. So don't have too high hopes of them fixing this any time soon on their end...

PR #425, opened 2024-09-25
refactor: define a single method to remove entities from authz

leafty:
Removing an entity from `spicedb` is equivalent to removing all relationships where the entity appears as a resource or as a subject. This means we should not need special cases for each entity type.

This refactor should also simplify the code for user removal (remove user entity + remove namespace entity). See https://github.com/SwissDataScienceCenter/renku-data-services/pull/393.


leafty:
Note: I am not sure if this will work for deleting users, as the permission check would probably fail.

PR #427, opened 2024-09-26
add suggestions for PR #393

leafty:


components/renku_data_services/authz/authz.py
```diff
@@ -1363,3 +1367,35 @@ async def _remove_user_namespace(self, user_id: str, zed_token: ZedToken | None
             updates=[RelationshipUpdate(operation=RelationshipUpdate.OPERATION_TOUCH, relationship=i) for i in rels]
         )
         return _AuthzChange(apply=apply, undo=undo)
+
+    async def _remove_user(
```
wesjdj, line 1371: I could be wrong here, but should this be a synchronous function because it queries the spicedb for the relationships?

wesjdj: Ah gotcha, I had it the wrong way around

PR #434, opened 2024-10-01
fix: allow session launcher parameters to be reset

olevski:
Allows the API to accept None as input for args, command and the session launcher resource class ID so that they can be reset to their defaults in patch endpoints.

### DIFF
--- a/components/renku_data_services/authn/dummy.py
+++ b/components/renku_data_services/authn/dummy.py
@@ -70,7 +70,7 @@ class DummyAuthenticator:
         )
 
         return base_models.APIUser(
-            is_admin_init=user_props.get("is_admin", False),
+            is_admin=user_props.get("is_admin", False),
             id=user_props.get("id", "some-id") if is_set else None,
             access_token=access_token,
             first_name=user_props.get("first_name", "John") if is_set else None,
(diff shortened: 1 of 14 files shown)


olevski:

- #434 ◀

Panaetius (changes requested):
We had to solve the same issue in renku-python in the past, but we solved it with a specific object instance, not a type: https://github.com/SwissDataScienceCenter/renku-python/blob/develop/renku/domain_model/constant.py#L23

The benefit is that you can just use object id equality instead of isinstance checks, so it functions much more like regular `None` does than just a type. E.g. https://github.com/SwissDataScienceCenter/renku-python/blob/develop/renku/domain_model/constant.py#L23

I.e. there's only ever a single `NO_VALUE` object that contains nothing, except for the python object id, which is the same.

This is a common idiom for sentinel values, there is a PEP to have better support for sentinels but it's still a draft: https://peps.python.org/pep-0661/

The downside is that the `object()` approach doesn't have a nice repr, though not sure if we care about that.

You can use a class with a repr, but then it's important that you use the type as the value, not an instance, to have it behave like `None` does. This is discussed in a bit more detail in https://python-patterns.guide/python/sentinel-object/ and https://www.revsys.com/tidbits/sentinel-values-python/

In any case, we should use one of the more standard approaches.

olevski:
@Panaetius ok I will switch to something like the example from renku python.

PR #439, opened 2024-10-04
fix: docker image check endpoint and gitlab authn

olevski:
/deploy renku=feat-jupyter-free-sessions extra-values=amalthea-sessions.deployCrd=false


olevski:

- #434
    - #439 ◀

components/renku_data_services/notebooks/api/classes/image.py
```diff
@@ -29,16 +29,20 @@ class ImageRepoDockerAPI:
 
     hostname: str
     oauth2_token: Optional[str] = field(default=None, repr=False)
+    # NOTE: We need to follow redirects so that we can authenticate with the image repositories properly.
+    # NOTE: If we do not use default_factory to create the client here requests will fail beause it can happen
```
Panaetius, line 33:
```suggestion
    # NOTE: If we do not use default_factory to create the client here requests will fail because it can happen
```

components/renku_data_services/notebooks/api/classes/image.py
```diff
@@ -54,56 +58,55 @@ def _get_docker_token(self, image: "Image") -> Optional[str]:
...
         return str(token_req.json().get("token"))
 
-    def get_image_manifest(self, image: "Image") -> Optional[dict[str, Any]]:
+    async def get_image_manifest(self, image: "Image") -> Optional[dict[str, Any]]:
         """Query the docker API to get the manifest of an image."""
         if image.hostname != self.hostname:
-            raise ImageParseError(
-                f"The image hostname {image.hostname} does not match " f"the image repository {self.hostname}"
+            raise errors.ValidationError(
+                message=f"The image hostname {image.hostname} does not match " f"the image repository {self.hostname}"
```
Panaetius, line 68:
```suggestion
                message=f"The image hostname {image.hostname} does not match the image repository {self.hostname}"
```

components/renku_data_services/notebooks/blueprints.py
```diff
@@ -1125,3 +1125,24 @@ async def _handler(
...
+    def check_docker_image(self) -> BlueprintFactoryResponse:
+        """Return the availability of the docker image."""
+
+        @authenticate_2(self.authenticator, self.internal_gitlab_authenticator)
+        async def _check_docker_image(
+            request: Request, user: AnonymousAPIUser | AuthenticatedAPIUser, internal_gitlab_user: APIUser
+        ) -> HTTPResponse:
+            image_url = request.get_args().get("image_url")
+            if not isinstance(image_url, str):
+                raise ValueError("required string of image url")
```
Panaetius, line 1138: suggestion: should this be a ValidationError instead? ValueError is not handled by our error handler.
Also, you can validate query args with `@validate(query=...)` which would let you check that it's a str in a more standard way

Panaetius (changes requested):

components/renku_data_services/notebooks/blueprints.py
olevski: done

Panaetius (approved):

olevski:
This fails on the KG failing to index a project launched with renku v1. So I will disregard this acceptance tests as we did not touch anything related to that in this PR.

PR #443, opened 2024-10-07
fix: update data connectors migrations

leafty:
Fixes two issues with the migrations for data connectors:
1.  The migrations cannot be rolled back after creating a data connector. This was caused by leftover entries in the `entity_slugs` table.
2. The entity slug constraint was not created. The change was not picked up by Alembic; it seems Alembic can only handle `CheckConstraint` creation together with table creation, but not new constraints on an existing table, updates to them, or removing them from a table.

Manually tested:
1. Create data connectors with the deployment at `HEAD`
2. Connect to a `data-service` pod and run
    ```bash
    /app/env/bin/python3 -m alembic -c /app/env/lib/python3.12/site-packages/renku_data_services/migrations/alembic.ini --name common downgrade 726d5d0e1f28
    ```
3. Check that the current revision is `726d5d0e1f28`, that is before data connectors
4. Re-apply migrations

Note that this series of actions would fail prior to this fix.

/deploy #notest

### DIFF
--- a/components/renku_data_services/migrations/versions/3cf2adf9896b_add_data_connectors.py
+++ b/components/renku_data_services/migrations/versions/3cf2adf9896b_add_data_connectors.py
@@ -73,6 +73,12 @@ def upgrade() -> None:
         referent_schema="storage",
         ondelete="CASCADE",
     )
+    op.create_check_constraint(
+        "either_project_id_or_data_connector_id_is_set",
+        "entity_slugs",
+        "CAST (project_id IS NOT NULL AS int) + CAST (data_connector_id IS NOT NULL AS int) BETWEEN 0 AND 1",
+        schema="common",
+    )
     op.create_table(
         "data_connector_secrets",
         sa.Column("user_id", sa.String(length=36), nullable=False),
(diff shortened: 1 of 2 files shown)


components/renku_data_services/migrations/versions/3cf2adf9896b_add_data_connectors.py
```diff
@@ -135,6 +141,8 @@ def downgrade() -> None:
         schema="storage",
     )
     op.drop_table("data_connector_secrets", schema="storage")
+    op.drop_constraint("either_project_id_or_data_connector_id_is_set", "entity_slugs", schema="common", type_="check")
+    op.execute("""DELETE FROM common.entity_slugs WHERE entity_slugs.data_connector_id IS NOT NULL""")
```
Panaetius, line 145: nitpick: I don't think this needs to be triple quoted.

leafty, line 145:
```suggestion
    op.execute("DELETE FROM common.entity_slugs WHERE entity_slugs.data_connector_id IS NOT NULL")
```

Panaetius (approved):

PR #445, opened 2024-10-07
refactor: use dedicated entities for deletion function returns

### LINKED ISSUE #442: return dedicated Deleted* type in db delete methods
Right now, a few delete methods (Projects, Group, ResourcePool) return dumped entities that represent persisted entities, even though the entity is not persisted anymore.

For example, https://github.com/SwissDataScienceCenter/renku-data-services/blob/a449135365cb0638eefe85f6a7044ca71db99bbb/components/renku_data_services/project/db.py#L290 returns `models.Project` but that entity has the semantics of "Project stored in the database". And you'd be allowed to use that type in methods that expect an existing database entry, even though no such entry exists anymore (like you could call `update` on them), which is from from a type system standpoint.

In addition, we don't actually use any of these entities values in the caller, only the ID, so it's a bit superfluous to fetch and transform all the fields ant potentially nested objects as a return value only to then just read the id of the deleted object and drop the data.

We should have dedicated return types (like a type alias for the id e.g. `DeletedProjectId: TypeAlias = str` or a `DeletedProject` entity with just an `id` field) that we return from `delete` methods in db repositories that make semantic sense, allow type-checking messages and authz changes properly and don't return unused data.

Panaetius:
closes #442


Panaetius:

- #445 ◀

leafty (commented):
Thanks for doing this @Panaetius! Can we also include a `DeletedDataConnector` here?

leafty (approved):
Looks great!

leafty (approved):

PR #449, opened 2024-10-10
feat: add documentation to projects

olloz26:
/deploy amalthea-sessions=main renku=feat-jupyter-free-sessions extra-values=amalthea-sessions.deployCrd=false


components/renku_data_services/project/blueprints.py
```diff
@@ -36,6 +36,25 @@ class ProjectsBP(CustomBlueprint):
     user_repo: UserRepo
     authenticator: base_models.Authenticator
 
+    def _project2dict(self, project: project_models.Project, with_documentation: bool = False) -> dict[str, Any]:
```
Panaetius, line 39: nitpick: I'd just put this outside of the class, as `self` is not actually used

components/renku_data_services/project/blueprints.py
```diff
@@ -112,29 +101,17 @@ def get_one(self) -> BlueprintFactoryResponse:
         @authenticate(self.authenticator)
         @validate_path_project_id
         async def _get_one(request: Request, user: base_models.APIUser, project_id: str) -> JSONResponse | HTTPResponse:
-            project = await self.project_repo.get_project(user=user, project_id=ULID.from_str(project_id))
+            with_documentation = bool(request.args.get("with_documentation", False))
```
Panaetius, line 104: issue: use query string validation instead with the apispec model (same as here: https://github.com/SwissDataScienceCenter/renku-data-services/blob/5b095d795f96247aa559a3083947283c534c355d/components/renku_data_services/project/blueprints.py#L43 )

Panaetius (changes requested):
looks great! just two smaller nitpicks

Panaetius (approved):

Issue #453, opened 2024-10-11
Add information in the API about the permissions on `Group` and `Project` entities


leafty:
The UI needs to know what are the current user's permissions on a group or project (or data connector). Currently, the UI tries to compute that information but fails to do so correctly (because it does not replicate cascading rules).

Add endpoints to compute permissions for the current user, so that the UI can know which role(s) are given a group or project ("viewer", "editor", "owner").

PR #454, opened 2024-10-11
feat: support getting permissions

leafty:
See: #453.

/deploy #notest


components/renku_data_services/project/blueprints.py
```diff
@@ -239,7 +242,10 @@ async def _get_all_members(_: Request, user: base_models.APIUser, project_id: st
                 user_id = member.user_id
                 user_info = await self.user_repo.get_user(id=user_id)
                 if not user_info:
-                    raise errors.MissingResourceError(message=f"The user with ID {user_id} cannot be found.")
+                    raise errors.MissingResourceError(
+                        message=f"The user with ID {
+                                                      user_id} cannot be found."
```
Panaetius, line 247: issue: not sure what's going on with the formatting here :slightly_smiling_face:

Panaetius (changes requested):
looks all good except for one small issue.

I am wondering though if we should return this just as part of project GET instead of (or in addition to?) a separate endpoint? I assume the UI wants to know the permission every time it get a project?

components/renku_data_services/project/blueprints.py
leafty: I think it's `black` formatting from a vscode extension. 🙃

leafty:
> I am wondering though if we should return this just as part of project GET instead of (or in addition to?) a separate endpoint? I assume the UI wants to know the permission every time it get a project?

I wanted to consider this, but then it means we would have to compute and send them on project list endpoints too. It may be fast enough that it's all fine I guess.

components/renku_data_services/project/api.spec.yaml
```diff
@@ -515,6 +541,16 @@ components:
...
+    ProjectPermissions:
+      description: The set of permissions on a project
+      type: object
+      properties:
+        write:
+          description: The user can edit the project
+          type: boolean
+        admin:
+          description: The user can delete the project and manage members
+          type: boolean
```
olevski, line 553: suggestion: I think it is better if this follows the scopes we have in authz.models more closely.

And currently the Scope enum is:

```
class Scope(Enum):
    """Types of permissions - i.e. scope."""

    READ: str = "read"
    WRITE: str = "write"
    DELETE: str = "delete"
    CHANGE_MEMBERSHIP: str = "change_membership"
    READ_CHILDREN: str = "read_children"
```

So it is better to replace `admin` with `delete` and `change_membership`.

Panaetius:
> I wanted to consider this, but then it means we would have to compute and send them on project list endpoints too. It may be fast enough that it's all fine I guess.

We hit authz in both cases anyways so we should theoretically be able to get this in one go.

But I'm also fine with it being a separate endpoint. Was just wondering what would be nicer in terms of ergonomics for clients (i.e. the UI). Since you're the one that will have to live with it, I leave it up to you :slightly_smiling_face:

leafty:
> We hit authz in both cases anyways so we should theoretically be able to get this in one go.

I don't think it is possible because the way we get the project ids is with `LookupResources` while the permission check for each level is done with `CheckPermission`. For a single project, we do a `CheckPermission`, but this is done with `read` and we want to know about the other ones. Maybe there would be a way to reduce the amount of network I/O, but it is not trivial. And also, I think it does not matter for now.

leafty:
> I am wondering though if we should return this just as part of project GET instead of (or in addition to?) a separate endpoint? I assume the UI wants to know the permission every time it get a project?

I had a look at adding a `permissions` field to the `project` API schema, but this breaks the `ETag` system. The `ETag` is computed only from the database project entity at the moment while permission changes are linked to authz. This means the permissions could change without affecting the etag of a project. While it is possible to have everything working with `permissions` as a new field in `project`, it requires a much bigger set of changes compared to adding a new API endpoint for them.

Panaetius (approved):

PR #461, opened 2024-10-15
refactor: change all serial id columns to be GENERATED AS IDENTITY ones

Panaetius:
/deploy


Panaetius:

- #461 ◀

olevski (changes requested):
This looks great Ralf. As discussed though lets add the locks for each table just to be safe. Thanks for finding the proper fix for this.

olevski (approved):

olevski (approved):

leafty:
Can we make sure this works on an already-populated deployment? Make a test PR in `renku`, deploy prior to this migration, then add some data (e.g. add a resource class), then migrate and check that items can still be created.

Panaetius:
I did that in the original PR before splitting off this change and locally, but just did it again in https://github.com/SwissDataScienceCenter/renku/pull/3820 to double check (Adding projects with repositories) and it works fine.

PR #462, opened 2024-10-16
fix: find images when there is an oci index

olevski:
This was causing the API to consider some images as non-existing. Whereas in reality they exist. I noticed this with this image: ghcr.io/salimkayal/renku-devcontainer-wizard:latest.


sgaist (approved):

olevski:
This fix was also applied in notebooks at: https://github.com/SwissDataScienceCenter/renku-notebooks/pull/1995

SalimKayal (approved):
LGTM

Panaetius (approved):

PR #463, opened 2024-10-16
fix: handle errors in background jobs nicely

### LINKED ISSUE #447: handle exceptions in background_jobs properly
We perform various migrations in background jobs, but if one step has an unhandled exception, none of the following steps will run. This has caused issues in the past.

At the same time, we want to know if there is something wrong, so we want the k8s job to fail if there is an issue so we know there's something wrong.

To this end, we should catch and collect exceptions in all migration steps and continue on errors, and then raise an exception at the end of the whole job if there were any exceptions.

Panaetius:
closes #447
/deploy


bases/renku_data_services/background_jobs/core.py
```diff
@@ -290,7 +291,11 @@ async def migrate_storages_v2_to_data_connectors(config: SyncConfig) -> None:
...
             logger.error(err)
             failed_storages.append(str(storage.storage_id))
+            if first_error is None:
+                first_error = err
 
     logger.info(f"Migrated {len(storages_v2)-len(failed_storages)}/{len(storages_v2)} data connectors.")
     if failed_storages:
         logger.error(f"Migration failed for storages: {failed_storages}.")
+    if first_error is not None:
+        raise first_error
```
leafty, line 301: Question: Would there be a way to raise an exception which somehow combines all exceptions which happen during the job?

bases/renku_data_services/background_jobs/main.py
```diff
@@ -23,21 +24,39 @@ async def short_period_sync() -> None:
...
+    async with error_handler:
+        await sync_admins_from_keycloak(config.kc_api, Authz(config.authz_config))
+    async with error_handler:
+        await fix_mismatched_project_namespace_ids(config)
+    async with error_handler:
+        await migrate_groups_make_all_public(config)
+    async with error_handler:
+        await migrate_user_namespaces_make_all_public(config)
+    async with error_handler:
+        await migrate_storages_v2_to_data_connectors(config)
```
leafty, line 43: I wonder if we could make this a bit more elegant.
```python
promises = [
  bootstrap_user_namespaces(config),
  config.syncer.events_sync(config.kc_api),
  ...
]

for promise in promises:
  async with error_handler:
    await promise
```

bases/renku_data_services/background_jobs/utils.py
```diff
@@ -0,0 +1,54 @@
...
+        self, exc_type: type[BaseException] | None, exc_value: BaseException | None, traceback: TracebackType | None, /
+    ) -> bool:
+        """Exit function called when the context manager exits."""
+        if exc_value:
+            self.errors.append(exc_value)
+
+        return True
+
+    def maybe_raise(self) -> None:
+        """If this handler captured any errors, we raise them here. Otherwise to nothing."""
```
leafty, line 52:
```suggestion
        """If this handler captured any errors, we raise them here. Otherwise do nothing."""
```

bases/renku_data_services/background_jobs/utils.py
```diff
@@ -0,0 +1,54 @@
...
+
+    message = "Background job failed:"
+    errors: list[BaseException]
+
+    def __format_errors__(self) -> str:
+        """Format contained errors for output."""
+        result = []
+        for error in self.errors:
+            result.append("".join(traceback.TracebackException.from_exception(error).format()))
+        return "\n".join(result)
```
leafty, line 21: Out of curiosity: Do you have an example of how this appears in the logs? Hopefully reading multiple errors is feasible and not hard to process.

leafty (commented):
Thanks, code looks good. There are a couple of comments below.

bases/renku_data_services/background_jobs/utils.py
Panaetius: certainly:
```
❯ python -m asyncio
asyncio REPL 3.12.5 (main, Aug 13 2024, 11:21:56) [GCC 14.2.1 20240805] on linux
Use "await" directly instead of "asyncio.run()".
Type "help", "copyright", "credits" or "license" for more information.
>>> import asyncio
>>> from renku_data_services.background_jobs.utils import ErrorHandler
>>> eh = ErrorHandler()
>>> async with eh:
...     raise ValueError()
...
>>> async with eh:
...     raise OverflowError()
...
>>> eh.maybe_raise()
  File "/home/zenon/.pyenv/versions/3.12.5/lib/python3.12/concurrent/futures/_base.py", line 456, in result
    return self.__get_result()
           ^^^^^^^^^^^^^^^^^^^
  File "/home/zenon/.pyenv/versions/3.12.5/lib/python3.12/concurrent/futures/_base.py", line 401, in __get_result
    raise self._exception
  File "/home/zenon/.pyenv/versions/3.12.5/lib/python3.12/asyncio/__main__.py", line 33, in callback
    coro = func()
           ^^^^^^
  File "<console>", line 1, in <module>
  File "/home/zenon/DEV/ETH/renku-data-services/bases/renku_data_services/background_jobs/utils.py", line 54, in maybe_raise
    raise BackgroundJobError(errors=self.errors)
renku_data_services.background_jobs.utils.BackgroundJobError: BackgroundJobError: Background job failed:
  File "<console>", line 2, in <module>
ValueError

  File "<console>", line 2, in <module>
OverflowError
```

bases/renku_data_services/background_jobs/main.py
Panaetius: I originally considered ding a loop but then thought that there could be cases where job2 shouldn't run if job1 fails, which would be
```
async with error_handler:
    await job1()
    await job2()
```

You could achieve the same in a loop probably by composing the jobs or having separate loops. But I thought the straight forward but repetitive way is the easiest to understand.

In the end,m this is essentially a monad, so I can look a bit if there's an elegant way of having monads in python.

leafty: It was a non-blocking comment anyway. I can try to make a PR later.

bases/renku_data_services/background_jobs/utils.py
leafty: Thanks. It think it would be nice if we could have:
```
❯ python -m asyncio
asyncio REPL 3.12.5 (main, Aug 13 2024, 11:21:56) [GCC 14.2.1 20240805] on linux
Use "await" directly instead of "asyncio.run()".
Type "help", "copyright", "credits" or "license" for more information.
>>> import asyncio
>>> from renku_data_services.background_jobs.utils import ErrorHandler
>>> eh = ErrorHandler()
>>> async with eh:
...     raise ValueError()
...
>>> async with eh:
...     raise OverflowError()
...
>>> eh.maybe_raise()
  File "/home/zenon/.pyenv/versions/3.12.5/lib/python3.12/concurrent/futures/_base.py", line 456, in result
    return self.__get_result()
           ^^^^^^^^^^^^^^^^^^^
  File "/home/zenon/.pyenv/versions/3.12.5/lib/python3.12/concurrent/futures/_base.py", line 401, in __get_result
    raise self._exception
  File "/home/zenon/.pyenv/versions/3.12.5/lib/python3.12/asyncio/__main__.py", line 33, in callback
    coro = func()
           ^^^^^^
  File "<console>", line 1, in <module>
  File "/home/zenon/DEV/ETH/renku-data-services/bases/renku_data_services/background_jobs/utils.py", line 54, in maybe_raise
    raise BackgroundJobError(errors=self.errors)
renku_data_services.background_jobs.utils.BackgroundJobError: BackgroundJobError: Background job failed:
=== Error 1/2 ===
  File "<console>", line 2, in <module>
ValueError
=== Error 2/2 ===
  File "<console>", line 2, in <module>
OverflowError
```

Or similar. Do not see it as a blocking comment though.

bases/renku_data_services/background_jobs/utils.py
```diff
@@ -0,0 +1,137 @@
...
+                for f in fn:
+                    result = await f
+                    handler.errors.extend(result.errors)
+
+            else:
+                handler = await fn
+        except BaseException as e:
+            self.errors.append(e)
+        else:
+            self.errors.extend(handler.errors)
```
leafty, line 78: I'm confused by the `else` here. What does `try / else` mean?

bases/renku_data_services/background_jobs/utils.py
leafty, line 1: I am not sure if I find this easier to read than the previous version.

leafty (commented):
See comment about `try / else`.

bases/renku_data_services/background_jobs/utils.py
leafty, line 78: Shouldn't this be the other way around? First match on `BackgroundJobError` and extend the list and if not, match on `BaseException` and append to the current list.

Panaetius:
```
try:
   [...]
except:
   [A]
else:
   [B]
```
executes A if there's an exception and B it there was no exception.

bases/renku_data_services/background_jobs/utils.py
```diff
@@ -0,0 +1,137 @@
...
+        except BaseException as e:
+            self.errors.append(e)
+        return self
+
+    async def flatmap(
+        self, fn: "Coroutine[None,None,ErrorHandlerMonad]|list[Coroutine[None,None,ErrorHandlerMonad]]"
+    ) -> Self:
+        """Await a single or list of coroutines that return ErrorHandlerMonads and track an errors if they occur.
+
+        For a list of coroutines, if any of them fail, subsequent ones aren't executed.
```
leafty, line 63: This doesn't see to be the case. In the `for` loop, the code appends the new errors to `handler`, but that does not stop the loop from executing the subsequent item from `fn`.

leafty: I guess this is down to what `a coroutine fails` is meant to convey: that it utterly fail or that the returned `ErrorHandlerMonad` contains errors.

leafty (dismissed):
After review, the code looks correct. There are two comments I will leave:

1. Maybe the `list[Coroutine]` cases should be removed to make the code simpler to read. It does not seem to be used anyway. The caller can chain `.map` or `.flatmap` call when needed or combine code into a single coroutine as desired.
2. It feels that introducing a `BackgroundJob` type (with `async run(self) -> None`) would achieve the same result but be more idiomatic. Python does unfortunately not offer any syntactic sugar for monad and as such are very verbose to use.

leafty (approved):
Nice!

leafty (changes requested):
Actually, this would probably execute all tasks at the same time!

leafty (approved):

PR #464, opened 2024-10-17
fix: remove potentially troublesome __eq__ from resource pools

### LINKED ISSUE #435: Remove __eq__ from resource pools
This is used only in testing and it makes little sense to use outside of tests. It may even be dangerous if we forget to update it properly or in some cases that are ambiguous.

Panaetius:
closes #435


Panaetius:

- #464 ◀

leafty (approved):
LGTM, thanks!

PR #466, opened 2024-10-17
fix: accept only lowercase slugs

olevski:
This also makes it so that we can return uppercase slugs for projects - because we did not fully lowercase and check for this before. But now we cannot save or modify any existing project slugs that have uppercase in them. Whereas if there are any projects with uppercase slugs we can still list, patch, delete them.


test/bases/renku_data_services/data_api/conftest.py
```diff
@@ -213,7 +213,10 @@ async def create_project_helper(
         headers = admin_headers if admin else user_headers
         user = admin_user if admin else regular_user
         payload = payload.copy()
-        payload.update({"name": name, "namespace": f"{user.first_name}.{user.last_name}"})
+        if "name" not in payload:
+            payload.update({"name": name})
+        if "namespace" not in payload:
+            payload.update({"namespace": f"{user.first_name}.{user.last_name}"}.lower())
```
Panaetius, line 219:
```suggestion
            payload.update({"namespace": f"{user.first_name}.{user.last_name}".lower()})
```

Panaetius (changes requested):
looks good other than a minor typo

Panaetius (dismissed):

olevski:
@Panaetius I fixed all the tests now.

Panaetius (approved):

PR #472, opened 2024-10-18
feat: include is_admin in the self /user endpoint

### LINKED ISSUE #287: Update the `/user` endpoint to include `is_admin`
Add the `is_admin` field to the API response.

See https://github.com/SwissDataScienceCenter/renku-ui/issues/3223: `renku-ui` can drop API queries to Keycloak, but to do so requires support for knowing whether the current user is an admin or not.

leafty:
Closes #287.


components/renku_data_services/users/api.spec.yaml
```diff
@@ -343,6 +343,29 @@ components:
       items:
         $ref: "#/components/schemas/UserWithId"
       uniqueItems: true
+    SelfUserInfo:
```
leafty, line 346: Note: `schemathesis` doesn't like `allOf` -> https://github.com/SwissDataScienceCenter/renku-data-services/actions/runs/11403216174/job/31729934175?pr=472.

Panaetius (approved):

PR #476, opened 2024-10-22
test: only load apispec one, don't recalculate routes on each request

Panaetius:
looking at other open PRs right now, running (main) tests takes around 32 minutes.
With these changes, this is down to 18 minutes.

We can further improve this by around -50% if we find a way to have the sanic client be session scoped (one sanic client per worker, not one per test). This should not cause any problems, but would require a significant refactor in how we get the function-scoped database and other `Config` properties into blueprints without rerunning the `register_all_handlers` method. Like we'd need a way to swap out the database connection of repositories per test without constructing a new sanic client. I've created #477 to further investigate this.

For reference, these are the speedups i saw on my machine for main tests with two workers, in seconds:
before memoizing yaml loading: 2205
after memoizing yaml: 1434
after the reusable sanic client (plus yaml memoizing): 1304

### DIFF
--- a/components/renku_data_services/app_config/config.py
+++ b/components/renku_data_services/app_config/config.py
@@ -9,6 +9,7 @@ a single connection will be reused. This allows for the configuration classes to
 instantiated multiple times without creating multiple database connections.
 """
 
+import functools
 import os
 import secrets
 from dataclasses import dataclass, field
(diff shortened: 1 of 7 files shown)


leafty (dismissed):
Very nice!

Panaetius:
@leafty sorry I just saw that I left in some testing code where I wanted to not spin up a new authz subprocess for each test that din't pan out, but forgot to remove.

leafty:
> @leafty sorry I just saw that I left in some testing code where I wanted to not spin up a new authz subprocess for each test that din't pan out, but forgot to remove.

Oh, the tests were green so I thought it worked :sweat_smile:

leafty (approved):

Panaetius:
> > @leafty sorry I just saw that I left in some testing code where I wanted to not spin up a new authz subprocess for each test that din't pan out, but forgot to remove.
> 
> Oh, the tests were green so I thought it worked 😅

ah it wasn't actually use anywhere :smile:

PR #478, opened 2024-10-23
feat: add data connectors

olevski:
Enables the data connectors to work with the new amalthea.

NOTE: The target branch is not main.

/deploy renku=feat-jupyter-free-sessions amalthea-sessions=main renku-ui=andrea/jupyter-free-build extra-values=amalthea-sessions.deployCrd=false


components/renku_data_services/data_connectors/db.py
```diff
@@ -478,11 +478,74 @@ def __init__(
...
+            raise errors.UnauthorizedError(message="You do not have the required permissions for this operation.")
+
+        can_read_project = await self.authz.has_permission(user, ResourceType.project, project_id, Scope.READ)
+        if not can_read_project:
+            raise errors.MissingResourceError(
+                message=f"The project ID with {project_id} does not exist or you dont have permission to access it"
+            )
+
+        async with self.session_maker() as session:
+            stmt = (
```
leafty, line 505: Some 2 cents, feel free to ignore: it feels like it would be simpler to update the ORM classes to allow for the secrets to be selectively retrieved here, instead of doing this complex query by hand.

olevski: I assume you mean you just use `relationship` on the data connector ORM class so that you can then query for the project connectors and for the secrets? If yes then I added this.

leafty: It looks better now, though the second `.where()` seems superfluous.

olevski:
NOTE: Saved secrets do not work until this is merged and released: https://github.com/SwissDataScienceCenter/csi-rclone/pull/37

But if you do not save the storage secret and then type it in at session start then things will work as expected.

components/renku_data_services/data_connectors/db.py
```diff
@@ -477,11 +477,47 @@ def __init__(
...
+                    schemas.DataConnectorToProjectLinkORM.project_id == project_id
+                ),
+                or_(
+                    # Data connectors with secrets for the specific user
+                    schemas.DataConnectorORM.secrets.any(
+                        schemas.DataConnectorSecretORM.user_id == user.id,
+                    ),
+                    # Data connectors without any secrets
+                    # See: https://docs.sqlalchemy.org/en/20/orm/queryguide/select.html#exists-forms-has-any
+                    ~schemas.DataConnectorORM.secrets.any(),
```
leafty, line 515: I'm pretty confused by that part of the statement. Why include it at all? If you do the following:
```python
stmt = select(schemas.DataConnectorORM).where(
                schemas.DataConnectorORM.project_links.any(
                    schemas.DataConnectorToProjectLinkORM.project_id == project_id
                )
            )
```

Would that not select all data connectors linked to a project?

olevski: Yes you are right. I will change it and then "manually" filter the secrets.

leafty: You can just use the `get_data_connector_secrets()` method below.

components/renku_data_services/notebooks/api.spec.yaml
```diff
@@ -1015,19 +1015,15 @@ components:
...
         storage_id:
           allOf:
             - "$ref": "#/components/schemas/Ulid"
-            - description: The storage ID is used to know which storage config from the DB should be overriden
+            - description: If the storage_id is provided then this config must replace an existing storage config in the session
       required:
-        - configuration
-        - source_path
-        - target_path
         - storage_id
```
leafty, line 1027: Note: it would probably make sense to update `storage_id` to be `data_connector_id` in the future.

components/renku_data_services/notebooks/apispec.py
```diff
@@ -259,10 +259,10 @@ class SessionLogsResponse(RootModel[Optional[Dict[str, str]]]):
...
 
 class SessionCloudStoragePost(BaseAPISpec):
-    configuration: Dict[str, Any]
-    readonly: bool = True
-    source_path: str
-    target_path: str
+    configuration: Optional[Dict[str, Any]] = None
+    readonly: Optional[bool] = None
+    source_path: Optional[str] = None
+    target_path: Optional[str] = None
```
leafty, line 265: nitpick ([PEP604](https://peps.python.org/pep-0604/)):
```suggestion
    configuration: dict[str, Any] | None = None
    readonly: bool | None = None
    source_path: str | None = None
    target_path: str | None = None
```

components/renku_data_services/notebooks/blueprints.py
```diff
@@ -269,59 +288,67 @@ async def _handler(
...
                     )
-                cloud_storage[csr_id] = csr
+                if csr.target_path is not None and not PurePosixPath(csr.target_path).is_absolute():
+                    csr.target_path = (work_dir / csr.target_path).as_posix()
+                # TODO: The UI always does this check if it is acceptable/safe
+                # if csr_id in dcs_secrets and csr.configuration is not None:
+                #     raise errors.ValidationError(
+                #         message=f"Overriding the storage configuration for storage with ID {csr_id} "
+                #         "is not allowed because the storage has an associated saved secret.",
+                #     )
```
leafty, line 337: Note: I am not sure about the validity of this comment. What if both `access_key_id` and `secret_access_key` are required, but I only saved `access_key_id`?

components/renku_data_services/notebooks/util/kubernetes_.py
```diff
@@ -62,9 +62,9 @@ def renku_2_make_server_name(safe_username: str, project_id: str, launcher_id: s
     # Must be no more than 63 characters because the name is used to create a k8s Service and Services
     # have more restrictions for their names beacuse their names have to make a valid hostname.
     # NOTE: We use server name as a label value, so, server name must be less than 63 characters.
-    # !NOTE: For now we limit the server name to a max of 42 characters.
-    # NOTE: This is 12 + 9 + 21 = 42 characters
-    return f"{prefix[:12]}-renku-2-{server_hash[:21]}"
+    # !NOTE: For now we limit the server name to a max of 25 characters.
+    # NOTE: This is 12 + 1 + 12 = 25 characters
+    return f"{prefix[:12]}-{server_hash[:12]}"
```
leafty, line 67: Now that we have access to all the data-services information, it would be nice to have a human-readable session name:
```
{prefix}-{ns_slug}-{project_slug}-{launcher_slug}-{hash}
```

Also, it would be nicer to try to use `Slug.from_name` for the `prefix` part to avoid percent-encoded characters in it.

leafty (approved):
Code looks good.

Question: Is this deployment fully functional? It would be nice to test the sessions with different combinations of data connectors (with saved or unsaved secrets, etc.).

components/renku_data_services/notebooks/api.spec.yaml
olevski: I did not want to change the api on purpose. But yes we should do this in the future.

components/renku_data_services/notebooks/apispec.py
olevski: This is auto generated code. And it has this old style of typing everywhere - not just in notebooks. There is a `--use-union-operator` option that I can add. But I will do that in a followup PR on main. Rebasing this and the feature branch against main is already a pain. This will make it harder.

components/renku_data_services/notebooks/api.spec.yaml
olevski: I opened an issue #494

components/renku_data_services/notebooks/blueprints.py
olevski: Yes I meant to remove this. I will remove it now.

components/renku_data_services/notebooks/apispec.py
leafty: Oh, this should be dismissed then.

components/renku_data_services/notebooks/util/kubernetes_.py
olevski: Yes I have been meaning to do this. And now I am using `Slug` as you suggested. We actually have `Slug.from_user` which fits this pretty well. I just still have to do some extra cleaning because k8s is picky about the names.

components/renku_data_services/notebooks/blueprints.py
olevski: Yes I meant to remove this before requesting a review but I forgot.

olevski:
> Question: Is this deployment fully functional? It would be nice to test the sessions with different combinations of data connectors (with saved or unsaved secrets, etc.).

The only thing that does not work here is decrypting and using saved secrets for data connectors. It should work though once the fix I submitted for the csi rclone is rolled out on dev (the PR for this is merged and it will be released with renku 0.60.0). Rolling out a parallel version of rclone just for a CI deployment does not really work because both rclones step over one another.

leafty (commented):
Trying to use the deployment from the UI, this is what I get from trying to launch a session:
```json
{
    "error": {
        "code": 1422,
        "message": "There are errors in the following fields, name: String should match pattern '^[a-z]([-a-z0-9]*[a-z0-9])?$'"
    }
}
```

olevski:
> Trying to use the deployment from the UI, this is what I get from trying to launch a session

Yes I just saw that. I am fixing it rn.

PR #418, opened 2024-09-20
[REBASE only]: Migrate notebooks and add new amalthea

olevski:
This combines all changes for the notebooks migration and new amalthea operator.

**NOTE: This should be rebased on main. DO NOT SQUASH AND MERGE. WHEN updating this branch remember to REABASE only too.**

/deploy renku=feat-jupyter-free-sessions amalthea-sessions=main renku-ui=andrea/jupyter-free-build renku-gateway=master extra-values=amalthea-sessions.deployCrd=false

PR #480, opened 2024-10-23
chore: remove validate_path_project_id

olevski:
This is not needed because we have a custom validator for the type ulid in sanic that we added at some point. So just annotating the path segment as type <project_id:ulid> will work the same.

Here is where the format is registered with Sanic: https://github.com/SwissDataScienceCenter/renku-data-services/blob/main/bases/renku_data_services/data_api/app.py#L31

### DIFF
--- a/components/renku_data_services/base_api/auth.py
+++ b/components/renku_data_services/base_api/auth.py
@@ -44,30 +44,6 @@ def authenticate(
     return decorator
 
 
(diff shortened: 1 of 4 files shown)


Panaetius:
this is already done in https://github.com/SwissDataScienceCenter/renku-data-services/pull/418

PR #481, opened 2024-10-23
test: don't recreate sanic routes for every test

### LINKED ISSUE #477: make `SanicReusableASGITestClient` session scoped to speed up tests
Currently, we construct a new `SanicReusableASGITestClient` for each test execution (e.g. the `sanic_client` fixture is function scoped), which carries a large overhead in terms of performance.

This is due to sanic's way of handling routes, where it generates parser code on server start for route handling, a relatively expensive operation (roughly 3.5 seconds each time, executed 272 times, which amounts to 15 minutes out of the ~20 minutes test execution time).

If we made `SanicReusableASGITestClient` session scope, this would drastically reduce test times. The main blocker is that we need to have a new test DB for each test, which we currently do by creating a new app Config object for each test, which is in turn needed for creating the sanic client and setting up routes, creating a circular dependency.

If we instead injected the database connection for each test, we should be able to have a static config object with only the connection in repositories changing. There might also be other approaches, such as using [update_config](https://sanic.dev/en/guide/running/configuration.html#from-a-class-or-object) on sanic for each test, though that probably carries a larger risk of weird state/exceptions.

You can see this taking up over 75% of test execution time in this pytest-profiling output (`utils:28:__aenter__` in yellow and below is route setup/calculation and general sanic server setup):
image

Panaetius:
This speeds up tests significantly by not having to do sanic route calculations for every test. It also contains some fixes to authz writes not being awaited. The main change is only creating the sanic test client once per test session, but then having newly created databases injected for each test to ensure isolation. This speeds things up significantly since creating a sanic client takes around 3-4 seconds.

Downsides are that it needs a global async event loop for tests (well, for each runner). This doesn't break things for us but can end up with hard to debug errors when e.g. a fixture fails and inadvertently closes the eventloop, which then causes a subsequent test to fail.
It also makes the fixtures more complex.

closes #477


Panaetius:
this PR currently has an issue when running with multiple workers. Everything is fine when running a single worker, as can be seen here: https://github.com/SwissDataScienceCenter/renku-data-services/actions/runs/11495703773/job/31995959209?pr=481
Execution with a single worker takes 7:37 minutes, as opposed to 18 minutes with two workers without this PR. That's up to 80% potential reduction in compute time (minus some loss for scheduling workers when parallelizing)

pyproject.toml
```diff
@@ -94,6 +94,7 @@ debugpy = "^1.8.7"
 pytest-xdist = { version = "^3.5.0", extras = ["psutil"] }
 types-requests = "^2.32.0.20241016"
 types-aiofiles = "^24.1.0.20240626"
+uvloop = "^0.21.0"
```
olevski, line 97: Shouldn't `uvloop` be a regular dependency? Because when we run the server we run with uvloop?

test/utils.py
```diff
@@ -1,17 +1,251 @@
...
+        user_store = DummyUserStore(user_always_exists=user_always_exists)
+        gitlab_client = DummyGitlabAPI()
+        dummy_users = [
+            user_preferences_models.UnsavedUserInfo(
+                id="user1", first_name="user1", last_name="doe", email="user1@doe.com"
+            ),
+            user_preferences_models.UnsavedUserInfo(
+                id="user2", first_name="user2", last_name="doe", email="user2@doe.com"
+            ),
+        ]
```
olevski, line 201: suggestion(non-blocking): I think these were hardcoded before but it would be nicer if these were also fixtures and they were added as part of the fixture that initializes the TestAppConfig.

test/bases/renku_data_services/data_api/test_data_connectors.py
```diff
@@ -248,7 +249,15 @@ async def test_post_data_connector_with_conflicting_slug(
 @pytest.mark.asyncio
 @pytest.mark.parametrize("headers_and_error", [("unauthorized_headers", 401), ("member_1_headers", 403)])
 async def test_post_data_connector_without_namespace_permission(
-    sanic_client: SanicASGITestClient, user_headers, headers_and_error, request
+    # NOTE: dynamically requesting async fixtures with an already running event loop causes errors in pytest.
+    # to prevent this, all used fixtures have to also be listed again, so the exist at test execution time and
```
olevski, line 253:
```suggestion
    # to prevent this, all used fixtures have to also be listed again, so they exist at test execution time and
```

test/bases/renku_data_services/data_api/test_data_connectors.py
```diff
@@ -368,7 +377,15 @@ async def test_get_one_by_slug_data_connector(
 @pytest.mark.asyncio
 @pytest.mark.parametrize("headers_name", ["unauthorized_headers", "member_1_headers"])
 async def test_get_one_data_connector_unauthorized(
-    sanic_client: SanicASGITestClient, create_data_connector, headers_name, request
+    # NOTE: dynamically requesting async fixtures with an already running event loop causes errors in pytest.
+    # to prevent this, all used fixtures have to also be listed again, so the exist at test execution time and
```
olevski, line 381:
```suggestion
    # to prevent this, all used fixtures have to also be listed again, so they exist at test execution time and
```

olevski (changes requested):
Thanks for making these changes Ralf! It looks good just very minor changes. Lets just bring up the shortcoming you explained with the rest of the team so that we are all on the same page and no one is surprised.

pyproject.toml
Panaetius: well it's really a depenceny of sanic at runtime, but in tests we call it directly to make sure everything runs in the same loop. We don't use uvloop directly in application code as far as I'm aware, so I thought it best to have it as a dependency for tests

olevski (dismissed):

olevski (dismissed):

olevski (approved):

Issue #485, opened 2024-10-25
Replace update methods in `db.py` files to not use untyped `dict[str, Any]` input.


leafty:
Using `dict[str, Any]` prevents type-checking from `mypy` and can cause bugs, e.g. #483 and #484.

List:
- [ ] TODO

PR #486, opened 2024-10-25
refactor: use typed patch instance for project updates

leafty:
See #485.

Refactor the project update method to used a typed `ProjectPatch` instance. This means that the update code is now type-checked.

/deploy

### DIFF
--- a/components/renku_data_services/project/db.py
+++ b/components/renku_data_services/project/db.py
@@ -5,8 +5,7 @@ from __future__ import annotations
 import functools
 from collections.abc import AsyncGenerator, Awaitable, Callable
 from datetime import UTC, datetime
-from enum import Enum
-from typing import Any, Concatenate, ParamSpec, TypeVar
+from typing import Concatenate, ParamSpec, TypeVar
 
 from sqlalchemy import Select, delete, func, select, update
 from sqlalchemy.ext.asyncio import AsyncSession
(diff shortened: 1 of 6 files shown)


Panaetius:
typo in the title of the PR :)

components/renku_data_services/project/db.py
```diff
@@ -285,9 +263,27 @@ async def update_project(
...
+        if patch.repositories is not None:
+            project.repositories = [
+                schemas.ProjectRepositoryORM(url=r, project_id=project_id_str, project=project)
+                for r in patch.repositories
+            ]
+            # Trigger update for ``updated_at`` column
+            await session.execute(update(schemas.ProjectORM).where(schemas.ProjectORM.id == project_id).values())
+        if patch.description is not None:
+            project.description = patch.description if patch.description else None
+        if patch.keywords is not None:
```
Panaetius, line 285: question: I think this would currently work but curious if you tested it. Can keywords be unset by sending an empty list `[]` ?

Panaetius (approved):
very nice. I had one question but feel free to ignore if that works.

components/renku_data_services/project/db.py
Panaetius: same for description with `""`. other fields can't be empty I think

leafty: Yes, it works. There is a corresponding test for project keywords. Manual testing also shows it works for the description (sending "" resets the description).

Issue #294, opened 2024-07-03
Reading groups and namespaces do not always check AuthzDB


olevski:
We made all groups and namespaces public for reading.

But in some cases when we list or read them we check the authorization DB and in some cases we dont.

It would be nice to make it so that we always check the authorization database for this. Even though the groups/namespaces are public. Because if we dont do this we could decide to change the permissions in the future but because we do not always check authzed then the permissions we add in authz may not be always enforced.

olevski:
For example see `get_group` and `get_groups` functions in `namespace/db.py`.

PR #487, opened 2024-10-25
refactor: use typed patch instance for group updates

leafty:
See #485.

Apparently, only the group creator was able to update the group. This is now fixed. -> Related to #294.

### DIFF
--- a/components/renku_data_services/namespace/core.py
+++ b/components/renku_data_services/namespace/core.py
@@ -3,7 +3,7 @@
 from renku_data_services.namespace import apispec, models
 
 
-def validate_project_patch(patch: apispec.GroupPatchRequest) -> models.GroupPatch:
+def validate_group_patch(patch: apispec.GroupPatchRequest) -> models.GroupPatch:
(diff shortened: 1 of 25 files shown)


components/renku_data_services/namespace/core.py
```diff
@@ -0,0 +1,12 @@
+"""Business logic for groups and namespaces."""
+
+from renku_data_services.namespace import apispec, models
+
+
+def validate_project_patch(patch: apispec.GroupPatchRequest) -> models.GroupPatch:
```
Panaetius, line 6:
```suggestion
def validate_group_patch(patch: apispec.GroupPatchRequest) -> models.GroupPatch:
```

Panaetius (changes requested):
nice! just one copy-paste mistake

components/renku_data_services/namespace/core.py
leafty: Oh thanks

Panaetius (approved):

PR #490, opened 2024-10-28
refactor: make resource pool initialisation async

### LINKED ISSUE #100: Make resource pool initialization async
The loading of the default resource pools in the db is not async. But it would be nice if it was.

https://github.com/SwissDataScienceCenter/renku-data-services/pull/97#discussion_r1407379041

Panaetius:
closes #100 

/deploy


bases/renku_data_services/data_api/main.py
```diff
@@ -69,7 +69,7 @@ def create_app() -> Sanic:
         # NOTE: in single process mode where we usually run schemathesis to get coverage the db migrations
         # specified below with the main_process_start decorator do not run.
         run_migrations_for_app("common")
-        config.rp_repo.initialize(config.db.conn_url(async_client=False), config.default_resource_pool)
+        asyncio.run(config.rp_repo.initialize(config.db.conn_url(async_client=False), config.default_resource_pool))
```
leafty, line 72: Non-blocking question: should this task be done here? This may predate the existence of the background jobs but now that we have them, wouldn't it make sense to do this type of task there?

Panaetius: this job needs quite a bit of information from the chart, like k8s namespace, default resource pool, default resource pool file and I'd rather not modify the code and the chart for this. I did remove three other methods here as two of them run in the background job anyways and the third could be added easily.

leafty: OK, good :+1:. Maybe we can track this as a lower priority task (that is, to move this init method to a background task).

Panaetius: https://github.com/SwissDataScienceCenter/renku-data-services/issues/492

leafty (approved):
LGTM

PR #491, opened 2024-10-29
fix: add creation date and created by to namespaces

### LINKED ISSUE #283: Add creation date to namespace in postgres
Related to #277 

We never save the creation date for namespaces in the db. We should fix this.

Panaetius:
closes #283


components/renku_data_services/namespace/orm.py
```diff
@@ -74,21 +75,43 @@ class NamespaceORM(BaseORM):
     )
     user: Mapped[UserORM | None] = relationship(lazy="joined", init=False, repr=False, viewonly=True)
 
+    @hybridproperty
```
leafty, line 78: Why use `hybridproperty` and not the standard `property` here? There does not seem to be documentation available on this decorator.

components/renku_data_services/namespace/orm.py
```diff
@@ -74,21 +75,43 @@ class NamespaceORM(BaseORM):
...
+        raise errors.ProgrammingError(
+            message=f"Found a namespace {self.slug} that has no group or user associated with it."
+        )
+
+    @hybridproperty
+    def creation_date(self) -> datetime | None:
+        """When this namespace was created."""
+        if self.group is not None:
+            return self.group.creation_date
+        return None
```
leafty, line 95: I wonder if we can get this from Keycloak somehow. Maybe we should store it?

components/renku_data_services/namespace/orm.py
leafty: Quite confusingly, there is documentation about `sqlachemy.ext.hybrid.hybrid_property` -> https://docs.sqlalchemy.org/en/20/orm/extensions/hybrid.html.
But using that decorator does not work, or requires the use of `inplace.expression`.

components/renku_data_services/db_config/config.py
```diff
@@ -53,6 +53,7 @@ def async_session_maker(self) -> Callable[..., AsyncSession]:
                 self.conn_url(),
                 pool_size=10,
                 max_overflow=0,
+                echo=True,
```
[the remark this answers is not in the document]
Panaetius: sorry this PR wasn't actually ready, I was still testing what queries the hybrid property generates.

components/renku_data_services/namespace/orm.py
Panaetius: yes, code completion error, sqlalchemy has both and this is the wrong one (this one doesn't have any documentation).

According to the docs the hybrid_property does not necessarily need an inplace, if only properties on the ORM are accessed in simple ways it can just work. But with relations that's not the case.

Hybrid properties first because I wanted to see how they work and also because it'd allow filtering on them in queries instead of having to calculate the values in python code.

This works though not exactly how I'd hoped. It creates queries like:
```
SELECT common.namespaces.id, common.namespaces.slug, common.namespaces.group_id, common.namespaces.user_id, groups_1.id AS id_1, groups_1.name, groups_1.created_by, groups_1.creation_date, groups_1.description, users_1.keycloak_id, users_1.first_name, users_1.last_name, users_1.email, users_1.secret_key, users_1.id AS id_2
FROM common.namespaces LEFT OUTER JOIN common.groups AS groups_1 ON groups_1.id = common.namespaces.group_id LEFT OUTER JOIN users.users AS users_1 ON users_1.keycloak_id = common.namespaces.user_id, common.groups
WHERE coalesce(common.groups.created_by, common.namespaces.user_id) = $1::VARCHAR ORDER BY common.namespaces.id
 LIMIT $2::INTEGER OFFSET $3::INTEGER
```

So it does allow filtering in the query directly (see the `WHERE coalesce(...)` which comes from the hybrid_property, but it does not show up in the SELECT. I expected something like `SELECT coalesce(...) as created_by` there as well.

I though it nice if we can filter all namespaces with a specific creator or all namespaces after a date efficiently at no real cost. Though right now this isn't used, so I can also swap it out for a regular property. Was more of a learning experience :)

components/renku_data_services/namespace/orm.py
Panaetius: there's a creationTimestamp in https://www.keycloak.org/docs-api/latest/rest-api/index.html#UserRepresentation so we can definitely get it in the full user sync. There doesn't seem to be documentation of the user events payload https://www.keycloak.org/docs-api/latest/rest-api/index.html#EventRepresentation at least not that I could find, though if there isn't a creation time in there, we could use the event timestamp of user register events.

components/renku_data_services/namespace/orm.py
```diff
@@ -100,6 +119,7 @@ def dump(self) -> models.Namespace:
             slug=self.slug,
             kind=models.NamespaceKind.user,
             created_by=self.user_id,
```
leafty, line 121: nitpick: `created_by=self.created_by`?

components/renku_data_services/namespace/orm.py
```diff
@@ -74,21 +74,40 @@ class NamespaceORM(BaseORM):
...
+        raise errors.ProgrammingError(
+            message=f"Found a namespace {self.slug} that has no group or user associated with it."
+        )
+
+    @property
+    def creation_date(self) -> datetime | None:
+        """When this namespace was created."""
+        return self.group.creation_date if self.group else None
+
     def dump(self) -> models.Namespace:
```
leafty, line 93: It seems that the `dump()` method could avoid branching by using some of the new `properties`. Maybe a new one needs to be added for `self.underlying_resource_id`.

leafty (dismissed):

leafty (dismissed):

PR #498, opened 2024-11-01
chore: set in app include so sentry detects our application code

Panaetius:
Sentry currently detects stack traces for renku code as "system" instead of in-app. This also means it doesn't link to github source code and treats errors as 3rd party library errors.

/deploy #notest extra-values=dataService.sentry.enabled=true,dataService.sentry.dsn=https://e47c95d665684bce89984bd4595268df@sentry.dev.renku.ch/13,dataService.sentry.environment=ci-renku-3548,dataService.sentry.sampleRate=0.5

### DIFF
--- a/bases/renku_data_services/data_api/main.py
+++ b/bases/renku_data_services/data_api/main.py
@@ -93,6 +93,7 @@ def create_app() -> Sanic:
                 enable_tracing=config.sentry.sample_rate > 0,
                 traces_sample_rate=config.sentry.sample_rate,
                 before_send=filter_error,
+                in_app_include=["renku_data_services"],
             )
 
         # we manually need to set the signals because sentry sanic integration doesn't work with using


leafty (commented):
Thanks for looking at it @Panaetius. Is there an example trace from Sentry you can paste or link to here?

Panaetius:
yes. e.g. https://sentry.dev.renku.ch/organizations/sentry/issues/8548/?project=13&query=is%3Aunresolved&referrer=issue-stream is how things are currently and you can see that our code has the "system" tag (top right):
image

and with this PR https://sentry.dev.renku.ch/organizations/sentry/issues/8611/?project=13&query=is%3Aunresolved&referrer=issue-stream
image
now says "in app" in the top right and has an "open line in github" link.

Note that the latter only works for code in `components/` as sentry want's to map errors to code 1:1 and I've mapped `import renku_data_services.*` to `components/renku_data_services/*`, which makes the most sense to me as that's the likeliest place for errors to occur and we can't have more finegrained mappings (or well, we'd have to manually map each module which doesn't seem worth the effort)

leafty (approved):
LGTM

PR #500, opened 2024-11-01
refactor: use typed patch instance for session updates

leafty:
See #485.

The session blueprint is also updated to use the `validated_json()` method for API responses.

### DIFF
--- a/components/renku_data_services/session/apispec_base.py
+++ b/components/renku_data_services/session/apispec_base.py
@@ -1,5 +1,7 @@
 """Base models for API specifications."""
 
+from typing import Any
+
 from pydantic import BaseModel, field_validator
 from ulid import ULID
 
@@ -12,8 +14,10 @@ class BaseAPISpec(BaseModel):
 
         from_attributes = True
 
-    @field_validator("id", mode="before", check_fields=False)
+    @field_validator("*", mode="before", check_fields=False)
     @classmethod
-    def serialize_id(cls, id: str | ULID) -> str:
-        """Custom serializer that can handle ULIDs."""
-        return str(id)
+    def serialize_ulid(cls, value: Any) -> Any:
+        """Handle ULIDs."""
+        if isinstance(value, ULID):
+            return str(value)
+        return value
--- a/components/renku_data_services/session/blueprints.py
+++ b/components/renku_data_services/session/blueprints.py
@@ -7,10 +7,11 @@ from sanic.response import JSONResponse
 from sanic_ext import validate
 from ulid import ULID
 
-import renku_data_services.base_models as base_models
+from renku_data_services import base_models
 from renku_data_services.base_api.auth import authenticate, validate_path_project_id
 from renku_data_services.base_api.blueprint import BlueprintFactoryResponse, CustomBlueprint
 from renku_data_services.session import apispec
+from renku_data_services.session.core import validate_environment_patch, validate_unsaved_environment
 from renku_data_services.session.db import SessionRepository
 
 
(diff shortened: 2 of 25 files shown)


components/renku_data_services/session/blueprints.py
```diff
@@ -26,9 +34,7 @@ def get_all(self) -> BlueprintFactoryResponse:
 
         async def _get_all(_: Request) -> JSONResponse:
             environments = await self.session_repo.get_environments()
-            return json(
-                [apispec.Environment.model_validate(e).model_dump(exclude_none=True, mode="json") for e in environments]
-            )
+            return validated_json(apispec.EnvironmentList, [self._dump_environment(e) for e in environments])
```
Panaetius, line 37: question: does `validate_and_dump` not work for a list of models? we use it that way in other places in the code and it seems to work fine. or why is `_dump_environment` needed? e.g. `GroupsBP.get_all`

components/renku_data_services/session/blueprints.py
```diff
@@ -92,12 +111,7 @@ def get_all(self) -> BlueprintFactoryResponse:
         @authenticate(self.authenticator)
         async def _get_all(_: Request, user: base_models.APIUser) -> JSONResponse:
             launchers = await self.session_repo.get_launchers(user=user)
-            return json(
-                [
-                    apispec.SessionLauncher.model_validate(item).model_dump(exclude_none=True, mode="json")
-                    for item in launchers
-                ]
-            )
+            return validated_json(apispec.SessionLaunchersList, [self._dump_launcher(item) for item in launchers])
```
Panaetius, line 114: question: same here, why is a dump method needed?

components/renku_data_services/session/blueprints.py
Panaetius: I've solved that with a custom serializers in the past: https://github.com/SwissDataScienceCenter/renku-data-services/blob/main/components/renku_data_services/namespace/apispec_base.py#L20 though if it can be `None` is also need a check for that to not end up with `'None'` as an id

leafty: Oh this would explain why it works on the `id` field but not with other fields. I don't know if we can really use this approach since it does not look like it can be easily be applied only to the classes which need it.

Panaetius: yes it applies to everything and not per class, unfortunately. You could use `isinstance()` to only have it apply to ULID and pass everything else on verbatim

leafty: See c2525fdca46b856a1f1b8336e1f95bbe3b5eb584.

components/renku_data_services/session/blueprints.py
leafty: Done in c2525fdca46b856a1f1b8336e1f95bbe3b5eb584.

components/renku_data_services/session/apispec_base.py
```diff
@@ -12,8 +14,10 @@ class Config:
...
 
-    @field_validator("id", mode="before", check_fields=False)
+    @field_validator("*", mode="before", check_fields=False)
     @classmethod
-    def serialize_id(cls, id: str | ULID) -> str:
-        """Custom serializer that can handle ULIDs."""
-        return str(id)
+    def serialize_ulid(cls, value: Any) -> Any:
+        """Handle ULIDs."""
+        if isinstance(value, ULID):
```
Panaetius, line 21: that's a nice way to handle it :+1: , we might want to add that for the other apispec_base.py's as well

Panaetius (approved):

components/renku_data_services/session/apispec_base.py
leafty: I was thinking that we should have a common `apispec_base.py` file because otherwise we have a lot of code fragmentation.

PR #507, opened 2024-11-05
feat: copy projects

m-alisafaee:
Adds an `projects/{project_id}/copies` endpoint to create copies of a project using `POST` and getting all copies of a project using `GET`.

If data connectors cannot be copied because of lack of permission, the copy of the project still succeeds but a `403` status code is returned.


components/renku_data_services/project/blueprints.py
```diff
@@ -61,21 +65,74 @@ def post(self) -> BlueprintFactoryResponse:
...
+                visibility=body.visibility.value if body.visibility is not None else None,
+                keywords=[kw.root for kw in body.keywords] if body.keywords is not None else [],
+                project_repo=self.project_repo,
+                session_repo=self.session_repo,
+                data_connector_to_project_link_repo=self.data_connector_to_project_link_repo,
+            )
+
+            return validated_json(apispec.Project, self._dump_project(project), status=201)
+
+        return "/projects/<project_id>/copies", ["POST"], _copy
```
Panaetius, line 115:
```suggestion
            _: Request, user: base_models.APIUser, project_id: ULID, body: apispec.ProjectPost
        ) -> JSONResponse:
            project = await copy_project(
                project_id=project_id,
                user=user,
                name=body.name,
                namespace=body.namespace,
                slug=body.slug,
                description=body.description,
                repositories=body.repositories,
                visibility=body.visibility.value if body.visibility is not None else None,
                keywords=[kw.root for kw in body.keywords] if body.keywords is not None else [],
                project_repo=self.project_repo,
                session_repo=self.session_repo,
                data_connector_to_project_link_repo=self.data_connector_to_project_link_repo,
            )

            return validated_json(apispec.Project, self._dump_project(project), status=201)

        return "/projects/<project_id:ulid>/copies", ["POST"], _copy
```

components/renku_data_services/project/blueprints.py
```diff
@@ -61,21 +65,74 @@ def post(self) -> BlueprintFactoryResponse:
...
+            except ValueError:
+                raise errors.ValidationError(message=f"Template project ID isn't valid: {project_id}")
+
+            projects = await self.project_repo.get_all_copied_projects(user=user, project_id=template_project_id)
+
+            projects_dump = [self._dump_project(p) for p in projects]
+
+            return validated_json(apispec.ProjectsList, projects_dump)
+
+        return "/projects/<project_id>/copies", ["GET"], _get_all_copies
```
Panaetius, line 134:
```suggestion
        async def _get_all_copies(_: Request, user: base_models.APIUser, project_id: ULID) -> JSONResponse:

            projects = await self.project_repo.get_all_copied_projects(user=user, project_id=project_id)

            projects_dump = [self._dump_project(p) for p in projects]

            return validated_json(apispec.ProjectsList, projects_dump)

        return "/projects/<project_id:ulid>/copies", ["GET"], _get_all_copies
```

components/renku_data_services/project/blueprints.py
```diff
@@ -231,4 +288,5 @@ def _dump_project(project: project_models.Project) -> dict[str, Any]:
             description=project.description,
             etag=project.etag,
             keywords=project.keywords or [],
+            template_id=str(project.template_id) if project.template_id else None,
```
Panaetius, line 291: Issue: I don't think `str()` is needed here, same a few lines higher. The code in project/apispec_base.py takes care of that. https://github.com/SwissDataScienceCenter/renku-data-services/blob/main/components/renku_data_services/session/apispec_base.py#L20 has a more generic version of that that does it for all fields, but hasn't been ported to other parts of the codebase.

components/renku_data_services/project/core.py
```diff
@@ -15,3 +22,54 @@ def validate_project_patch(patch: apispec.ProjectPatch) -> models.ProjectPatch:
...
+
+async def copy_project(
+    project_id: ULID,
+    user: APIUser,
+    name: str,
+    namespace: str,
+    slug: str | None,
+    description: str | None,
+    repositories: list[models.Repository] | None,
+    visibility: str | None,
```
Panaetius, line 35: nitpick: you should already pass in the `Visibility | None` object here, not str. Blueprint should be dealing with parsing input.

Panaetius (changes requested):
Functionality looks good, mostly style/convention issues.

Panaetius (approved):

PR #511, opened 2024-11-06
fix: environment kind assignment

olevski:

PR #513, opened 2024-11-07
fix: migrate old session launchers

olevski:


olevski:

- #418
    - #511
        - #512
            - #513 ◀

.devcontainer/generate_ulid_func.sql
```diff
@@ -0,0 +1,86 @@
+ALTER DATABASE renku WITH is_template TRUE;
```
Panaetius, line 1: question: I've used this before for testing and it's very useful. it's also useful in migrations, but since  I only needed a one-off back then, I didn't add it. But I wonder if we should just add this through a migration to our database? It's also useful when e.g. debugging in a PR deployment and testing migrations there.

components/renku_data_services/migrations/versions/584598f3b769_expand_and_separate_environments_from_.py
```diff
@@ -26,22 +26,48 @@
...
     )
+    op.execute("UPDATE sessions.environments SET environment_kind = 'GLOBAL' WHERE environment_kind is NULL")
+    # NOTE: When the session launcher has environment_id set to null then it is a custom environment
+    # NOTE: We populate the name column in the new environment with the session launcher id
+    # This way we can join the newly add environments with their launchers and update the foreign key in the launchers
+    # with the newly created environments.
+    # NOTE: Since postgres cannot autogenerate ulids we use the session launcher ulid when creating the environment
+    op.execute(
+        "INSERT INTO sessions.environments(id, name, created_by_id, creation_date, container_image, default_url, environment_kind) "
+        "SELECT generate_ulid(), id, created_by_id, creation_date, container_image, default_url, 'CUSTOM' "
```
Panaetius, line 45: issue: actually this would fail in prod, no? because the generate_ulid function is only available in the devcontainer, so locally and in CI?

Panaetius (changes requested):

components/renku_data_services/migrations/versions/584598f3b769_expand_and_separate_environments_from_.py
olevski: I added it to the db init in the renku repo PR: https://github.com/SwissDataScienceCenter/renku/pull/3712

Panaetius (approved):

Panaetius (approved):

PR #519, opened 2024-11-08
fix: add user secret key when using saved secrets

olevski:
/deploy renku=feat-jupyter-free-sessions amalthea-sessions=main renku-ui=andrea/jupyter-free-build renku-gateway=master extra-values=amalthea-sessions.deployCrd=false

### DIFF
--- a/components/renku_data_services/notebooks/api/schemas/cloud_storage.py
+++ b/components/renku_data_services/notebooks/api/schemas/cloud_storage.py
@@ -151,8 +151,16 @@ class RCloneStorage(ICloudStorageRequest):
         namespace: str,
         labels: dict[str, str] | None = None,
         annotations: dict[str, str] | None = None,
+        user_secret_key: str | None = None,
     ) -> client.V1Secret:
         """The secret containing the configuration for the rclone csi driver."""
+        string_data = {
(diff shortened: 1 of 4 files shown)


olevski:

- #519 ◀

leafty:
Note: for user secrets, we inject the data service API URL along with the user's Renku token to decrypt the secrets and not the cipher key directly. Should this approach be used for data connector secrets as well?

m-alisafaee (approved):
Thanks Tasko!

olevski:
> Note: for user secrets, we inject the data service API URL along with the user's Renku token to decrypt the secrets and not the cipher key directly. Should this approach be used for data connector secrets as well?

> The init user secrets container then grabs the key from the internal API endpoint.

@leafty for the data connectors I followed the same thing we do right now in the renku repo. There we add the the decryption key in the secret that contains the configuration. And the csi rclone finishes the decryption of the saved secret.

We cannot use an init container because the pod cannot start at all (including the init containers) until the volumes are mounted. And the volume (i.e. data connector) cannot be mounted until rclone has the fully decrypted secret in order to do the mounting. So we cannot use an init container to decrypt the data connector secrets.

PR #526, opened 2024-11-12
fix: correct implementation for multi-arch images

### LINKED ISSUE #525: Need to support new list type for docker v2 manifests
See `image.py` -> https://github.com/SwissDataScienceCenter/renku-data-services/blob/220a56016eb17e5e6a71e856f9d3c80a86a546f6/components/renku_data_services/notebooks/api/classes/image.py#L16.

A new response type needs to be handled: `application/vnd.docker.distribution.manifest.list.v2+json`. This should be similar to `application/vnd.oci.image.index.v1+json` where the response lists multiple images for different platforms.

leafty:
Fixes #525.

The `ImageRepoDockerAPI` class now properly supports list/index types which are returned for multi-arch container images.

Note that the default platform is `linux/amd64`; supporting more platforms is possible, but probably not relevant until sessions can be launched on various platforms.


test/components/renku_data_services/notebooks/test_notebooks_image_checks.py
```diff
@@ -199,7 +199,8 @@ async def test_public_image_check(image: str, exists_expected: bool) -> None:
     "image,expected_path",
     [
         ("jupyter/minimal-notebook:x86_64-python-3.11.6", PurePosixPath("/home/jovyan")),
-        ("nginx", None),
+        ("nginx", PurePosixPath("/")),
```
leafty, line 202: Note that this test was previously wrong: the image working dir for `nginx` is `/`, not `None`.

components/renku_data_services/notebooks/api/classes/image.py
```diff
@@ -80,16 +90,42 @@ async def get_image_manifest(self, image: "Image") -> Optional[dict[str, Any]]:
...
+        content_type = res.headers.get("Content-Type")
+        if content_type in [ManifestTypes.docker_v2_list.value, ManifestTypes.oci_v1_index.value]:
+            index_parsed = res.json()
+
+            def platform_matches(manifest: dict[str, Any]) -> bool:
+                platform: dict[str, Any] = manifest.get("platform", {})
+                return platform.get("architecture") == platform_architecture and platform.get("os") == platform_os
+
+            manifest = next(filter(platform_matches, index_parsed.get("manifests", [])), None)
+            image_digest: str | None = manifest.get("digest") if manifest else None
```
Panaetius, line 105:
```suggestion
            manifest = next(filter(platform_matches, index_parsed.get("manifests", [])), {})
            image_digest: str | None = manifest.get("digest")
```

Panaetius (approved):
thank you. Just have a minor code suggestion.

olevski (dismissed):

olevski (approved):

PR #527, opened 2024-11-12
feat: add polybox storage

andre-code:
PR to modify cloud storage schema to add polybox storage

/deploy


components/renku_data_services/notebooks/api/schemas/cloud_storage.py
```diff
@@ -216,6 +216,30 @@ def config_string(self, name: str) -> str:
...
+            self.configuration["provider"] = ""
+
+        if access == "shared" and storage_type == "polybox":
+            self.configuration["url"] = "https://polybox.ethz.ch/public.php/webdav/"
+        if access == "shared" and storage_type == "switchDrive":
+            self.configuration["url"] = "https://drive.switch.ch/public.php/webdav/"
+        if access == "personal" and storage_type == "polybox":
+            self.configuration["url"] = "https://polybox.ethz.ch/remote.php/webdav/"
+        if access == "personal" and storage_type == "switchDrive":
+            self.configuration["url"] = "https://drive.switch.ch/remote.php/webdav/"
```
Panaetius, line 235: nitpick: these should be `elif` since they're mutually exclusive.

components/renku_data_services/storage/rclone.py
```diff
@@ -156,6 +157,91 @@ def __patch_schema_remove_oauth_propeties(spec: list[dict[str, Any]]) -> None:
...
+        public_link_help: str,
+    ) -> None:
+        """Create a modified copy of WebDAV storage and add it to the schema."""
+        # Find WebDAV storage schema and create a modified copy
+        storage_copy = RCloneValidator.__find_webdav_storage(spec)
+        storage_copy.update({"Prefix": prefix, "Name": name, "Description": description})
+
+        custom_options = [
+            {
+                "Name": "access",
```
Panaetius, line 184: question: I'm a bit confused by this, this is a field called access, but lower down its values are used as provider. In rclone schemas, the providers should be in the field `provider` by convention.

components/renku_data_services/storage/rclone.py
```diff
@@ -176,36 +176,10 @@ def __add_webdav_based_storage(
     ) -> None:
         """Create a modified copy of WebDAV storage and add it to the schema."""
         # Find WebDAV storage schema and create a modified copy
-        storage_copy = RCloneValidator.__find_webdav_storage(spec)
+        storage_copy = RCloneValidator.__find_storage(spec, "webDav")
```
Panaetius, line 179:
```suggestion
        storage_copy = RCloneValidator.__find_storage(spec, "webdav")
```
I think this is a typo? at least it was lower case before.

components/renku_data_services/storage/rclone.py
andre-code: Oh, I see where the confusion came from. I was trying to make it flexible by adding an option for access (personal and shared), but it ended up being a bit unclear. I’ll change it so the modes are set in the provider options instead, and I’ll use the provider property to specify when an option applies to a specific mode (personal or shared).

components/renku_data_services/storage/rclone.py
andre-code: yes, it is. thanks!

components/renku_data_services/storage/rclone.py
```diff
@@ -156,6 +157,91 @@ def __patch_schema_remove_oauth_propeties(spec: list[dict[str, Any]]) -> None:
...
+                    },
+                ],
+                "Required": False,
+                "Type": "string",
+                "ShortOpt": "",
+                "Hide": 0,
+                "IsPassword": False,
+                "NoPrefix": False,
+                "Advanced": False,
+                "Exclusive": False,
```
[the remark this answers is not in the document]
andre-code: ok, thanks for clarify that

components/renku_data_services/storage/rclone.py
```diff
@@ -156,6 +157,91 @@ def __patch_schema_remove_oauth_propeties(spec: list[dict[str, Any]]) -> None:
...
+                "Value": None,
+                "Examples": [
+                    {"Value": "personal", "Help": "Use Private to connect a folder that only you use", "Provider": ""},
+                    {
+                        "Value": "shared",
+                        "Help": "To connect a folder you share with others, both personal & shared folders.",
+                        "Provider": "",
+                    },
+                ],
+                "Required": False,
```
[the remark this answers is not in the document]
andre-code: yes, it is

Panaetius (approved):

components/renku_data_services/notebooks/api/schemas/cloud_storage.py
```diff
@@ -216,6 +216,30 @@ def config_string(self, name: str) -> str:
...
+            self.configuration["url"] = "https://polybox.ethz.ch/remote.php/webdav/"
+        elif access == "personal" and storage_type == "switchDrive":
+            self.configuration["url"] = "https://drive.switch.ch/remote.php/webdav/"
+
+        # Extract the user from the public link
+        if access == "shared" and storage_type in {"polybox", "switchDrive"}:
+            public_link = self.configuration.get("public_link", "")
+            user_identifier = public_link.split("/")[-1]
+            self.configuration["user"] = user_identifier
+
```
Panaetius, line 242: just as a reminder, these changes also need to be backported to https://github.com/SwissDataScienceCenter/renku-notebooks/blob/master/renku_notebooks/api/schemas/cloud_storage.py#L155 to work with Renku v1.

PR #529, opened 2024-11-13
feat: add support for session secret slots

leafty:


components/renku_data_services/project/blueprints.py
```diff
@@ -232,3 +236,90 @@ def _dump_project(project: project_models.Project) -> dict[str, Any]:
...
+
+    project_session_secret_repo: ProjectSessionSecretRepository
+    authenticator: base_models.Authenticator
+
+    def get_session_secret_slots(self) -> BlueprintFactoryResponse:
+        """Get the session secret slots of a project."""
+
+        @authenticate(self.authenticator)
+        async def _get_session_secret_slots(_: Request, user: base_models.APIUser, project_id: ULID) -> JSONResponse:
+            secret_slots = await self.project_session_secret_repo.get_all_session_secret_slots_from_project(
```
olevski, line 253: nitpick: Feel free to ignore. A shorter name for `project_session_secret_repo` would not hurt - i.e. something like `session_secret_repo` or `secret_slots_repo`.

components/renku_data_services/project/api.spec.yaml
```diff
@@ -305,6 +305,114 @@ paths:
           $ref: "#/components/responses/Error"
       tags:
         - projects
+  /projects/{project_id}/secret_slots:
```
olevski, line 308: suggestion: We should be consistent with the names. We call these `secret_slots` here in the path but further below there is `session_secret_slots`. I think it is clear enough if we just use `secret_slots` everywhere. Or regardless of what we decide we should use the same term across the board.

test/bases/renku_data_services/data_api/test_session_secrets.py
```diff
@@ -0,0 +1,367 @@
...
+
+import pytest
+from sanic_testing.testing import SanicASGITestClient
+from ulid import ULID
+
+from renku_data_services.users.models import UserInfo
+from test.bases.renku_data_services.data_api.utils import merge_headers
+
+
+@pytest.fixture
```
olevski, line 13: Ralf just merged his PR about the tests improvements. But now all fixtures should be async. Otherwise the load order of fixtures gets weird. Because I think pytest has weird rules about how it loads async vs sync fixtures.

olevski (changes requested):
Thanks Flora, just a few minor things - all else looks good.

components/renku_data_services/project/api.spec.yaml
leafty: Sure, though I am a bit afraid about having more types of secret slots in the future.

test/bases/renku_data_services/data_api/test_session_secrets.py
leafty: Done in 06a08030c0fa9cfadfa5121259854cd3533aee96.

components/renku_data_services/project/blueprints.py
leafty: Done in 6155cf330ac226353e7f3f58523a322d9e47ac50.

components/renku_data_services/project/api.spec.yaml
leafty: The API is updated in 6155cf330ac226353e7f3f58523a322d9e47ac50: the endpoint is now `/projects/{project_id}/session_secret_slots` for consistency.

Issue #509, opened 2024-11-05
[ShapeUp] Session Secrets in Renku 2.0


leafty:
Part of https://github.com/SwissDataScienceCenter/renku/issues/3832.

Content:
1. #529
2. #531
3. #517
4. #537
5. #538
6. #548

PR #537, opened 2024-11-19
feat: support multiple names for key_mapping with kubernetes secrets

leafty:
Part of #509.

Add support for multiple names in `key_mapping` when creating a kubernetes secret. This allows the same user secret to be supplied more than once in sessions.


components/renku_data_services/secrets/core.py
```diff
@@ -44,10 +44,19 @@ async def create_k8s_secret(
     if len(missing_secret_ids) > 0:
         raise errors.MissingResourceError(message=f"Couldn't find secrets with ids {', '.join(missing_secret_ids)}")
 
-    if key_mapping:
-        if set(key_mapping) != requested_secret_ids:
+    def ensure_list(value: str | list[str]) -> list[str]:
+        return [value] if isinstance(value, str) else value
+
+    key_mapping_list = {key: ensure_list(key_mapping[key]) for key in key_mapping} if key_mapping else None
```
Panaetius, line 50: nitpick: it threw me off a bit at first glance that this is called `key_mapping_list` when it's a dict

components/renku_data_services/secrets/core.py
```diff
@@ -44,10 +44,19 @@ async def create_k8s_secret(
...
+
+    if key_mapping_list:
+        if set(key_mapping_list) != requested_secret_ids:
```
Panaetius, line 53:
```suggestion
        if key_mapping_list.keys() != requested_secret_ids:
```

components/renku_data_services/secrets/core.py
```diff
@@ -44,10 +44,19 @@ async def create_k8s_secret(
...
             raise errors.ValidationError(message="Key mapping must include all requested secret IDs")
-        if len(key_mapping) != len(set(key_mapping.values())):
+
+        all_keys: list[str] = []
+        for value in key_mapping_list.values():
+            all_keys.extend(value)
```
Panaetius, line 58: suggestion:
```suggestion
        all_keys = [key for value in key_mapping_list.values() for key in value]
```

Panaetius (changes requested):

components/renku_data_services/secrets/core.py
leafty: Do you have a suggestion? The variable is `key_mapping` except we ensure all values are `list[str]`.

Panaetius: not really, maybe something like aggregated_key_mappings

Panaetius (approved):

PR #540, opened 2024-11-21
feat: add a flag to mark projects as template

m-alisafaee:
* Adds a `template` flag to `projects` table to mark a project as a template project
* Adds a filter to get writable copies when getting list of projects that are copied from a specific project


components/renku_data_services/migrations/versions/08ac2714e8e2_add_template_flag_to_projects.py
```diff
@@ -0,0 +1,28 @@
...
+# revision identifiers, used by Alembic.
+revision = "08ac2714e8e2"
+down_revision = "ea52d750e389"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column("projects", sa.Column("template", sa.Boolean(), default=False, nullable=False), schema="projects")
```
m-alisafaee, line 21: Suggestions for a better name are welcome! We can name it `is_template` which shows the intention better.

Panaetius: `is_template` does sound better to me

components/renku_data_services/project/api.spec.yaml
```diff
@@ -193,6 +193,13 @@ paths:
           required: true
           schema:
             $ref: "#/components/schemas/Ulid"
+        - in: query
+          description: When true, only return projects that user have write access to them
```
Panaetius, line 197:
```suggestion
          description: When true, only return projects that the user has write access to
```

components/renku_data_services/project/models.py
```diff
@@ -63,6 +64,8 @@ class ProjectPatch:
     description: str | None
     keywords: list[str] | None
     documentation: str | None
+    template_id: str | None
```
Panaetius, line 67: suggestion: I wonder if we can type it as `Literal[""]|None` to make it clearer what can/can't be set.

Panaetius (changes requested):

components/renku_data_services/project/models.py
m-alisafaee: Good point! I've tried to do the same in the API spec but the generated model has an issue. I've filed a bug report for it: https://github.com/koxudaxi/datamodel-code-generator/issues/2191

Panaetius (approved):

Panaetius (approved):

PR #542, opened 2024-11-24
fix: make work and mount dir optional

olevski:
We should use the image work dir - so leaving it blank uses the docker image work dir. The mount dir can be empty because we can default to the work dir when it is not set. This allows us to not require so many inputs from users when creating environments.

/deploy renku=release-0.61.0


components/renku_data_services/notebooks/core.py
```diff
@@ -288,6 +288,22 @@ async def docker_image_exists(config: NotebooksConfig, image_url: str, internal_
...
+    """Returns the working directory for the image.
+
+    If the user is logged in the internal GitLab (Renku V1), set the
+    credentials for the check.
+    """
+
+    parsed_image = Image.from_path(image_url)
+    image_repo = parsed_image.repo_api()
+    if parsed_image.hostname == config.git.registry and internal_gitlab_user.access_token:
+        image_repo = image_repo.with_oauth2_token(internal_gitlab_user.access_token)
```
Panaetius, line 303: suggestion: there's one other place in code that does this check, but it's not really related to the workdir or image existence checks. it's more of an implementation detail. I'd move this logic somewhere more suited for it, e.g. directly in `repo_api()` or change the builder pattern to `maybe_with_oauth2` and pass it in there.

leafty (commented):
Curiosity question: I though we used `$HOME/work` so that the mount would not removed files and folders from the home directory (e.g. dotfiles). Should we default to `$HOME/work` when using the value from the container image metadata? Otherwise this may wipe some important dotfiles when starting the session.

Panaetius (changes requested):

components/renku_data_services/migrations/versions/1ef98b967767_add_command_and_args_to_environment.py
```diff
@@ -16,11 +16,12 @@
 branch_labels = None
 depends_on = None
 
+root_dir = "/home/jovyan/work"
 default_command = ["sh", "-c"]
 default_args = [
     "/entrypoint.sh jupyter server --ServerApp.ip=0.0.0.0 --ServerApp.port=8888 --ServerApp.base_url=$RENKU_BASE_URL_PATH "
     '--ServerApp.token="" --ServerApp.password="" --ServerApp.allow_remote_access=true '
-    "--ContentsManager.allow_hidden=true --ServerApp.allow_origin=*",
+    f'--ContentsManager.allow_hidden=true --ServerApp.allow_origin=* --ServerApp.root_dir="{root_dir}"',
```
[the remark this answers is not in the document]
olevski: I added a new migration.

components/renku_data_services/notebooks/core.py
olevski: Ok I added this in all places. Except in one where the for the v1 sessions it was important to distinguish whether the image is publicaly or pirvately available.

olevski:
@leafty I made it so that if the mount dir is not specified we fallback to the image work dir followed by `/work` as you suggested.

Panaetius (approved):

Panaetius (approved):

Issue #546, opened 2024-11-27
Urgent: fix copy projects / template projects migrations


leafty:
```
Traceback (most recent call last):
  File "/app/env/lib/python3.12/site-packages/sqlalchemy/engine/base.py", line 1967, in _exec_single_context
    self.dialect.do_execute(
  File "/app/env/lib/python3.12/site-packages/sqlalchemy/engine/default.py", line 941, in do_execute
    cursor.execute(statement, parameters)
  File "/app/env/lib/python3.12/site-packages/psycopg/cursor.py", line 97, in execute
    raise ex.with_traceback(None)
psycopg.errors.NotNullViolation: column "is_template" of relation "projects" contains null values

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/app/env/bin/alembic", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/app/env/lib/python3.12/site-packages/alembic/config.py", line 636, in main
    CommandLine(prog=prog).main(argv=argv)
  File "/app/env/lib/python3.12/site-packages/alembic/config.py", line 626, in main
    self.run_cmd(cfg, options)
  File "/app/env/lib/python3.12/site-packages/alembic/config.py", line 603, in run_cmd
    fn(
  File "/app/env/lib/python3.12/site-packages/alembic/command.py", line 406, in upgrade
    script.run_env()
  File "/app/env/lib/python3.12/site-packages/alembic/script/base.py", line 586, in run_env
    util.load_python_file(self.dir, "env.py")
  File "/app/env/lib/python3.12/site-packages/alembic/util/pyfiles.py", line 95, in load_python_file
    module = load_module_py(module_id, path)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/env/lib/python3.12/site-packages/alembic/util/pyfiles.py", line 113, in load_module_py
    spec.loader.exec_module(module)  # type: ignore
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap_external>", line 995, in exec_module
  File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed
  File "/app/env/lib/python3.12/site-packages/renku_data_services/migrations/env.py", line 38, in <module>
    run_migrations(all_metadata)
[...]

PR #547, opened 2024-11-27
fix: set server default for projects.is_template

m-alisafaee:
Fixes: #546


olevski (changes requested):
I know alembic just generates the migration from the code. But we should still add the `server_default=false()`  in the ORM definition in `components/renku_data_services/project/orm.py`

components/renku_data_services/migrations/versions/9df92d455b11_make_work_and_mount_dir_nullable.py
```diff
@@ -1,7 +1,7 @@
 """make work and mount dir nullable
 
 Revision ID: 9df92d455b11
-Revises: 1ef98b967767
+Revises: ea52d750e389
```
leafty, line 4: ?

components/renku_data_services/migrations/versions/08ac2714e8e2_add_template_flag_to_projects.py
```diff
@@ -18,7 +18,9 @@
 
 def upgrade() -> None:
     # ### commands auto generated by Alembic - please adjust! ###
-    op.add_column("projects", sa.Column("is_template", sa.Boolean(), default=False, nullable=False), schema="projects")
+    op.add_column(
+        "projects", sa.Column("is_template", sa.Boolean(), server_default=sa.false(), nullable=False), schema="projects"
+    )
```
leafty, line 23: nitpick: this is what `sqlalchemy` generates with the current ORM class.

```suggestion
    op.add_column(
        "projects",
        sa.Column("is_template", sa.Boolean(), server_default=sa.text("false"), nullable=False),
        schema="projects",
    )
```

components/renku_data_services/migrations/versions/9df92d455b11_make_work_and_mount_dir_nullable.py
m-alisafaee: This wasn't matching the version in the code: https://github.com/SwissDataScienceCenter/renku-data-services/blob/c57ec3ad9869f58ddbdeab99046ec9efba567c4e/components/renku_data_services/migrations/versions/9df92d455b11_make_work_and_mount_dir_nullable.py#L14

components/renku_data_services/migrations/versions/08ac2714e8e2_add_template_flag_to_projects.py
m-alisafaee: They are equivalent. I've used your suggestion since it's consistent and is used in other places.

leafty (approved):

olevski (approved):

PR #552, opened 2024-11-29
chore: Makefile improvements

sambuc:


olevski (approved):
Thanks Lionel. That is really nice!

PR #555, opened 2024-11-29
chore: fix make schemas

sambuc:
* ignore return value in case of meaningful changes
* hide comments from the make output


olevski:
I tested this with and without a change. Also with a change in the yaml spec that does not create a change in the generated dataclasses and it works as expected!

olevski (approved):

PR #557, opened 2024-11-29
kpack & Shipwright compatible k3d cluster setup

sgaist:
## Describe your changes

This PR adds a script that allows to setup a k3d cluster in such a way that kpack and Shipwright can use it.

A k3d registry is created externally so it can be re-used if the cluster must be re-created without incurring the time required to populate it.

kpack and Shipwright can be optionally (and independently) deployed.

In the context of kpack, the image required for the builder is copied to the registry using a Kubernetes job.

### DIFF
--- a/.devcontainer/kpack/python-builder.yaml
+++ b/.devcontainer/kpack/python-builder.yaml
@@ -4,7 +4,7 @@ metadata:
   name: python-builder
   namespace: default
 spec:
-  tag: k3d-myregistry.localhost:12345/apps/python-builder
+  tag: kpack-registry.local:5000/apps/python-builder:latest
   stack:
     name: base
     kind: ClusterStack
--- a/Makefile
+++ b/Makefile
@@ -142,11 +142,7 @@ help:  ## Display this help.
 
 .PHONY: k3d_cluster
 k3d_cluster:  ## Creates a k3d cluster for testing
-	k3d node delete k3d-myregistry.localhost
-	k3d cluster delete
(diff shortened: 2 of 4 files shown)


olevski (commented):
Would this be a bit easier to maintain if the `setup-k3d-cluster.sh` script was instead just a python script we invoke with the dev environment?

sgaist:
@olevski sorry for the late reply, I missed the notification.

I initially wrote it as a bash script to simplify the Makefile with regard to the waiting phases and customization.

I don't mind rewriting it in python but I am not sure to follow the "we invoke with the dev environment". What do you have in mind ?

sgaist:
I was rather seeing it as a mean to setup your cluster once so you can then use it for development in the same way that is done currently.

olevski:
> I was rather seeing it as a mean to setup your cluster once so you can then use it for development in the same way that is done currently.

That is how I understood it. And I meant this as a non-blocking question/suggestion. Which I should have clarified earlier. So its not a big deal either way.

Panaetius:
since we're leaning more and more towards shipwright instead of kpack, only the registry setup in this PR seems really relevant. But setting up shipwright in a devenv would be nice.

sgaist:
@Panaetius would you like me to refactor this PR and port it to Shipwright or rather make a new one ?

Panaetius:
I'm fine with either

Panaetius (approved):

PR #562, opened 2024-12-03
fix: include direct member flag in total group query

andre-code:
PR to fix the total groups value, now including the flag for direct members when provided.

/deploy


leafty (approved):
LGTM

leafty:
@andre-code any reason this is not merged yet?

PR #563, opened 2024-12-06
fix: add Renku env variables to Renku 2.0 sessions

### LINKED ISSUE #528: Sessions: document and add more Renku environment variables.
Currently, a session will have two Renku environment variables:
* `RENKU_BASE_URL`: the base URL of the session, e.g. `https://ci-renku-3712.dev.renku.ch/sessions/flora-thieba-dcae314cd1a4`
* `RENKU_BASE_URL_PATH`: the path component of the `RENKU_BASE_URL`, e.g. `/sessions/flora-thieba-dcae314cd1a4`

Suggested new environment variables:
* `RENKU_SESSION`: always equal to `"1"` inside a Renku session. This can be used by scripts or entrypoints.
* `RENKU_SESSION_PORT`: the port which should provide the session frontend, e.g. `8888` for JupyterLab.
* `RENKU_SESSION_IP` (less important): the IP address the frontend should listen to. Right now, most session images will listen on `0.0.0.0`, but we can allow image authors to listen only on the intended IP address
* `RENKU_WORKING_DIR`: the value of the `working_directory` setting. This should allow frontends to open the intended workspace and not just `~`.

sambuc:
Closes #528.

Add the following environment variables:

* RENKU_MOUNT_DIR    Folder which contains the container state between
                     hibernations

* RENKU_SESSION      Set to "1" inside a Renku user session

* RENKU_SESSION_IP   IP where the session web server should be listening

* RENKU_SESSION_PORT Port where the session web server should be listening

* RENKU_WORKING_DIR  Default current working directory of applications within the session container.

/deploy

### DIFF
--- a/components/renku_data_services/notebooks/blueprints.py
+++ b/components/renku_data_services/notebooks/blueprints.py
@@ -478,6 +478,11 @@ class NotebooksNewBP(CustomBlueprint):
                         env=[
                             SessionEnvItem(name="RENKU_BASE_URL_PATH", value=base_server_path),
                             SessionEnvItem(name="RENKU_BASE_URL", value=base_server_url),
+                            SessionEnvItem(name="RENKU_MOUNT_DIR", value=storage_mount.as_posix()),
+                            SessionEnvItem(name="RENKU_SESSION", value="1"),
+                            SessionEnvItem(name="RENKU_SESSION_IP", value="0.0.0.0"),  # nosec B104
+                            SessionEnvItem(name="RENKU_SESSION_PORT", value=f"{environment.port}"),
+                            SessionEnvItem(name="RENKU_WORKING_DIR", value=work_dir.as_posix()),
                         ],
                     ),
                     ingress=Ingress(


leafty (commented):
Where is the documentation? If none is provided, update the PR title accordingly.

Otherwise, code looks good.

leafty (approved):
LGTM.

One small question is if having `0.0.0.0` as the IP to listen to is a good idea or not.

sambuc:
Ideally, we should set it to the IP used by the reverse proxy to contact the container, so that a service starting within the container may bind in the most specific way possible using this environment variable.

Do you know if the IP is known at this point in time? Or a DNS/IP alias which may be used for the same purpose?

leafty:
> Ideally, we should set it to the IP used by the reverse proxy to contact the container, so that a service starting within the container may bind in the most specific way possible using this environment variable.
> 
> Do you know if the IP is known at this point in time? Or a DNS/IP alias which may be used for the same purpose?

Conceptually, this is relatively easy:
https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/#use-pod-fields-as-values-for-environment-variables

But I do not know if this is something that can be requested from the amaltheasession CRD.

sambuc:
Here some preliminary content for the documentation:

# Session Environment

The following variables are set on startup of the session in the environment. You may use these to configure your service dynamically at startup time.

## Variables 

- `RENKU_DISABLE_VERSION_CHECK` Set to `1` 
- `RENKU_BASE_URL` Full root URL to your session
- `RENKU_BASE_URL_PATH` Part of the URL which is the root path to this session
- `RENKU_MOUNT_DIR` Folder which contains the container state between hibernations
- `RENKU_SESSION` Set to "1" inside a `RENKU user session
- `RENKU_SESSION_IP` IP where the session web server should be listening
- `RENKU_SESSION_PORT` Port where the session web server should be listening
- `RENKU_WORKING_DIR` Default current working directory of applications within the session container.

PR #567, opened 2024-12-10
feat: save storage secrets 

olevski:
Migrating
https://github.com/SwissDataScienceCenter/renku-notebooks/pull/1927

Depends on #566 and #565 

Also will keep this as draft and resolve/merge only when the new data connectors code and secrets that @leafty is working on is merged.

/deploy renku-ui=main renku-gateway=feat-shutdown-notebooks renku-notebooks=feat-use-different-user-labels-in-watcher

### DIFF
--- a/components/renku_data_services/notebooks/api/schemas/cloud_storage.py
+++ b/components/renku_data_services/notebooks/api/schemas/cloud_storage.py
@@ -8,9 +8,8 @@ from typing import Any, Final, Optional, Protocol, Self
 from kubernetes import client
 from marshmallow import EXCLUDE, Schema, ValidationError, fields, validates_schema
 
-from renku_data_services.base_models import APIUser
 from renku_data_services.notebooks.api.classes.cloud_storage import ICloudStorageRequest
-from renku_data_services.notebooks.config import NotebooksConfig
+from renku_data_services.storage.models import CloudStorage
 
 _sanitize_for_serialization = client.ApiClient().sanitize_for_serialization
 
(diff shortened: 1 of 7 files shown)


Panaetius:
would this be a good time to switch to direct calls in code instead of calling the data svc REST api from within data svc?

olevski:
> would this be a good time to switch to direct calls in code instead of calling the data svc REST api from within data svc?

I would rather do this in a separate pr. I really dont want to bloat these PRs with anything else other than moving the code over form the notebook service. Then if you want to dig out what or why we did something it will be easier.

olevski:
Here is a follow up PR for this: https://github.com/SwissDataScienceCenter/renku-data-services/issues/586

olevski:
This is ready for review. The stuff in the pr description is needed only to be able to test and direct requests to the data service instead of the notebook service.

Also I removed http calls to the data service for cloud storage. Which were there prior to this from the code that got moved over from notebooks.

components/renku_data_services/notebooks/api/schemas/cloud_storage.py
```diff
@@ -57,54 +56,55 @@ def __init__(
...
         else:
             source_path = data["source_path"]
             target_path = data["target_path"]
             configuration = data["configuration"]
             readonly = data.get("readonly", True)
-        mount_folder = str(work_dir / target_path)
 
-        await config.storage_validator.validate_storage_configuration(configuration, source_path)
-        return cls(source_path, configuration, readonly, mount_folder, name, config)
+        # NOTE: This is used only in Renku v1, there we cannot save secrets for storage
```
leafty, line 95:
```suggestion
        # NOTE: This is used only in Renku v1, there we do not save secrets for storage
```

We could save them but we don't.

leafty (approved):
LGTM, seems to be working for renku 1.0 sessions.

PR #569, opened 2024-12-10
chore: remove Renku2UserServer code

olevski:
This code is not used. For Renku v2 we are launching sessions with the new operator.

/deploy


Panaetius (approved):

Panaetius (approved):

olevski:
:facepalm: There seems to be some kind of request throttling that happens somewhere (I think in k3d) because locally the tests pass but here I see 429 too many requests. I will do a bit more digging.

olevski:
> 🤦 There seems to be some kind of request throttling that happens somewhere (I think in k3d) because locally the tests pass but here I see 429 too many requests. I will do a bit more digging.

I cannot really replicate this.

Also the current tests are failing because the selenium acceptance tests need to access renkulab.io and they fail because prod is down. So I will merge this.

Panaetius (approved):

PR #574, opened 2024-12-11
fix: properly validate docker image names

### LINKED ISSUE #551: ContainerImage isn't validated in environments endpoints
We should validate that the image name for an environment is actually a valid image identifier. Currently it accepts any character, including spaces. This then causes session launch to fail because it can't find the image, but is very difficult to spot that this happened in the input field. (so ` renku/myimage` doesn't work where `renku/myimage` works).

ideally we'd also validate this in the UI.
For reference, one of these two won't launch(due to a space in the beginning):
image
image

Panaetius:
closes #551 (at least on the backend, still needs to be done on the frontend)


components/renku_data_services/session/api.spec.yaml
```diff
@@ -535,6 +535,8 @@ components:
       description: A container image
       type: string
       maxLength: 500
+      # NOTE: regex for an image name, optionally with a tag or sha256 specified
```
olevski, line 538: Can we please reference where we got this from in a comment?
is it from https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pulling-manifests?

test/bases/renku_data_services/data_api/test_sessions.py
```diff
@@ -80,11 +80,20 @@ async def test_get_session_environment(
...
 
 @pytest.mark.asyncio
-async def test_post_session_environment(sanic_client: SanicASGITestClient, admin_headers) -> None:
+@pytest.mark.parametrize(
+    "image_name",
+    [
+        "renku/renku",
+        "u/renku/renku:latest",
+        "docker.io/renku/renku:latest",
+        "renku/renku@sha256:eceed25752d7544db159e4144a41ed6e96e667f39ff9fa18322d79c33729a18c",
```
olevski, line 89: I know it is probably fine. But lets add one or more gitlab repo image names here. Just to be safe. Because a large chunk of our images are from gitlab image repos.

olevski (changes requested):
Thanks Ralf two minor changes.

PR #581, opened 2024-12-19
fix: allow mount and work directories to be reset for session environments

### LINKED ISSUE #580: Removing/clearing certain nullable fields on a session launcher fails
Seen on renku 0.61.2

To replicate:
- create a session launcher
- set the mount directory to something
- try to remove it and have it be blank (this will not work)

The request succeeds but the value is left the same (i.e. the mount directory is not cleared). This is because the backend probably drops all null fields in the validation and therefore removes the mount dir field.

This is probably the same for the workdir and other session launcher fields that can be null.

leafty:
Closes #580.

Note: tested that the UI works as intended: clearing out the working or mount directory will reset the value to `None`.

/deploy #notest


components/renku_data_services/session/models.py
```diff
@@ -99,8 +99,8 @@ class EnvironmentPatch:
     container_image: str | None = None
     default_url: str | None = None
     port: int | None = None
-    working_directory: PurePosixPath | None = None
-    mount_directory: PurePosixPath | None = None
+    working_directory: PurePosixPath | ResetType | None
+    mount_directory: PurePosixPath | ResetType | None
```
olevski, line 103: suggestion: For consistency's sake it is better to keep the default `None` value on these just like all the other  fields.

Also we can remove `EnvironmentUpdate` above this. I know you did not touch. But I noticed it is not used anywhere at all. It is a leftover from some other PR where we should have got rid of it.

components/renku_data_services/session/core.py
```diff
@@ -31,14 +31,28 @@ def validate_unsaved_environment(
...
+        if patch.working_directory == ""
+        else None
+    )
+    mount_directory = (
+        PurePosixPath(patch.mount_directory)
+        if patch.mount_directory
+        else RESET
+        if patch.mount_directory == ""
+        else None
+    )
```
olevski, line 47: This and the above statement are a bit hard to parse. A match statement like this is roughtly the same number of lines but easier to understand:

```
    match patch.mount_directory:
        case "":
            mount_directory = RESET
        case str():
            mount_directory = PurePosixPath(patch.mount_directory)
        case _:
            mount_directory = None

```

olevski (changes requested):
Thanks @leafty. 2 minor things. All looks good though.

components/renku_data_services/session/models.py
leafty: Done in 8914a70067a9b6c7332d276020c64226e17c0b48.

components/renku_data_services/session/core.py
leafty: Done in 8914a70067a9b6c7332d276020c64226e17c0b48.

olevski (approved):

PR #585, opened 2024-12-19
fix: patching wrong environment variables when resuming

olevski:
Migrating
https://github.com/SwissDataScienceCenter/renku-notebooks/pull/1923

/deploy


leafty (approved):
LGTM

PR #590, opened 2024-12-20
feat: add disk storage to session launchers

### LINKED ISSUE #589: Persist the requested disk size for session launchers
See https://github.com/SwissDataScienceCenter/renku-ui/issues/3451.

Add a `disk_size` field alongside the existing `resource_class_id` field in session launchers, so that users can request more/less disk space when they create or edit session launchers.

leafty:
Closes #589.

Add a new `disk_storage` field to session launchers so that a custom disk size can be persisted.

Note: the disk storage size is not validated when saved.

See also: https://github.com/SwissDataScienceCenter/renku-ui/pull/3463
Deployment: https://renku-ci-ui-3463.dev.renku.ch/


components/renku_data_services/session/core.py
```diff
@@ -118,9 +119,11 @@ def validate_session_launcher_patch(
         resource_class_id = RESET
     else:
         resource_class_id = patch.resource_class_id
+    disk_storage = RESET if "disk_storage" in data_dict and data_dict["disk_storage"] is None else patch.disk_storage
```
m-alisafaee, line 122: issue: What if we don't want to change the disk storage in the PATCH? If we don't set it, the value will be `None` and it causes the disk storage to be reset (in the followup code).

test/bases/renku_data_services/data_api/test_sessions.py
```diff
@@ -451,6 +455,7 @@ async def test_patch_session_launcher(
     assert res.json.get("name") == patch_payload["name"]
     assert res.json.get("description") == patch_payload["description"]
     assert res.json.get("resource_class_id") == patch_payload["resource_class_id"]
+    assert res.json.get("disk_storage") == 3
```
m-alisafaee, line 458: suggestion: We should add a test to check disk storage isn't modified if it isn't sent in the PATCH.

m-alisafaee (commented):
Thanks Flora! Please see my comments.

components/renku_data_services/session/api.spec.yaml
```diff
@@ -419,32 +419,15 @@ components:
...
-          creation_date: "2023-11-01T17:32:28Z"
-          description: JupyterLab session environment
-          environment_kind: GLOBAL
-          container_image: rocker/rstudio
-          default_url: "/rstudio"
-          port: 8080
-          working_directory: /home/rstudio/work
-          mount_directory: /home/rstudio/work
-          uid: 1000
-          gid: 1000
```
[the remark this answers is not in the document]
leafty: Auto-generated examples are a better reflection of the actual schema. If we provide one, we need to actually update it with each schema update. The removed example was not valid anymore, even before this PR.

components/renku_data_services/session/core.py
leafty: A value of `None` does NOT reset the value during the patch operation.

test/bases/renku_data_services/data_api/test_sessions.py
m-alisafaee: I've added a commit with the test.

m-alisafaee (approved):

test/bases/renku_data_services/data_api/test_sessions.py
leafty: :pray: Thank you!

Issue #595, opened 2025-01-06
Some ETag values are not correctly computed


leafty:
Examples:
1. `project` ETag does not vary if the namespace slug has been changed.

PR #596, opened 2025-01-06
fix: update project ETag when the namespace slug is changed

leafty:
See #595.

Makes the `etag` value change if the namespace slug changes. This fixes a bug where clients may keep an outdated cached version of a `project` object when the parent namespace gets updated (e.g. the group slug is modified).

/deploy


components/renku_data_services/utils/etag.py
```diff
@@ -2,11 +2,28 @@
...
+    values.extend(arg for arg in args)
+    to_hash = "-".join(_get_hashable_string(value) for value in values)
+    etag = md5(to_hash.encode(), usedforsecurity=False).hexdigest().upper()
+    return f'"{etag}"'
+
+
+def _get_hashable_string(value: Any) -> str:
+    if isinstance(value, datetime):
+        return value.isoformat()
+    return f"{value}"
```
Panaetius, line 29:
```suggestion
    return str(value)
```

Panaetius (approved):
Looks good :+1: Just a minor suggestion, feel free to ignore

components/renku_data_services/utils/etag.py
leafty: Oh, this is better, thanks.

Panaetius (approved):

PR #597, opened 2025-01-08
fix: do not call dataservice through network from itself

### LINKED ISSUE #586: Remove network calls to "data_service" from code migrated from notebooks
These should all be direct calls to postgres through the db adaptors we have in the codebase. Not calls via requests or httpx.

Panaetius:
closes #586 

I remember there were also calls to storage service but those seem to have been migrated already

/deploy

### DIFF
--- a/components/renku_data_services/notebooks/utils.py
+++ b/components/renku_data_services/notebooks/utils.py
@@ -1,9 +1,6 @@
 """Utilities for notebooks."""
 
-import httpx
-
 import renku_data_services.crc.models as crc_models
-from renku_data_services.base_models.core import AuthenticatedAPIUser
 from renku_data_services.notebooks.crs import (
     MatchExpression,
     NodeAffinity,
@@ -13,7 +10,6 @@ from renku_data_services.notebooks.crs import (
     RequiredDuringSchedulingIgnoredDuringExecution,
     Toleration,
 )
-from renku_data_services.utils.cryptography import get_encryption_key
 
 
 def merge_node_affinities(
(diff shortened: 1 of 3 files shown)


components/renku_data_services/notebooks/utils.py
```diff
@@ -99,18 +95,3 @@ def tolerations_from_resource_class(resource_class: crc_models.ResourceClass) ->
     for tol in resource_class.tolerations:
         output.append(Toleration(key=tol, operator="Exists"))
     return output
-
-
-async def get_user_secret(data_svc_url: str, user: AuthenticatedAPIUser) -> str | None:
-    """Get the user secret key from the secret service."""
-
-    async with httpx.AsyncClient(timeout=5) as client:
```
eikek, line 107: For my education: is this result now coming from the `UserRepo` that has been added as a dependency into the `blueprints.py` above (now it is a library call, instead of a network one)?

eikek (approved):

components/renku_data_services/notebooks/utils.py
Panaetius: yes, now it calls the database repo directly. before it did an HTTP call to the respective endpoint

Panaetius: this code was ported from renku-notebooks, where an http request made sense. but it doesn't make sense for a service to call itself

eikek: I see, thank you! 🙏🏼

PR #598, opened 2025-01-08
fix: merge all apispecs correctly

Panaetius:
there was a bug in the merging code, the notebooks spec file was saved in the `repositories` variable overwriting the repositories apispec, so this never ended up on our swagger page.

I rewrote the whole section to make is simpler and prevent this kind of error from happening again in the future.

/deploy


eikek (approved):
Nice!

PR #601, opened 2025-01-08
Feat/shipwright strategies

sgaist:
## Describe your changes

This PR is the first iteration to implement support for using paketo buildpacks as build strategy for Shipwright.

This implementation calls to the original tools used by kpack.

There's a test validating the build and use of the strategy.

The full builder is currently used as base image so that the strategy can cover a larger set of projects.


.devcontainer/devcontainer.json
```diff
@@ -28,7 +28,10 @@
 			"command": "alias k=kubectl"
 		},
 		"ghcr.io/devcontainers-contrib/features/rclone:1": {},
-		"./k3d": {}
+		"./k3d": {},
+		"ghcr.io/devcontainers/features/go:1": {
+			"version": "1.23"
+		}
```
olevski, line 34: question: I think go is not needed?

olevski (changes requested):
A general question. I am only looking at a diff of this PR. But is the idea that the `kubectl` CLI will be called in production?

Also it is better to use a `BuildStrategy` rather than a `ClusterBuildStrategy`. Since the cluster one is cluster scoped and not namespaced. IMO it is better to have it namespaced - i.e. safer and it requries less permissions in the cluster.

.devcontainer/devcontainer.json
sgaist: Good point, I forgot to remove it !

sgaist: Done

sgaist:
> A general question. I am only looking at a diff of this PR. But is the idea that the `kubectl` CLI will be called in production?
> 

AFAIK, no. I wrote the test to confirm that things are working as intended with the current state of development.
There's an update to the API in the work that will handle the shipwright resources creation beside the strategy so this test should be replaced with the new one.

> Also it is better to use a `BuildStrategy` rather than a `ClusterBuildStrategy`. Since the cluster one is cluster scoped and not namespaced. IMO it is better to have it namespaced - i.e. safer and it requires less permissions in the cluster.

Agreed, I'll push a new version using a `BuildStrategy`.

olevski (approved):

PR #604, opened 2025-01-13
fix: add vendor for polybox storage

olevski:
This is meant to fix the issue here: https://www.notion.so/renku/Prevent-File-changed-pop-ups-for-Polybox-folders-1730df2efafc80948abee9f248cbef9b?pvs=4

It is a popup that shows up every time you mount a polybox storage and you edit a file:
image (19)

There may be a few other options we can cleanup in the csi-rclone code. But this one is the main one that is causing the issue.

/deploy


olevski:
This is currently blocked by the UI and how it interprets fields in the rclone config and which of these fields are sent to the backend when a data connector is created.

Besically it seems that there is no way to have a field in the rclone config that has a default and is also sent to the backend when a cloud storage is created. If we want the ui to send the field to the backend then the field should be set to required without a default value.

@lorenzo-cavazzi correctly if I am wrong here. But this is my understanding from our quick discussion.

leafty:
@olevski I think you may want to deploy against `renku=release-0.64.0` so that you include the latest UI fixes for data connectors.

leafty (dismissed):
This fix seems to work for polybox storage :+1: :clap:

leafty (approved):

PR #606, opened 2025-01-14
fix: renable s3 customizations for data connectors

Panaetius:
this add back s3 and other customisations for the storage schema

/deploy

### DIFF
--- a/test/bases/renku_data_services/data_api/test_storage.py
+++ b/test/bases/renku_data_services/data_api/test_storage.py
@@ -600,3 +600,4 @@ async def test_storage_schema(storage_test_client) -> None:
     providers = next(p for p in s3["options"] if p["name"] == "provider")
     assert providers
     assert providers.get("examples")
+    assert any(e["value"] == "Switch" for e in providers.get("examples"))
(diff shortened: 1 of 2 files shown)


test/bases/renku_data_services/data_api/test_storage.py
```diff
@@ -600,3 +600,4 @@ async def test_storage_schema(storage_test_client) -> None:
     providers = next(p for p in s3["options"] if p["name"] == "provider")
     assert providers
     assert providers.get("examples")
+    assert any(e["value"] == "Switch" for e in providers.get("examples"))
```
leafty, line 603: It would probably be better to have one test per patch in the future.

leafty (approved):
Note, can be solved later: maybe the webdav patches should work like the other patches instead of being invoked as special cases.

Fix looks OK to merge :+1:

Panaetius:
I created https://github.com/SwissDataScienceCenter/renku-data-services/issues/607 and https://github.com/SwissDataScienceCenter/renku-data-services/issues/608 for follow-up issues

PR #611, opened 2025-01-17
feat: Adding new relation for listing private-only entities

eikek:
/deploy


olevski (approved):

leafty:
Non-blocking: should this be merged into a feature branch instead? As is, the authorization changes are not *used* anywhere in code. If these changes are part of a larger set of changes, it is easier for everyone contributing to the repository if a feature branch is used.

olevski:
Yeah lets merge this into a feature branch. That way if we have to modify the schema before an official release we wont have to do migrations.

PR #614, opened 2025-01-21
Feat/image builder api

SalimKayal:
Implementation for the new custom environments with build parameters.

# TODO:
- [ ] Connect to ShipWright
- [ ] Clean up validation code to be in the same place


components/renku_data_services/session/api.spec.yaml
```diff
@@ -261,6 +353,8 @@ components:
           $ref: "#/components/schemas/Description"
         container_image:
           $ref: "#/components/schemas/ContainerImage"
+        builder_id:
+          $ref: "#/components/schemas/ULID"
```
Panaetius, line 357: do we want a separate endpoint for creating builders and then have image and builder_id as properties here, or should they just be POST on this endpoint with a `oneOf:` body?

components/renku_data_services/session/api.spec.yaml
```diff
@@ -531,11 +641,105 @@ components:
...
       maxLength: 500
+    ImageBuilder:
+      description: A container image builder
+      type: object
+      additionalProperties: false
+      properties:
+        builder_id:
+          $ref: "#/components/schemas/Ulid"
+        container_image:
+          $ref: "#/components/schemas/ContainerImage"
```
Panaetius, line 652: what is the container image for? If it's the run image it should be called that.

It's not clear to me if we let users customize this or if it's something the backend sets as it decides what frontend there is (at least for now, until we have custom buildpacks for the frontends)

components/renku_data_services/session/api.spec.yaml
```diff
@@ -531,11 +641,105 @@ components:
...
+        repository:
+          $ref: "#/components/schemas/Repository"
+        branch:
+          $ref: "#/components/schemas/branch"
```
Panaetius, line 656: This can probably be any github ref, not just branches? In shipwright builds this is called `revision`

components/renku_data_services/session/api.spec.yaml
```diff
@@ -531,11 +641,105 @@ components:
...
+        subdir:
+          $ref: "#/components/schemas/ImageBuilderSubDir"
+        venv_kind:
+          $ref: "#/components/schemas/VEnvKind"
+        editor_kind:
+          $ref: "#/components/schemas/EditorKind"
+      required:
+        - builder_id
+        - container_image
```
Panaetius, line 665: Is this required or do we define a default?

components/renku_data_services/session/api.spec.yaml
```diff
@@ -531,11 +641,105 @@ components:
...
+        repository: https://gitlab.com/my_namespace/my_project_repo.git
+        branch: main
+        subdir: "."
+        venv_kind: CONDA
+        editor_kind: JUPYTERLAB
+    ImageBuilderPost:
+      description: Data required to create an image builder
+      type: object
+      properties:
+        builder_id:
```
Panaetius, line 681: I don't think we post the ID, it's generated on insert

components/renku_data_services/session/api.spec.yaml
SalimKayal: This was one of the main questions. I am open to both. Waiting for input.

components/renku_data_services/session/api.spec.yaml
```diff
@@ -498,6 +592,21 @@ components:
...
+        - PIP
+        - R
+        - DOCKERFILE
+    EditorKind:
+      description: User's Editor Choice.
+      type: string
+      enum:
+        - VSCODIUM
+        - JUPYTERLAB
+        - STREAMLIT
```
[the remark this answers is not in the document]
SalimKayal: Noted. I will make it consistent

components/renku_data_services/session/api.spec.yaml
SalimKayal: For me it was the output image. Though there is a repetition with the one defined in the environment. If the image name is deterministic then this is superfluous.

components/renku_data_services/session/api.spec.yaml
SalimKayal: we can make that deterministic.

components/renku_data_services/session/api.spec.yaml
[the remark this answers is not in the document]
SalimKayal: Just one thing, the UPPER CASE might have been because of the `global` reserved keyword. How do you think this should be resolved ?

components/renku_data_services/session/api.spec.yaml
```diff
@@ -531,11 +641,105 @@ components:
       description: A description for the resource
       type: string
       maxLength: 500
+    ImageBuilder:
```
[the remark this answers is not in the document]
SalimKayal: instead of `ImageBuilder` would something like `ImageBuildParameters`do the trick ?

components/renku_data_services/session/api.spec.yaml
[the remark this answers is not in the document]
Panaetius: oh I didn't think of that. I wonder if there's some customization that can be done in the datamodel codegen. Also note that this is a breaking change in the API and shoudl be discussed with the UI

components/renku_data_services/session/api.spec.yaml
```diff
@@ -8,6 +8,98 @@ info:
...
+        default:
+          $ref: "#/components/responses/Error"
+      tags:
+        - image_builder
+  /image_builder/{builder_id}:
+    get:
+      summary: Get an image builder
+      parameters:
+        - in: path
+          name: builder_id
```
leafty, line 36: Convention:

```suggestion
          name: id
```

components/renku_data_services/session/api.spec.yaml
leafty, line 357:
```suggestion
          $ref: "#/components/schemas/Ulid"
```

components/renku_data_services/session/api.spec.yaml
```diff
@@ -437,7 +531,7 @@ components:
           name: Rstudio
           creation_date: "2023-11-01T17:32:28Z"
           description: JupyterLab session environment
-          environment_kind: GLOBAL
+          environment_kind: global
```
leafty, line 534: Suggestion: Instead of fixing this, it is preferable to NOT provide examples for objects. Examples are derived by swagger from single field examples and as such are easier to keep well maintained.

components/renku_data_services/session/api.spec.yaml
```diff
@@ -498,6 +592,21 @@ components:
       minLength: 1
       maxLength: 99
       example: My Renku Session :)
+    VEnvKind:
```
leafty, line 595: I think this would be better named "flavor" (or "flavor kind"). For some flavors (maybe R or node, etc.) we would not necessarily provide a "virtual environment".

components/renku_data_services/session/api.spec.yaml
```diff
@@ -514,9 +623,10 @@ components:
       description: Kind of environment to use
       type: string
       enum:
-        - GLOBAL
-        - CUSTOM
-      example: CUSTOM
+        - global
+        - custom
+        - builder
```
leafty, line 628: Changing the casing creates an API breaking change. Do we want this?

components/renku_data_services/session/api.spec.yaml
leafty, line 1: What is the update to `POST /session_launchers` and `PATCH /session_launchers/{launcher_id}`?

1. For `POST`, we should not expect the client to provide any of the environment fields, that is `name, description, container_image, default_url, uid, gid, working_directory, mount_directory, port, command, args` as these should be generated from the "builder" definition.
2. For `PATCH`, there is no defined way to update the "builder" at the moment. Also, it should be allowed to go from `CUSTOM` to `BUILDER` type of session launcher.

components/renku_data_services/session/api.spec.yaml
```diff
@@ -531,11 +641,96 @@ components:
...
       type: string
       maxLength: 500
       example: renku/renkulab-py:3.10-0.18.1
+    Repository:
+      description: The repository used as a basis for a custom builder
+      type: string
+      maxLength: 500
+      example: "https://github.com/voila-gallery/gaussian-density"
+    Revision:
+      description: The revision of the repository used as a basis for a custom builder
```
leafty, line 725:
```suggestion
      description: The revision of the repository used as a basis for a custom builder. This can be a git branch, a git tag or a git commit.
```

leafty (commented):
Some comments about the API spec:

components/renku_data_services/session/api.spec.yaml
Panaetius: See comments above, currently our API is inconsistent in terms of casing (most enums are lower case, this file is one of the few places where it's upper case). This isn't really related to this PR, my thinking was just, if we change the API here anyways, might be a good point to make it consistent

components/renku_data_services/session/api.spec.yaml
SalimKayal: I agree that "virtual environment" is not ideal. I'm not sure about "flavor" because it's imho not explicit enough. I don't have yet a better name so if I can't come with something more satisfying let's go with flavor.

components/renku_data_services/session/api.spec.yaml
SalimKayal: excellent point. Will look at that

components/renku_data_services/session/api.spec.yaml
SalimKayal: Does it ? I'm currently unsure about all the implications. Let's discuss this in a technical meeting.

components/renku_data_services/session/api.spec.yaml
m-alisafaee: I went with the term `BuilderVariant`.

leafty:
@m-alisafaee Can you fix the commit list? This PR is against `kpack-resources`, so either rebase from there or update the `kpack-resources` branch first. The changeset is not readable at the moment.

components/renku_data_services/session/api.spec.yaml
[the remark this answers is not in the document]
m-alisafaee: Renamed to `BuildParameter` for now.

components/renku_data_services/session/api.spec.yaml
m-alisafaee: I undid this change since it requires a database migration as well. I'll re-introduce it later once working on the db.

components/renku_data_services/session/api.spec.yaml
m-alisafaee: I removed these endpoints. Now, `BuildParameter` is part of session launcher API.

components/renku_data_services/session/api.spec.yaml
m-alisafaee: See new changes.

components/renku_data_services/session/api.spec.yaml
```diff
@@ -539,6 +550,48 @@ components:
       description: A description for the resource
       type: string
       maxLength: 500
+    BuildParameter:
```
leafty, line 553:
```suggestion
    BuildParameters:
```

components/renku_data_services/session/api.spec.yaml
```diff
@@ -539,6 +550,48 @@ components:
...
+          $ref: "#/components/schemas/SubDir"
+        builder_variant:
+          $ref: "#/components/schemas/BuilderVariant"
+        frontend_variant:
+          $ref: "#/components/schemas/FrontendVariant"
+      required:
+        - repository
+        - builder_variant
+        - frontend_variant
+    BuildParameterPost:
```
leafty, line 572:
```suggestion
    BuildParametersPost:
```

components/renku_data_services/session/api.spec.yaml
```diff
@@ -416,6 +405,10 @@ components:
           $ref: "#/components/schemas/EnvironmentArgs"
         is_archived:
           $ref: "#/components/schemas/IsArchived"
+        build_parameter:
+          $ref: "#/components/schemas/BuildParameter"
+        environment_image_source:
+          $ref: "#/components/schemas/EnvironmentImageSource"
```
leafty, line 411: This allows me to send:

```json
{
  "build_parameters": {
    "repository": "https://github.com/SwissDataScienceCenter/renku-data-services.git"
  },
  "environment_image_source": "image"
}
```

which does not make sense.

components/renku_data_services/session/api.spec.yaml
```diff
@@ -274,6 +274,10 @@ components:
           $ref: "#/components/schemas/Description"
         container_image:
           $ref: "#/components/schemas/ContainerImage"
+        build_parameter:
+          $ref: "#/components/schemas/BuildParameter"
+        environment_image_source:
+          $ref: "#/components/schemas/EnvironmentImageSource"
```
leafty, line 280: For GET calls, this is OK, though not great because the types do allow non-sensical values to be sent, e.g.

```json
{
  "build_parameters": { [object] },
  "environment_image_source": "image"
}
```

or

```json
{
  <no build_parameters>,
  "environment_image_source": "build"
}
```

leafty (commented):
The type union on `EnvironmentPostInLauncher` is nice :+1:

I think the same should be done to `Environment` and `EnvironmentPatchInLauncher` because non-sensical values would be allowed.

m-alisafaee:
> EnvironmentPatchInLauncher

I did the same for `EnvironmentPatchInLauncher` but not for `Environment`, since we always return `global` envs from that endpoint and they don't support `build` option yet. Instead, I've applied it to `EnvironmentGetInLauncher`.

components/renku_data_services/session/api.spec.yaml
```diff
@@ -377,16 +385,23 @@ components:
...
-        - $ref: "#/components/schemas/EnvironmentPatch"
+        - oneOf:
+          - $ref: "#/components/schemas/EnvironmentPatch"
+          - $ref: "#/components/schemas/BuildParametersPatch"
         - type: object
           properties:
-            environment_kind:
-              $ref: "#/components/schemas/EnvironmentKind"
+          environment_kind:
+            $ref: "#/components/schemas/EnvironmentKind"
```
leafty, line 402: Schema does not validate:

```suggestion
            environment_kind:
              $ref: "#/components/schemas/EnvironmentKind"
```

components/renku_data_services/session/api.spec.yaml
```diff
@@ -301,41 +301,49 @@ components:
...
-        - $ref: "#/components/schemas/EnvironmentPost"
+        - oneOf:
+          - $ref: "#/components/schemas/EnvironmentPost"
+          - $ref: "#/components/schemas/BuildParametersPost"
         - type: object
           properties:
-            environment_kind:
-              $ref: "#/components/schemas/EnvironmentKind"
+          environment_kind:
+            $ref: "#/components/schemas/EnvironmentKind"
```
leafty, line 344: Invalid schema:

```suggestion
            environment_kind:
              $ref: "#/components/schemas/EnvironmentKind"
```

components/renku_data_services/session/api.spec.yaml
```diff
@@ -377,16 +385,23 @@ components:
...
       required:
         - name
         - container_image
+        - environment_image_source
     EnvironmentPatchInLauncher:
       allOf:
-        - $ref: "#/components/schemas/EnvironmentPatch"
+        - oneOf:
+          - $ref: "#/components/schemas/EnvironmentPatch"
+          - $ref: "#/components/schemas/BuildParametersPatch"
```
leafty, line 398: Note: what if I want to change `default_url` with `environment_image_source=build`?

components/renku_data_services/session/api.spec.yaml
```diff
@@ -301,41 +301,49 @@ components:
...
+        - $ref: "#/components/schemas/Environment"
+        - type: object
+          properties:
+            environment_image_source:
+              $ref: "#/components/schemas/EnvironmentImageSourceImage"
+          required:
+            - environment_image_source
+    EnvironmentWithBuildGet:
+      allOf:
+        - $ref: "#/components/schemas/Environment"
```
leafty, line 326: This has `container_image` as required, non-nullable with a regex pattern. What value is this going to be before the first image is built?

leafty (commented):
There are two invalid schema errors (see below). There are more potential pitfalls (see below), but we can start from here.

components/renku_data_services/session/api.spec.yaml
m-alisafaee: Good point! I've made it optional in this case.

components/renku_data_services/session/api.spec.yaml
```diff
@@ -539,6 +589,38 @@ components:
...
+          $ref: "#/components/schemas/Repository"
+        builder_variant:
+          $ref: "#/components/schemas/BuilderVariant"
+        frontend_variant:
+          $ref: "#/components/schemas/FrontendVariant"
+      required:
+        - repository
+        - builder_variant
+        - frontend_variant
+    BuildParametersPost:
```
leafty, line 607: We need to keep `BuildParametersPatch` as well to allow partial update of the parameters.

leafty: It forces ALL parameters to be given, no partials.

m-alisafaee: Right, I've added a `BuildParametersPatch` with optional fields.

leafty (commented):
:+1: OK for `api.spec.yaml` 🙌

components/renku_data_services/session/models.py
```diff
@@ -25,6 +25,44 @@ class EnvironmentKind(StrEnum):
...
+    conda = "conda"
+    pip = "pip"
+
+
+class FrontendVariant(StrEnum):
+    """The environment frontend choice."""
+
+    vscodium = "vscodium"
+    jupyterlab = "jupyterlab"
+    streamlit = "streamlit"
```
leafty, line 47: Remove, as not supported.

```suggestion
```

components/renku_data_services/session/models.py
```diff
@@ -25,6 +25,44 @@ class EnvironmentKind(StrEnum):
...
+
+    image = "image"
+    build = "build"
+
+
+class BuilderVariant(StrEnum):
+    """The type of environment builder."""
+
+    conda = "conda"
+    pip = "pip"
```
leafty, line 39: The builder will just build anything Python, so the choice is only "python" at the moment.

```suggestion
    python = "python"
```

leafty (commented):
Small comment about the parameters 👇

olevski (changes requested):
A few minor comments and questions.

components/renku_data_services/session/orm.py
```diff
@@ -163,3 +181,27 @@ def dump(self) -> models.SessionLauncher:
...
+    __tablename__ = "build_parameters"
+
+    id: Mapped[ULID] = mapped_column("id", ULIDType, primary_key=True, default_factory=lambda: str(ULID()), init=False)
+    """Id of this session build parameters object."""
+
+    repository: Mapped[str] = mapped_column("repository", String(500))
+
+    builder_variant: Mapped[str] = mapped_column("builder_variant", String(99))
+
+    frontend_variant: Mapped[str] = mapped_column("frontend_variant", String(99))
```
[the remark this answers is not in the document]
leafty: No, because the options are too variable. In the future, these may be defined by an admin. It would be way too painful to require an API spec change each time an option is added or modified.

components/renku_data_services/session/api.spec.yaml
```diff
@@ -368,25 +386,32 @@ components:
...
           $ref: "#/components/schemas/EnvironmentCommand"
         args:
           $ref: "#/components/schemas/EnvironmentArgs"
         is_archived:
-          $ref: "#/components/schemas/IsArchived"
+          allOf:
+            - $ref: "#/components/schemas/IsArchived"
           default: false
+        environment_image_source:
+          $ref: "#/components/schemas/EnvironmentImageSourceImage"
```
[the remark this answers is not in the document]
leafty: This field is returned in responses, see `GET /projects/<project_id>/session_launchers`:

```json
[
    {
        "id": "01JKWRH07VXW873VQCCDKTPN71",
        "project_id": "01JHTC2SE2CNJQC2KMMEZNA8J6",
        "name": "Flora Test",
        "creation_date": "2025-02-12T09:41:00Z",
        "description": "Something",
        "environment": {
            "id": "01JKWRP3H8W8C2819HKB270AWT",
            "name": "Flora Test",
            "creation_date": "2025-02-12T09:43:47Z",
            "description": "Generated environment for Flora Test",
            "default_url": "/lab",
            "uid": 1000,
            "gid": 1000,
            "port": 8888,
            "is_archived": false,
            "container_image": "image:unknown-at-the-moment",
            "build_parameters": {
                "repository": "https://github.com/SwissDataScienceCenter/renku-ui.git",
                "builder_variant": "pip",
                "frontend_variant": "vscodium"
            },
            "environment_image_source": "build",
            "environment_kind": "CUSTOM"
        },
        "resource_class_id": 3
    },
    {
        "id": "01JHTCXVTXJM92N8BY11S12ANZ",
        "project_id": "01JHTC2SE2CNJQC2KMMEZNA8J6",
        "name": "RenkuLab Python",
        "creation_date": "2025-01-17T15:08:26Z",
        "description": "test",
        "environment": {
            "id": "01JHTCXVS1H94HCA3A75RX9ABG",
            "name": "RenkuLab Python",
            "creation_date": "2025-01-17T15:08:26Z",
            "default_url": "/lab",
            "uid": 1000,
            "gid": 1000,
            "port": 8888,
            "is_archived": false,
            "container_image": "renku/renkulab-py:3.11-0.25.0",
            "environment_image_source": "image",
            "environment_kind": "CUSTOM"
        },
        "resource_class_id": 1
    }
]
```

components/renku_data_services/session/api.spec.yaml
```diff
@@ -296,36 +294,57 @@ components:
...
+              $ref: "#/components/schemas/EnvironmentImageSourceBuild"
+            environment_kind:
+              allOf:
+                - $ref: "#/components/schemas/EnvironmentKind"
+              default: custom
+          required:
+            - build_parameters
+            - environment_image_source
+            - environment_kind
+    EnvironmentPostInLauncherHelper:
```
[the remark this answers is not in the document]
leafty: Session environments with `environment_kind = CUSTOM` should have the same life cycle as their parent session launcher. How would this work then? Do we then accept that there are going to be orphan session environments?

components/renku_data_services/session/orm.py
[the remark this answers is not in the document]
olevski: Ok yeah we can leave this as is. But these are part of the API whether we put them explicitly in the spec or we hide them.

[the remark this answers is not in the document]
leafty: Well, for now. On the future we can let admins add options to a database of build recipes and validate against that.

components/renku_data_services/session/core.py
```diff
@@ -26,6 +27,63 @@ def validate_unsaved_environment(
...
+        )
+
+    return models.UnsavedBuildParameters(
+        repository=environment.repository,
+        builder_variant=environment.builder_variant,
+        frontend_variant=environment.frontend_variant,
+    )
+
+
+def validate_build_parameters_patch(environment: apispec.BuildParametersPatch) -> models.BuildParametersPatch:
```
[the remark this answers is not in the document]
leafty: At the moment, the UI validates the choices (needs to be a repository which responds to the probe API). We can at least validate that the repository is listed in the project in the backend here.

leafty (commented):
[can be solved later] Sending a PATCH request to modify the description of the session launcher resets the session environment fields which is not desirable (the session cannot be launched).

components/renku_data_services/session/blueprints.py
```diff
@@ -135,12 +135,22 @@ def patch(self) -> BlueprintFactoryResponse:
...
+                # NOTE: This is required to deal with the multiple possible types for the environment field: If some
+                # random fields are passed then the validation chooses the environment type to be EnvironmentIdOnlyPatch
+                # which might not be the case and would set the session's environment ID to None.
+                # TODO: Check how validation exactly works for Union types to see if we can do this in a clear way.
+                if isinstance(body.environment, apispec.EnvironmentIdOnlyPatch) and "id" not in request.json.get(
+                    "environment", {}
+                ):
+                    raise errors.ValidationError(
+                        message="There are errors in the following fields, id: Input should be a valid string"
+                    )
```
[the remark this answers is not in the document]
mohammad-alisafaee: The issue is that, `environment` can be one of `EnvironmentIdOnlyPatch` or `EnvironmentPatchInLauncher` and it's `model_validate` that decides which one to return when validating. When the data contains some invalid fields, that belong to none of these two, it gets confused and returns `EnvironmentIdOnlyPatch` (instead of raising an error). As I wrote in the comment above, this is a problem with all Union types. So, I had to do this check manually to make sure that such cases (i.e. invalid fields in data) are correctly covered.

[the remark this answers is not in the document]
mohammad-alisafaee: There's a test that will fail if we don't do this check: https://github.com/SwissDataScienceCenter/renku-data-services/blob/c8fc9936089947811a99f6ebb64dc8937c154935/test/bases/renku_data_services/data_api/test_sessions.py#L613-L613

components/renku_data_services/session/core.py
```diff
@@ -26,6 +27,63 @@ def validate_unsaved_environment(
...
+def validate_unsaved_build_parameters(
+    environment: apispec.BuildParameters | apispec.BuildParametersPatch,
+) -> models.UnsavedBuildParameters:
+    """Validate an unsaved build parameters object."""
+    if environment.builder_variant is None:
+        raise errors.ValidationError(message="The field 'builder_variant' is required")
+    if environment.frontend_variant is None:
+        raise errors.ValidationError(message="The field 'frontend_variant' is required")
+    if environment.repository is None:
+        raise errors.ValidationError(message="The field 'repository' is required")
```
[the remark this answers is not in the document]
mohammad-alisafaee: A `PATCH` request shouldn't require them if users don't want to change them, but when a change is requested all fields must be provided. That's why we can't make them required in the spec.

components/renku_data_services/session/core.py
```diff
@@ -26,6 +27,63 @@ def validate_unsaved_environment(
...
+                f"Valid values are {[e.value for e in models.BuilderVariant]}"
+            )
+        )
+    if environment.frontend_variant not in models.FrontendVariant:
+        raise errors.ValidationError(
+            message=(
+                f"Invalid value for the field 'frontend_variant': {environment.frontend_variant}: "
+                f"Valid values are {[e.value for e in models.FrontendVariant]}"
+            )
+        )
```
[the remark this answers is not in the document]
mohammad-alisafaee: As Flora mentioned in her comment, we decided to make strings to allow changing them without changing the spec everytime.

components/renku_data_services/session/orm.py
```diff
@@ -54,7 +54,13 @@ class EnvironmentORM(BaseORM):
     mount_directory: Mapped[PurePosixPath | None] = mapped_column("mount_directory", PurePosixPathType, nullable=True)
     uid: Mapped[int] = mapped_column("uid")
     gid: Mapped[int] = mapped_column("gid")
-    environment_kind: Mapped[models.EnvironmentKind] = mapped_column("environment_kind")
+    environment_kind: Mapped[models.EnvironmentKind] = mapped_column(
+        "environment_kind", Enum(models.EnvironmentKind, values_callable=lambda e: [v.value for v in e])
+    )
```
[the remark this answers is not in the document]
mohammad-alisafaee: No. This was required when changing the case of the `EnvironmentKind` to lowercase, to tell SQLAlchemy to use enum's values for the field (by default, it uses enum names). I'll remove this since it's not required ATM.

test/bases/renku_data_services/data_api/conftest.py
```diff
@@ -211,6 +211,26 @@ async def sanic_client(
...
 
 
+@pytest_asyncio.fixture
+async def post(sanic_client, user_headers, admin_headers):
+    async def post_helper(url: str, admin: bool = False, payload: dict[str, Any] = None) -> TestingResponse:
+        headers = admin_headers if admin else user_headers
+        _, response = await sanic_client.post(url, headers=headers, json=payload)
+        return response
+
+    return post_helper
```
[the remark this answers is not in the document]
mohammad-alisafaee: I used these in some tests, but then removed them and forgot to remove the fixtures. I'll apply your suggestion if I decided to keep them.

components/renku_data_services/project/api.spec.yaml
```diff
@@ -769,9 +769,8 @@ components:
         - https://github.com/SwissDataScienceCenter/project-1.git
         - git@github.com:SwissDataScienceCenter/project-2.git
     Repository:
-      description: A project's repository
+      description: A git repository URL
       type: string
-      example: git@github.com:SwissDataScienceCenter/project-1.git
```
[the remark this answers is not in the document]
leafty: The point is to remove examples from `object` not from basic fields. Examples for basic fields are useful and should be given. This often helps making the example payloads realistic.
Example for objects are generated by swagger by assembling the examples from the fields in the object when there is no explicit example, so swagger often comes up with something better than what maintainers do by hand (because they rarely update the example when the schema changes).

In this case, we should have a https+git URL (since we do not support ssh+git URLs), e.g. `https://github.com/SwissDataScienceCenter/renku.git`.

components/renku_data_services/session/api.spec.yaml
```diff
@@ -296,36 +294,57 @@ components:
...
-        description: JupyterLab session environment
-        container_image: renku-jupyter:latest
-        default_url: "/lab"
-        port: 8080
-        working_directory: /home/jovyan/work
-        mount_directory: /home/jovyan/work
-        uid: 1000
-        gid: 1000
-        is_archive: false
+    Environment:
```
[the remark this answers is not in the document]
olevski: Ok leave for cooldown.

olevski (approved):

PR #615, opened 2025-01-22
feat: authz schema changes for data connectors

olevski:
NOTE: This is targeting a build / feature branch.

This adds the initial changes for the authorization schema required for the project disentanglement pitch.

Below is a diff of the old and new schemas. I did not want to rename any relations for now. I will do this towards the end of the build in a separate PR.

Screenshot From 2025-03-06 02-07-33


olevski:
NOTE: the tests here fail but that is fine. This is targeting a feature branch and the next PR in the stack fixes all the tests.

eikek (approved):
lgtm

PR #617, opened 2025-01-23
feat: make projects dc owners in authzed

olevski:
squashme: store entity slugs for dcs in projects

squashme: update the apispec for data connectors

squashme: initial code changes for data connectors


olevski:

- #617 ◀

components/renku_data_services/base_models/core.py
```diff
@@ -207,6 +207,62 @@ def __repr__(self) -> str:
...
+            slugs.extend(other.slugs)
+        else:
+            raise errors.ValidationError(
+                message="A path can be constructed from an entity path and a slug,string or entity path, "
+                f"but the 'divisor' is of type {type(other)}"
+            )
+        return EntityPath(slugs)
+
+    @classmethod
+    def join(cls, *slugs: Slug | str) -> Self:
```
Panaetius, line 236: suggestion: I think truediv can be expressed using join or the other way around, whichever is easier, to reduce code duplication.

components/renku_data_services/base_models/core.py
```diff
@@ -207,6 +207,62 @@ def __repr__(self) -> str:
...
+                )
+        return cls(res)
+
+    @classmethod
+    def from_string(cls, path: str) -> Self:
+        """Create an entity path from a single string which may contain several slugs."""
+        res: list[Slug] = []
+        for slug in path.split("/"):
+            res.append(Slug(slug))
+        return cls(res)
```
Panaetius, line 257:
```suggestion
        return cls.join(path.split("/"))
```

components/renku_data_services/base_models/core.py
```diff
@@ -207,6 +207,62 @@ def __repr__(self) -> str:
         return self.value
 
 
+@dataclass(frozen=True, eq=True)
+class EntityPath:
```
Panaetius, line 211: suggestion: should there be some validation here that it only has 1 or 2 segments and not more? 

Or more generally, have subclasses for the different slugs that validate number of segments, and then the typechecker can make sure that you don't pass a dataset slug to a method that expects a namespace?

if this class (or a subclass) had a limit to 1 or 2, you could also add properties like `first` and `second` (terrible names, just to illustrate) and set `__match_args__ = ("first","second")` and then some of the code below could just be

```
path = EntityPath.from_string(body.namespace)
namespace, project = path # could just pattern match directly on the line above as well
```

But maybe that's overengineering, hence just a suggestion :man_shrugging:

components/renku_data_services/data_connectors/db.py
```diff
@@ -263,6 +314,9 @@ async def update_data_connector(
             raise errors.MissingResourceError(message=not_found_msg)
         old_data_connector = data_connector.dump()
 
+        if old_data_connector.project and old_data_connector.project.slug != patch.project_slug:
+            raise NotImplementedError("Moving the data connector to another project or namespace is not supported")
```
Panaetius, line 318:
```suggestion
            raise NotImplementedError("Moving a data connector to another project or namespace is not supported")
```

otherwise it sounds like there's something wrong with this dataconnector specifically that makes it so it can't be moved

Panaetius (changes requested):

components/renku_data_services/base_models/core.py
olevski: Done

components/renku_data_services/base_models/core.py
olevski: No that is a good idea. Because in a subsequent PR I use it in a match/case statement. So that is pretty much exactly what I need. I will add this.

olevski: Done in https://github.com/SwissDataScienceCenter/renku-data-services/pull/617/commits/81f1b6ccdcc254da6825fb6b7729d068632f2259

Panaetius (approved):
this is really nice now!

PR #635, opened 2025-02-06
fix: use correct uid/gid when cloning git repos

### LINKED ISSUE #629: Sessions: git clone does not respect gid/uid the session will run as
Reproduce: use gid/uid = `1001:1001`

The repositories will be cloned as `100:1000` which makes the repositories not usable in the session

Running `git-clone` as the proper user works, so the solution is to update the spec of the `git-clone` container.

olevski:
This was hardcoded for the old sessions so we just re-used the same hardcoded values from v1 sessions. But for v2 we should take this info from the environment.

closes #629 

/deploy

### DIFF
--- a/components/renku_data_services/notebooks/api/amalthea_patches/git_proxy.py
+++ b/components/renku_data_services/notebooks/api/amalthea_patches/git_proxy.py
@@ -21,9 +21,6 @@ async def main_container(
     config: NotebooksConfig,
     repositories: list[Repository],
     git_providers: list[GitProvider],
-    uid: int = 1000,
-    gid: int = 1000,
-    fs_group: int = 100,
 ) -> client.V1Container | None:
     """The patch that adds the git proxy container to a session statefulset."""
     if not user.is_authenticated or not repositories or user.access_token is None or user.refresh_token is None:
@@ -70,9 +67,8 @@ async def main_container(
     container = client.V1Container(
         image=config.sessions.git_proxy.image,
         security_context={
-            "fsGroup": fs_group,
-            "runAsGroup": gid,
-            "runAsUser": uid,
+            "runAsGroup": 1000,
+            "runAsUser": 1000,
             "allowPrivilegeEscalation": False,
             "runAsNonRoot": True,
         },
--- a/components/renku_data_services/notebooks/api/amalthea_patches/git_proxy.py
+++ b/components/renku_data_services/notebooks/api/amalthea_patches/git_proxy.py
@@ -21,6 +21,9 @@ async def main_container(
     config: NotebooksConfig,
     repositories: list[Repository],
     git_providers: list[GitProvider],
+    uid: int = 1000,
+    gid: int = 1000,
+    fs_group: int = 100,
 ) -> client.V1Container | None:
     """The patch that adds the git proxy container to a session statefulset."""
     if not user.is_authenticated or not repositories or user.access_token is None or user.refresh_token is None:
@@ -67,9 +70,9 @@ async def main_container(
     container = client.V1Container(
         image=config.sessions.git_proxy.image,
         security_context={
-            "fsGroup": 100,
-            "runAsGroup": 1000,
-            "runAsUser": 1000,
+            "fsGroup": fs_group,
+            "runAsGroup": gid,
+            "runAsUser": uid,
             "allowPrivilegeEscalation": False,
             "runAsNonRoot": True,
         },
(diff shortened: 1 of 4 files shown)


leafty (commented):
Thanks! Let's deploy this to check that it works as intended.

components/renku_data_services/notebooks/api/amalthea_patches/git_proxy.py
```diff
@@ -67,9 +70,9 @@ async def main_container(
     container = client.V1Container(
         image=config.sessions.git_proxy.image,
         security_context={
-            "fsGroup": 100,
-            "runAsGroup": 1000,
-            "runAsUser": 1000,
+            "fsGroup": fs_group,
+            "runAsGroup": gid,
+            "runAsUser": uid,
```
leafty, line 75: This container does not need to be updated. In fact, it should not.

components/renku_data_services/notebooks/api/amalthea_patches/git_proxy.py
```diff
@@ -67,9 +70,9 @@ async def main_container(
     container = client.V1Container(
         image=config.sessions.git_proxy.image,
         security_context={
-            "fsGroup": 100,
-            "runAsGroup": 1000,
-            "runAsUser": 1000,
+            "fsGroup": fs_group,
```
leafty, line 73: `fsGroup` is part of [PodSecurityContext](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#podsecuritycontext-v1-core), not part of [SecurityContext](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#securitycontext-v1-core).

That means it applies to the pod spec but not to the containers specs.

components/renku_data_services/notebooks/blueprints.py
```diff
@@ -334,14 +334,22 @@ async def _handler(
...
             )
-            extra_containers = await get_extra_containers(self.nb_config, user, repositories, git_providers)
+            extra_containers = await get_extra_containers(
+                self.nb_config,
+                user,
+                repositories,
+                git_providers,
+                uid=environment.uid,
+                gid=environment.gid,
+                fs_group=environment.gid,
```
leafty, line 344: No `fs_group` in containers.

```suggestion
```

components/renku_data_services/notebooks/core_sessions.py
```diff
@@ -79,11 +79,20 @@ async def get_extra_containers(
     user: AnonymousAPIUser | AuthenticatedAPIUser,
     repositories: list[Repository],
     git_providers: list[GitProvider],
+    uid: int,
+    gid: int,
+    fs_group: int,
```
leafty, line 84:
```suggestion
```

components/renku_data_services/notebooks/core_sessions.py
```diff
@@ -79,11 +79,20 @@ async def get_extra_containers(
...
     conts: list[ExtraContainer] = []
     git_proxy_container = await git_proxy.main_container(
-        user=user, config=nb_config, repositories=repositories, git_providers=git_providers
+        user=user,
+        config=nb_config,
+        repositories=repositories,
+        git_providers=git_providers,
+        uid=uid,
+        gid=gid,
+        fs_group=fs_group,
```
leafty, line 95:
```suggestion
```

components/renku_data_services/notebooks/core_sessions.py
```diff
@@ -428,6 +441,9 @@ async def patch_session(
         user,
         repositories,
         git_providers,
+        gid=session.spec.session.runAsGroup,
+        fs_group=session.spec.session.runAsGroup,
```
leafty, line 445:
```suggestion
```

leafty (changes requested):
The container to fix is `git-clone`. The `git-proxy` container should not be updated (it does not touch the file system).

components/renku_data_services/notebooks/api/amalthea_patches/git_proxy.py
olevski: Removed it

olevski:
Well jupyter still does not work when you use non 1001 uid and gid. I see this error in the logs:

```
File "/opt/conda/lib/python3.11/pathlib.py", line 1116, in mkdir                                                                                                   │
│     os.mkdir(self, mode)                                                                                                                                             │
│ PermissionError: [Errno 13] Permission denied: '/home/jovyan/.local/share/jupyter/runtime'
```

But that is not our fault. At least we are doing the right thing and respecting the stuff the user setup. Some images will only work with specific uid/gid.

leafty (approved):
LGTM Thanks 👍

PR #643, opened 2025-02-12
fix: mount cloud storage in work dir, not (pvc)mount path

### LINKED ISSUE #631: Sessions: `working_directory` is not respected when different from `mount_directory`
Repositories and remote data folders should be mounted at `working_directory`, not `mount_directory`.

Panaetius:
closes #631 

/deploy renku-notebooks=repos-in-workdir


olevski (approved):

leafty (approved):
👍 Tested in deployment

PR #646, opened 2025-02-13
refactor: make rclone storage patches consistent, refactor the code, add tests

### LINKED ISSUE #607: change rclone webdav customisation to be in line with others
currently the WebDAv customisation for the rclone schema is invoked in a different way than all the other patches ([called directly](https://github.com/SwissDataScienceCenter/renku-data-services/blob/main/components/renku_data_services/storage/rclone.py#L267) vs. [found by the `__patch` prefix](https://github.com/SwissDataScienceCenter/renku-data-services/blob/main/components/renku_data_services/storage/rclone.py#L257))

This should be made consistent.

### LINKED ISSUE #608: add more tests for storage patches
right now there's only a single test for a storage patch, https://github.com/SwissDataScienceCenter/renku-data-services/blob/main/test/bases/renku_data_services/data_api/test_storage.py#L602

we should make sure all the patches (e.g. https://github.com/SwissDataScienceCenter/renku-data-services/blob/main/components/renku_data_services/storage/rclone.py#L62) are tested so they don't inadvertently get lost

Panaetius:
closes #607 
closes #608


components/renku_data_services/storage/rclone_patches.py
```diff
@@ -0,0 +1,264 @@
...
+    "mailru",
+    "onedrive",
+    "pcloud",
+    "pikpak",
+    "premiumzeme",
+    "putio",
+    "sharefile",
+    "yandex",
+    "zoho",
+]
```
olevski, line 39: suggestion: It would be nice if these are both sets probably. Also it does not hurt to use `typing.Final` on them to make sure no one modifies them in some function somewhere else.

components/renku_data_services/storage/rclone_patches.py
```diff
@@ -0,0 +1,264 @@
...
+        spec,
+        prefix="polybox",
+        name="PolyBox",
+        description="Polybox",
+        url_value="https://polybox.ethz.ch/remote.php/webdav/",
+        public_link_help="Shared folder link. E.g., https://polybox.ethz.ch/index.php/s/8NffJ3rFyHaVyyy",
+    )
+
+
+def patch_switchdrive_storage(spec: list[dict[str, Any]]) -> list[dict[str, Any]]:
```
olevski, line 238: suggestion: I think it does not hurt to make this function and all others non public and only make `apply_pathes` public. Just to make sure people are not trying to do funny things reusing these functions elsewhere. The suggestion below requires edits for the function wherever it is used in this file.

```suggestion
def __patch_switchdrive_storage(spec: list[dict[str, Any]]) -> list[dict[str, Any]]:
```

components/renku_data_services/storage/rclone_patches.py
```diff
@@ -0,0 +1,264 @@
...
+
+    This returns the original entry for in-place modification.
+    """
+    storage = next((s for s in spec if s["Prefix"] == prefix), None)
+    if not storage:
+        raise errors.ValidationError(message=f"'{prefix}' storage not found in schema.")
+    return storage
+
+
+def patch_schema_remove_unsafe(spec: list[dict[str, Any]]) -> list[dict[str, Any]]:
```
olevski, line 53: question: I am being really nitpicky here. But should we just not return anything from any of these functions? Because they all modify the spec in place as far as I can tell. So there is really no need to return anything. We can also keep copying and then return stuff OR we dont return anything and modify in place OR we at least mention that we modify in place in the doc strings. But I personally think that just not returning anything makes it really clear what is going on.

olevski (changes requested):
Thanks for doing this Ralf! just a few minor things.

olevski (approved):

PR #647, opened 2025-02-17
refactor: use the k8s cache for buildruns

leafty:
PR Stack:
* #625
  * (this) #647
    * #648
      * #657
      * #656

Related PRs:
* https://github.com/SwissDataScienceCenter/renku-notebooks/pull/2007
* https://github.com/SwissDataScienceCenter/renku/pull/3878

/deploy #notest renku=build/session-env-builders renku-ui=leafty/session-env-builders-5 renku-notebooks=leafty/shipwright-buildrun-cache extra-values=dataService.imageBuilders.enabled=true,dataService.imageBuilders.outputImagePrefix=harbor.dev.renku.ch/flora-dev/,dataService.imageBuilders.pushSecretName=flora-docker-secret

PR #648, opened 2025-02-17
feat: get image build logs

leafty:
PR Stack:
* #625
  * #647
    * (this) #648
      * #657
      * #656

Important note: the code generator cannot be used for Tekton CRDs because this is what they use in the API spec:
```yaml
type: object
# One can use x-kubernetes-preserve-unknown-fields: true
# at the root of the schema (and inside any properties, additionalProperties)
# to get the traditional CRD behaviour that nothing is pruned, despite
# setting spec.preserveUnknownProperties: false.
#
# See https://kubernetes.io/blog/2019/06/20/crd-structural-schema/
# See issue: https://github.com/knative/serving/issues/912
x-kubernetes-preserve-unknown-fields: true
```
In other words, there is no API spec!  (In openAPI format)

/deploy #notest renku=build/session-env-builders renku-ui=leafty/session-env-builders-5 renku-notebooks=leafty/shipwright-buildrun-cache extra-values=dataService.imageBuilders.enabled=true,dataService.imageBuilders.outputImagePrefix=harbor.dev.renku.ch/flora-dev/,dataService.imageBuilders.pushSecretName=flora-docker-secret

### DIFF
--- a/components/renku_data_services/session/k8s_client.py
+++ b/components/renku_data_services/session/k8s_client.py
@@ -6,6 +6,7 @@ import httpx
 from kr8s import NotFoundError, ServerError
 from kr8s.asyncio.objects import APIObject, Pod
 from kubernetes.client import ApiClient
+from sanic.log import logger
 
 from renku_data_services import errors
 from renku_data_services.errors.errors import CannotStartBuildError


components/renku_data_services/session/db.py
```diff
@@ -876,6 +876,51 @@ async def update_build(self, user: base_models.APIUser, build_id: ULID, patch: m
...
+                    )
+            if not authorized:
+                raise errors.MissingResourceError(
+                    message=f"Build with id '{build_id}' does not exist or you do not have access to it."
+                )
+
+        build_model = build.dump()
+
+        if self.shipwright_client is None:
+            raise errors.MissingResourceError(message=f"Build with id '{build_id}' does not have logs.")
```
Panaetius, line 918: question: shouldn't this be a 5xx server error? this happening would mean the server is not set up correctly?

leafty: For now, `shipwright_client` is meant to be `None` in two cases: when the feature is not enabled and when running tests. In the future, it would be nicer to setup mocks like in `components/renku_data_services/k8s/clients.py` to improve tests.
In https://github.com/SwissDataScienceCenter/renku-data-services/pull/656 (not ready yet), it should not be possible to run code which would touch shipwright when the feature is not enabled (the routes are not enabled).

Panaetius (approved):

Issue #649, opened 2025-02-17
Copying projects does not copy session environments


leafty:
There is a bug with copying projects: when a project with session launchers using custom environments are copied, the session launchers are copied but not the session environments. This means that users can mess up session environments on the source project.
Reproduce:
1. Make a project and then create a session launcher with the custom image option
2. Copy the project.
3. Update the session launcher on the copied project, e.g. change the image or change the URL or change the port
4. Visit the source project page and reload -> the session launcher has been updated here too!

PR #651, opened 2025-02-17
fix: create new custom environment when copying project

mohammad-alisafaee:
Fixes: #649

### DIFF
--- a/components/renku_data_services/session/db.py
+++ b/components/renku_data_services/session/db.py
@@ -89,6 +89,37 @@ class SessionRepository:
         session.add(environment)
         return environment
 
+    def __copy_environment(
+        self,
+        user: base_models.APIUser,
+        session: AsyncSession,
+        environment: models.Environment,
+    ) -> schemas.EnvironmentORM:
+        if user.id is None:
+            raise errors.UnauthorizedError(
+                message="You have to be authenticated to insert an environment in the DB.", quiet=True
+            )
+        new_environment = schemas.EnvironmentORM(
+            name=environment.name,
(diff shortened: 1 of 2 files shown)


components/renku_data_services/session/db.py
```diff
@@ -89,6 +89,37 @@ def __insert_environment(
...
+            default_url=environment.default_url,
+            port=environment.port,
+            working_directory=environment.working_directory,
+            mount_directory=environment.mount_directory,
+            uid=environment.uid,
+            gid=environment.gid,
+            environment_kind=environment.environment_kind,
+            command=environment.command,
+            args=environment.args,
+            creation_date=datetime.now(UTC).replace(microsecond=0),
```
leafty, line 116: Non-blocking note: why do we even do this `.replace(microsecond=0)`? This is not useful and the ULID will already contain millisecond-precision information.

olevski (approved):

components/renku_data_services/session/db.py
olevski: I have no clue @leafty I have seen it in a few places in the code and wondered myself.

Panaetius: https://github.com/SwissDataScienceCenter/renku-data-services/pull/175/commits/fbeae73ba0832bf7b4f53dbd0e8fbffe74914603#diff-e6ad745574f64b4241525aaa0d39f7736545db13f97f9092f00b723257970b9cR228 is the first occurrence of it that I could find but it doesn't offer any clue on why it was added.

PR #655, opened 2025-02-18
chore: cleanup unneeded code

Panaetius:
removes references to kpack as we decided to go with shipwright instead

I also merged in main and fixed some broken tests


test/components/renku_pack_builder/test_environment_build.py
leafty, line 1: If some of tests here do not seem to work, should they be marked as skipped for now?

leafty (approved):
Thanks!

test/components/renku_pack_builder/test_environment_build.py
Panaetius: done. I'm not even sure what that test is supposed to test? that shipwright/buildruns works correctly?

PR #660, opened 2025-02-20
fix: migrate environments from copied projects

olevski:
There is a bug where environments were not copied when a project was copied. So the environment was shared between the original project and the copied project. This migration makes new environments (copied from the original project) for all copied projects.

/deploy


leafty (changes requested):
> template_id is set when copying a project from a template project and it cannot be modified. This field can be either null or an empty string; a null value won't change it while an empty string value will delete it, meaning that the project is unlinked from its template

This means we MUST look for duplicate environment IDs with `kind == CUSTOM` to find all problematic projects and session launchers. Looking only at template_id is not enough because it can be changed after a copy is made.

olevski:
> Looking only at template_id is not enough because it can be changed after a copy is made.
Ok thanks good catch. I did not realize that that projects can be unlinked.

Panaetius (approved):

olevski:
related to #649

PR #666, opened 2025-02-21
feat: sync entity changes to the solr staging table

eikek:
- Add a decorator that intercepts db calls to additionally update the search staging table

/deploy renku=build/search-migration-part1

### DIFF
--- a/components/renku_data_services/search/core.py
+++ b/components/renku_data_services/search/core.py
@@ -88,6 +88,8 @@ async def update_solr(search_updates_repo: SearchUpdatesRepo, solr_client: SolrC
             else:
                 counter = counter + len(entries)
                 await search_updates_repo.mark_processed(ids)
+
+            await solr_client.delete("deleted:true")
         except Exception as e:
             logger.error(f"Error while updating solr with entities {ids}", exc_info=e)
             try:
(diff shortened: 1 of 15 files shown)


components/renku_data_services/search/decorators.py
```diff
@@ -0,0 +1,85 @@
+"""Message queue implementation for redis streams."""
```
eikek, line 1: fixed that one, just waiting for the deploy to push the commit

components/renku_data_services/search/core.py
```diff
@@ -88,6 +88,8 @@ async def update_solr(search_updates_repo: SearchUpdatesRepo, solr_client: SolrC
             else:
                 counter = counter + len(entries)
                 await search_updates_repo.mark_processed(ids)
+
+            await solr_client.delete("deleted:true")
```
olevski, line 92: Eike can you please put a comment here explaining what this does. I think you explained it to me. Where we "soft-delete" solr documents during syncing and this cleans up these soft-deleted entities from solr fully. Right? Withtout this we would keep accumulating soft deleted documents in solr?

olevski (changes requested):
Looks great Eike. Just one small request to add a comment for a single line because without some background / info it looks strange.

components/renku_data_services/search/core.py
eikek: oh sure! Yes, this correct (soft-delete). I also forgot to try-catch it.

olevski (approved):

PR #673, opened 2025-02-26
fix: set resource pool culling in AmaltheaSessions

olevski:
/deploy

We were not using the idle and max hibernation culling from the resource pools in v2 sessions.

### DIFF
--- a/components/renku_data_services/notebooks/blueprints.py
+++ b/components/renku_data_services/notebooks/blueprints.py
@@ -2,7 +2,6 @@
 
 from dataclasses import dataclass
 from pathlib import PurePosixPath
-from typing import cast
 from urllib.parse import urlparse
 
 from sanic import Request, empty, exceptions, json
@@ -31,6 +30,7 @@ from renku_data_services.notebooks.config import NotebooksConfig
 from renku_data_services.notebooks.core_sessions import (
     get_auth_secret_anonymous,
     get_auth_secret_authenticated,
+    get_culling,
     get_data_sources,
     get_extra_containers,
     get_extra_init_containers,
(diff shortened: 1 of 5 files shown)


components/renku_data_services/notebooks/blueprints.py
```diff
@@ -413,13 +417,7 @@ async def _handler(
...
                     initContainers=extra_init_containers,
                     extraVolumes=extra_volumes,
-                    culling=Culling(
-                        maxAge=f"{self.nb_config.sessions.culling.registered.max_age_seconds}s",
-                        maxFailedDuration=f"{self.nb_config.sessions.culling.registered.failed_seconds}s",
-                        maxHibernatedDuration=f"{self.nb_config.sessions.culling.registered.hibernated_seconds}s",
-                        maxIdleDuration=f"{self.nb_config.sessions.culling.registered.idle_seconds}s",
-                        maxStartingDuration=f"{self.nb_config.sessions.culling.registered.pending_seconds}s",
-                    ),
+                    culling=get_culling(resource_pool, self.nb_config),
```
rokroskar, line 420: is it possible to update this when a session is resumed? In case the pool policy changes, hibernated sessions that restart will keep the old policy but it would be nice if the current one could be applied.

rokroskar (commented):
thanks for adding this @olevski!

components/renku_data_services/notebooks/blueprints.py
Panaetius: if I'm reading this correctly, the resource pool is loaded from db each time and so should have up to date values, and the `get_culling` gives the values from the pool highest priority, so this should already do what you suggested?

olevski: And when the policy changes then a session that is hibernated and then resumed will use the new culling policy.

rokroskar (approved):

PR #675, opened 2025-02-27
feat: return build error reason

leafty:
Propagate the reason an image build failed by capturing the corresponding `condition.reason` from Shipwright.

See docs: [Understanding the state of a BuildRun](https://shipwright.io/docs/build/buildrun/#understanding-the-state-of-a-buildrun)

Example response:
```
GET /environments/01JN3MZ8E9WMQX34Z860ZT6NV3/builds
```
```json
[
    {
        "id": "01JN5SF67AHKYHEN2F6H5HYFJH",
        "environment_id": "01JN3MZ8E9WMQX34Z860ZT6NV3",
        "created_at": "2025-02-28T08:06:21.119234Z",
        "error_reason": "StepOutOfMemory",
        "status": "failed"
    }
]
```

Also, the build cancellation is now done according to Shipwright's cancellation docs (https://shipwright.io/docs/build/buildrun/#canceling-a-buildrun).

/deploy #notest renku=build/session-env-builders renku-notebooks=leafty/shipwright-buildrun-cache renku-ui=build/session-env-builders extra-values=dataService.imageBuilders.enabled=true,dataService.imageBuilders.pushSecretName=flora-docker-secret,dataService.imageBuilders.buildRunRetentionAfterFailedSeconds=86400,dataService.imageBuilders.nodeSelector.renku.io/node-purpose=image-build,dataService.imageBuilders.tolerations[0].key=renku.io/dedicated,dataService.imageBuilders.tolerations[0].operator=Equal,dataService.imageBuilders.tolerations[0].effect=NoSchedule,dataService.imageBuilders.tolerations[0].value=image-build,dataService.imageBuilders.outputImagePrefix=harbor.dev.renku.ch/flora-dev/


components/renku_data_services/session/k8s_client.py
```diff
@@ -105,6 +105,12 @@ async def delete_build_run(self, name: str) -> None:
             logger.exception(f"Cannot delete build {name} because of {e}")
         return None
 
+    async def cancel_build_run(self, name: str) -> BuildRun:
+        """Cancel a Shipwright BuildRun."""
+        build = await ShipwrightBuildRunV1Beta1Kr8s.get(name=name, namespace=self.namespace)
+        await build.patch({"spec": {"state": "BuildRunCanceled"}})
```
Panaetius, line 111: question: do we need to check if the build exists here?

components/renku_data_services/session/k8s_client.py
```diff
@@ -356,13 +366,13 @@ async def update_image_build_status(self, buildrun_name: str) -> models.Shipwrig
                 update=models.ShipwrightBuildStatusUpdateContent(
                     status=models.BuildStatus.failed,
                     completed_at=completion_time,
+                    error_reason=condition.reason if condition is not None else None,
                 )
             )
 
     async def cancel_image_build(self, buildrun_name: str) -> None:
```
Panaetius, line 373: question: what's the reason for this method? it's a bit odd to have a method that only exists to call another method. I don't see a reason not to call cancel_build_run directly in call-sites?

Panaetius (changes requested):

components/renku_data_services/session/k8s_client.py
leafty: If the buildrun does not exist, then this will return an error from the `.get()` call.

Note that this should not happen because this is called just after the status of the buildrun has been refreshed.

components/renku_data_services/session/k8s_client.py
leafty: OK, will remove.

leafty: Done in [95c6161](https://github.com/SwissDataScienceCenter/renku-data-services/pull/675/commits/95c616144bc8862699efb3a8afd6a81f2fd3f3ae).

Panaetius (approved):

PR #677, opened 2025-02-27
chore: update amalthea schemas to latest version

olevski:
There are non-breaking changes to the AmaltheaSessions CRD that we should pull in.

This does that.

/deploy


leafty (approved):
LGTM

PR #697, opened 2025-03-03
fix: handle missing solr core on startup

olevski:
/deploy

### DIFF
--- a/components/renku_data_services/solr/solr_migrate.py
+++ b/components/renku_data_services/solr/solr_migrate.py
@@ -151,7 +151,9 @@ class SchemaMigrator:
     async def __current_version0(self, client: DefaultSolrClient) -> VersionDoc | None:
         """Return the current schema version document."""
         resp = await client.get_raw(self.__docId)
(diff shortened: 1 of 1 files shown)


Panaetius (approved):

olevski:
The problem is that we need to create a new core in solr. But a 2nd or new core is not created by the bitnami helm chart if solr has already been initialized. And that is the case when this will hit dev.renku.ch or other places where renku has already been deployed.

The helm chart cannot be simply updated with this. We need to make a migration or job in the server do this - because solr will not do it.

leafty:
As far as I can see, the solr user account provisioned for data services has admin permissions. Would it be possible then to call an API to create the core as part of the migration?

eikek:
Oh that is really disappointing. It is possible to create a core using solr cli inside the solr container.

eikek:
I think, if we expect the core to exist, I would go with a 500 response if that is not the case.

olevski:
This will be addressed more holistically in another PR related to addressing #698

PR #699, opened 2025-03-04
fix: reuse custom environment when patching session launcher

mohammad-alisafaee:


components/renku_data_services/session/core.py
```diff
@@ -238,23 +239,26 @@ def validate_session_launcher_patch(
...
                 if current == "image" and new == "build":
-                    if not patch.environment.build_parameters:
-                        raise errors.ValidationError(
-                            message=(
-                                "There are errors in the following fields, environment.build_parameters: Must be set"
-                            )
-                        )
+                    assert patch.environment.build_parameters is not None  # Make mypy happy in the following statement
+                    # NOTE: The environment type is changed to be built, so, all required fields should be passed (as in
+                    # a POST request). No need to get values from the current env, since they will be set by the build.
```
Panaetius, line 252: I'm not sure assert is the right thing to do here. if this were to happen, it'd trigger an assertion error which probably isn't handled correctly by our error handling.

Is using `if patch.environment.build_parameters is None:` with the raise that we had before not enough to keep mypy happy (like changing the `not ...` to `is None`)?

components/renku_data_services/session/blueprints.py
```diff
@@ -128,6 +128,11 @@ def post(self) -> BlueprintFactoryResponse:
         async def _post(_: Request, user: base_models.APIUser, body: apispec.SessionLauncherPost) -> JSONResponse:
             new_launcher = validate_unsaved_session_launcher(body, builds_config=self.session_repo.builds_config)
             launcher = await self.session_repo.insert_launcher(user=user, launcher=new_launcher)
+
+            if isinstance(new_launcher.environment, models.UnsavedBuildParameters):
+                new_build = validate_unsaved_build(environment_id=launcher.environment.id)
+                await self.session_repo.start_build(user=user, build=new_build)
+
```
leafty, line 135: Opinion: there should be no business logic in the Blueprints file. The logic should exist in `db.py` or `core.py`.

test/bases/renku_data_services/data_api/test_sessions.py
```diff
@@ -976,40 +1043,6 @@ async def test_starting_session_anonymous(
     assert session_res.json["name"] in [i["name"] for i in res.json]
 
 
-@pytest.mark.asyncio
-async def test_post_build(sanic_client: SanicASGITestClient, user_headers, create_project) -> None:
```
leafty, line 980: Why is this test removed? We still need to check that triggering a build is working as expected.

components/renku_data_services/session/db.py
```diff
@@ -654,24 +623,73 @@ async def __update_launcher_environment(
...
+                launcher.environment.working_directory = None  # TODO: This should come from the build
+                launcher.environment.mount_directory = None  # TODO: This should come from the build
+                launcher.environment.uid = 1000  # TODO: This should come from the build
+                launcher.environment.gid = 1000  # TODO: This should come from the build
+                launcher.environment.environment_kind = models.EnvironmentKind.CUSTOM
+                launcher.environment.command = None  # TODO: This should come from the build
+                launcher.environment.args = None  # TODO: This should come from the build
+                launcher.environment.environment_image_source = models.EnvironmentImageSource.build
+                launcher.environment.build_parameters_id = build_parameters_orm.id
+                launcher.environment.build_parameters = build_parameters_orm
```
Panaetius, line 691: I know this is not really related to this PR(it was like this before) but I think we should handle this a bit better. Is there an issue for these TODOs? I think we'd either want some polymorphism in the db with an `UnbuiltEnvironment` type or to have a field in environment like `environment.state` that can be something like `not_built` and to allow the fields to be None (with dedicated types for built/unbuilt in code that validate that it's not None except for in this case)

leafty (commented):
Thank you @mohammad-alisafaee!

I will test that the deployment works as expected.

Some early comments from me below.

leafty (commented):
This works as expected when deployed. I had to do a small adjustment on the UI.

components/renku_data_services/session/core.py
mohammad-alisafaee: We are already checking that `patch.environment.build_parameters` is not `None` in the previous if block and this check was redundant. However, `mypy` isn't smart enough to deduct that `patch.environment.build_parameters` is not null. I'm not sure what's the general policy to handle such cases, but we've used assertion before. I can re-introduce the if/raise block if you think that's a better way of handling it (maybe it's more readable).

components/renku_data_services/session/db.py
mohammad-alisafaee: We'll do some work to store the build result and update the environment once a build finishes (there's a scope's task for that in the build note). I'll create an issue for it and add your comment there.

test/bases/renku_data_services/data_api/test_sessions.py
mohammad-alisafaee: We start the build when the session launcher is posted and starting the build again will fail. We can stop this build and then rebuild it again but that's what we exactly do in `test_get_build`. I'll re-introduce this test with some modification to at least keep the action name in the test.

Panaetius (approved):

PR #702, opened 2025-03-04
feat: copy projects with built environment

mohammad-alisafaee:


test/bases/renku_data_services/data_api/test_projects.py
```diff
@@ -1314,6 +1314,94 @@ async def test_project_copy_creates_new_custom_environment_instance(
...
+    assert environment["container_image"] == launcher["environment"]["container_image"]
+    assert environment["default_url"] == launcher["environment"]["default_url"]
+    assert environment["uid"] == launcher["environment"]["uid"]
+    assert environment["gid"] == launcher["environment"]["gid"]
+    assert environment.get("working_directory") == launcher["environment"].get("working_directory")
+    assert environment.get("mount_directory") == launcher["environment"].get("mount_directory")
+    assert environment["port"] == launcher["environment"]["port"]
+    assert environment.get("command") == launcher["environment"].get("command")
+    assert environment.get("args") == launcher["environment"].get("args")
+    assert environment["is_archived"] == launcher["environment"]["is_archived"]
```
Panaetius, line 1376:
```suggestion
    assert environment["id"] != launcher["environment"]["id"]
    assert environment["creation_date"] != launcher["environment"]["creation_date"]
    assert environment == snapshot(exclude=props("id","creation_date"))
```

you could use the snapshot fixture here if you want. makes it a bit simpler. Downside is that if it returns bad values when you snapshot it, it will check against the bad values in the future. the plus side is, if we add something in the future that isn't currently covered by any of the assert, the snapshot would catch that and fail.
So maybe use it in addition to the assert if you're worried about snapshotting bad data?

Panaetius (approved):

PR #703, opened 2025-03-05
test: make schemathesis tests less flaky

### LINKED ISSUE #696: Flaky schemathesis test when listing resource pools
Flaky schemathesis tests on GET `api/data/resource_pools`

```
schemathesis.exceptions.CheckFailed: 
E       
E       1. Accepted negative data
E       
E           Allowed statuses: 400, 401, 403, 404, 422, 428, 5xx
E       
E       [200] OK:
E       
E           `[]`
E       
E       Reproduce with: 
E       
E           curl -X GET -H 'host: mockserver:1234' -H 'authorization: ***"is_admin": true, "id": "admin", "name": "Admin Doe", "first_name": "Admin", "last_name": "Doe", "email": "admin.doe@gmail.com", "full_name": "Admin Doe"}' http://localhost/api/data/resource_pools
E       
E       Falsifying example: async_run(
E           sanic_client=<test.utils.SanicReusableASGITestClient object at 0x7fd4d9ab6360>,
E           admin_headers={'Authorization': '***"is_admin": true, "id": "admin", "name": "Admin Doe", "first_name": "Admin", "last_name": "Doe", "email": "admin.doe@gmail.com", "full_name": "Admin Doe"}'},
E           requests_statistics=RequestsStatistics(len=22),
E           case=,
E       )
```

Panaetius:
since the issue is that sanic uses https://docs.python.org/3/library/urllib.parse.html#urllib.parse.parse_qs with `keep_empty=False`, the easiest way to catch all cases where this is an issue is to just use httpx to encode the url and then decode it again and see if any key would be removed by this, and then ignore those test cases.

~Note: there's an additional flaky test where calling a deletion endpoint with something non-existing passes with a 204, but schemathesis expects that to fail. Schemathesis has a `negative_data_rejection.accepted-statuses` setting, but this can only be set when using the schemathesis CLI runner, not when doing openapi tests from within pytest. It can also not really be monkeypatched as it comes from a dataclass and monkeypatching dataclasses is a pain... see schemathesis/schemathesis/issues/2755~ I've made it so we just ignore `negative_data_rejection` on DELETE with 204 

I did `watch -e pytest -m schemathesis` for a long while and didn't have any failures.

closes #696


leafty (approved):
LGTM Thanks 👏

PR #705, opened 2025-03-05
fix: Ensure an existing solr core

eikek:
- Allows to create and get the status of a solr core

/deploy

### DIFF
--- a/components/renku_data_services/solr/solr_client.py
+++ b/components/renku_data_services/solr/solr_client.py
@@ -350,7 +350,7 @@ class DefaultSolrClient(SolrClient):
     async def get(self, id: str) -> QueryResponse:
         """Get a document by id, returning a `QueryResponse`."""
         resp = await self.get_raw(id)
-        if resp.status_code != 200:
+        if not resp.is_success:
             raise SolrClientGetByIdException(id, resp)
         else:
             return QueryResponse.model_validate(resp.json())
@@ -358,7 +358,7 @@ class DefaultSolrClient(SolrClient):
     async def query(self, query: SolrQuery) -> QueryResponse:
         """Query documents, returning a `QueryResponse`."""
         resp = await self.query_raw(query)
-        if resp.status_code != 200:
+        if not resp.is_success:
             raise SolrClientQueryException(query, resp)
         else:
             return QueryResponse.model_validate(resp.raise_for_status().json())
@@ -426,11 +426,11 @@ class SolrAdminClient(AbstractAsyncContextManager, ABC):
     """
 
     @abstractmethod
-    async def status(self) -> dict[str, Any] | None:
+    async def status(self, core_name: str | None) -> dict[str, Any] | None:
         """Return the status of the connected core."""
         ...
 
-    async def create(self, core_name: str) -> None:
+    async def create(self, core_name: str | None) -> None:
         """Create a core."""
         ...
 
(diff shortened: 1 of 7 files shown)


components/renku_data_services/solr/solr_client.py
```diff
@@ -70,6 +66,13 @@ def __str__(self) -> str:
         )
 
 
+class SolrClientException(Exception, ABC):
```
olevski, line 69: suggestion. We have a "central" parent error for all errors in the data services. I think it would be good to just inherit from that here. It is in `errors.errors`. That error is very similar and it has a message field. It is also nicely handled in our centralized error handling.

components/renku_data_services/solr/solr_client.py
```diff
@@ -364,3 +413,71 @@ async def delete(self, query: str) -> Response:
...
+
+    Url: https://solr.apache.org/guide/solr/latest/configuration-guide/coreadmin-api.html
+    """
+
+    @abstractmethod
+    async def status(self, core_name: str | None) -> dict[str, Any] | None:
+        """Return the status of the connected core."""
+        ...
+
+    async def create(self, core_name: str | None) -> None:
```
olevski, line 429: question: This is not `@abstractmethod` on purpose?

components/renku_data_services/solr/solr_client.py
```diff
@@ -364,3 +413,71 @@ async def delete(self, query: str) -> Response:
...
+        return await self.delegate.__aexit__(exc_type, exc, tb)
+
+    async def status(self, core_name: str | None) -> dict[str, Any] | None:
+        """Return the status of the connected core or the one given by `core_name`."""
+        core = core_name or self.config.core
+        resp = await self.delegate.get(f"/{core}")
+        if not resp.is_success:
+            raise SolrClientStatusException(self.config, resp)
+        else:
+            data = resp.raise_for_status().json()["status"][self.config.core]
```
olevski, line 467: question: Do you want to use the new custom exceptions here that you defined above?

components/renku_data_services/solr/solr_client.py
```diff
@@ -364,3 +413,71 @@ async def delete(self, query: str) -> Response:
...
+
+    async def create(self, core_name: str | None) -> None:
+        """Create a core with the given `core_name` or the name provided in the config object."""
+        core = core_name or self.config.core
+        data = {"create": {"name": core, "configSet": "_default"}}
+        resp = await self.delegate.post(
+            "",
+            content=json.dumps(data).encode("utf-8"),
+            headers={"Content-Type": "application/json"},
+        )
```
olevski, line 479: The http client can do the json serialization and header setting for you.

```suggestion
        resp = await self.delegate.post(
            "",
            json=data,
        )
```

components/renku_data_services/solr/solr_migrate.py
```diff
@@ -132,13 +136,36 @@ class MigrationState(BaseModel):
...
         self.__config = cfg
         self.__docId: str = "VERSION_ID_EB779C6B-1D96-47CB-B304-BECF15E4A607"
 
+    async def ensure_core(self) -> None:
+        """Ensures an existing core.
+
+        If no core is found, one is created using the admin api.
+        """
+        async with DefaultSolrAdminClient(self.__config) as client:
+            status = await client.status(None)
```
olevski, line 159: nitpick: It may "sound" a bit better if the method is called `core_status` instead of just status. But dont worry this is fine. Feel free to ignore this.

olevski (changes requested):
Thanks Eike. Just a few minor comments.

components/renku_data_services/solr/solr_client.py
eikek: Ah no thanks, just forgot (not used to marking it like this… 😅)

components/renku_data_services/solr/solr_client.py
eikek: Thank you!

eikek:
Sorry @olevski I forgot about that one… I made all the changes, thanks four 👀

components/renku_data_services/solr/solr_client.py
eikek: Ah no thanks, just forgot (not used to marking it like this… 😅)

olevski:
@eikek I think we can merge this without needing to merge the renku PR. Right? I konw if we do not merge the renku pr then this wont fully  run. But then we dont have to wait for the renku pr. So I will just edit the PR description to see and if everything is green then I will merge this.

olevski:
@eikek as soon as the `0.67.0` release on renku is tagged then merge this.

olevski:
@eikek renku `0.67.0` has been released and tagged. So this can now be merged.

PR #706, opened 2025-03-05
Build/external resources

sambuc:
feat: add multi-kubernetes-cluster client
* Add `KubeClusterSettings` and a cluster field of that type to
  ResourcePool

* Explicitly pass the ApiClient parameters in k8s/clients.py and
  simplify`__init__(...)` functions

* Refactor k8s_client.py
  * Made sure we had a single sanitize function pointer, no need for
    multiple ones.
  * Introduce `K8sClientProto[_SessionType, _Kr8sType](ABC)` as the public
    interface for the Kubernetes cluster wrapper
  * Added `DummyK8sClient` for tests
  * Added `MultipleK8sClient` which manages multiple Kubernetes API
    connections.
  * Prefixed private items (classes, variables & methods) with `_` to
    be more explicit.

* Added the cluster name as a parameter in the hashed string used for
  unique references of user sessions.
  
  /deploy

### DIFF
--- a/components/renku_data_services/crc/db.py
+++ b/components/renku_data_services/crc/db.py
@@ -183,6 +183,43 @@ class ResourcePoolRepository(_Base):
                 output.append(rp.dump(quota))
             return output
 
+    async def get_resource_pool_from_class(
+        self, api_user: base_models.APIUser, resource_class_id: int
+    ) -> models.ResourcePool:
+        """Get the resource pool the class belongs to."""
+        async with self.session_maker() as session:
+            stmt = (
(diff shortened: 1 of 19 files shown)


sambuc:
Ho, thank you @leafty, I was wondering why nothing was happening.

lorenzo-cavazzi:
@sambuc  the `release-0.65.0` branch has been merged into master and deleted. You should remove it or the action will fail

sambuc:
@lorenzo-cavazzi thank you for the information!

I was rebasing and solving the conflicts but though I would let it run before pushing the fixes.

PR #710, opened 2025-03-06
feat: support private container images in sessions

andre-code:
This PR enables launching sessions with private container images by dynamically creating and injecting GitLab image pull secrets. If the user is authenticated and has a valid GitLab access token, a secret (<server_name>-image-secret) is generated and attached when needed.

/deploy renku=master  renku-ui=main


leafty (commented):
Will this approach work with pause/resume when the initial token has expired?

andre-code:
> Will this approach work with pause/resume when the initial token has expired?

Yes, it patches the session with an updated GitLab token, so it should work after pausing and resuming the session.

components/renku_data_services/notebooks/blueprints.py
```diff
@@ -370,10 +373,28 @@ async def _handler(
...
+                needs_pull_secret = await requires_image_pull_secret(self.nb_config, image, internal_gitlab_user)
+
+                if needs_pull_secret:
+                    image_pull_secret_name = f"{server_name}-image-secret"
+
+                    image_secret = get_gitlab_image_pull_secret(
+                        self.nb_config, user, image_pull_secret_name, internal_gitlab_user.access_token
+                    )
+                    if image_secret:
+                        secrets_to_create.append(image_secret)
```
leafty, line 388: `get_gitlab_image_pull_secret()` always returns a secret:

```suggestion
                    secrets_to_create.append(image_secret)
```

leafty (commented):
Code looks good, one small comment below.

Very important: a new issue needs to be opened:

*Users without access to the private image can still launch the session from swagger and it can be started if the host node has the image in its local registry*
=> The pull policy needs to be updated to `Always`

olevski:
Here is one part of the support for `ImagePullPolicy`. In amalthea:
https://github.com/SwissDataScienceCenter/amalthea/pull/905

We should open another PR to enable the use of that :point_up_2:  in data services. And make sure we release both this and amalthea at once.

components/renku_data_services/notebooks/core_sessions.py
```diff
@@ -456,6 +508,27 @@ async def patch_session(
     if extra_containers:
         patch.spec.extraContainers = extra_containers
 
+    if isinstance(user, AuthenticatedAPIUser) and internal_gitlab_user.access_token is not None:
+        image = session.spec.session.image
+        server_name = session.metadata.name
+        needs_pull_secret = await requires_image_pull_secret(nb_config, image, internal_gitlab_user)
+
+        if needs_pull_secret and session.spec.imagePullSecrets is not None:
```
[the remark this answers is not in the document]
andre-code: Sure, I just wanted to make sure we replace it when a pull image exists, but if we are adding it regardless, I'll remove that condition and only check if the pull secret is needed

components/renku_data_services/notebooks/blueprints.py
andre-code: sure, I have removed this conditional

olevski (dismissed):

olevski:
So I think it is best to merge #742 into here and then merge this into main and do a release along with the ui.

@leafty and/or @andre-code  you mind reviewing #742?

olevski (approved):

Issue #712, opened 2025-03-07
Improve error handling for git URLs when adding a new code repo


lorenzo-cavazzi:
The endpoint should:
 
- Validate the URL
- Resolve to the correct git URL
- Prevent exact duplicates

More info here: https://www.notion.so/renku/Validate-git-URL-when-adding-a-code-repo-1ac0df2efafc80709d31c7bbbdfb6896

leafty:
We can close this and open a new issue about resolving user input into the canonical URL of a repo.

E.g. if the user gives us `https://github.com/SwissDataScience/renku`, this will work but will not be saved as `https://github.com/SwissDataScience/renku.git` at the moment. This means users can provide duplicate repos to a project if they add both URLs.

PR #714, opened 2025-03-07
fix: validate git URLs

leafty:
See: #712.

What is done with this change:
* Validate the repository URL (when the project is created or updated): the repository URL must be an HTTP or HTTPS URL. SSH is not supported by RenkuLab at the moment.
* Remove exact duplicates.

What is not included with this change:
* Resolve the user-provided URL to a canonical URL (e.g. https://github.com/SwissDataScienceCenter/renku-data-services -> https://github.com/SwissDataScienceCenter/renku-data-services.git). This would require connecting validation code with connected services and is left for a future change.

/deploy #notest renku=release-0.66.0

### DIFF
--- a/components/renku_data_services/project/core.py
+++ b/components/renku_data_services/project/core.py
@@ -204,7 +204,7 @@ def _validate_repository(repository: str) -> str:
     """Validate a git repository."""
     stripped = repository.strip()
     parsed = urlparse(stripped)
-    if parsed.scheme != "http" and parsed.scheme != "https":
+    if parsed.scheme not in ["http", "https"]:
         raise errors.ValidationError(message=f'The repository URL "{repository}" is not a valid HTTP or HTTPS URL.')
     return stripped
 
--- a/components/renku_data_services/project/core.py
+++ b/components/renku_data_services/project/core.py
@@ -205,7 +205,7 @@ def _validate_repository(repository: str) -> str:
     stripped = repository.strip()
     parsed = urlparse(stripped)
     if parsed.scheme != "http" and parsed.scheme != "https":
-        raise errors.ValidationError(message=f"The repository URL {repository} is not a valid HTTP or HTTPS URL.")
+        raise errors.ValidationError(message=f'The repository URL "{repository}" is not a valid HTTP or HTTPS URL.')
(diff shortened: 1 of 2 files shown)


components/renku_data_services/project/core.py
```diff
@@ -182,6 +186,29 @@ def validate_session_secrets_patch(
...
+            without_duplicates.append(repo)
+            seen.add(repo)
+    return without_duplicates
+
+
+def _validate_repository(repository: str) -> str:
+    """Validate a git repository."""
+    stripped = repository.strip()
+    parsed = urlparse(stripped)
+    if parsed.scheme != "http" and parsed.scheme != "https":
```
Panaetius, line 207: suggestion(feel free to ignore):
```suggestion
    if parsed.scheme not in ["http", "https"]:
```

test/bases/renku_data_services/data_api/test_projects.py
```diff
@@ -225,6 +225,42 @@ async def test_project_creation_with_conflicting_slug(sanic_client, user_headers
...
 
+@pytest.mark.asyncio
+async def test_project_creation_with_duplicate_repositories(sanic_client, user_headers, regular_user) -> None:
+    namespace = regular_user.namespace.slug
+    payload = {
+        "name": "My Project",
+        "namespace": namespace,
+        "repositories": [
+            "https://github.com/SwissDataScienceCenter/renku-data-services.git",
+            "https://github.com/SwissDataScienceCenter/renku-data-services.git",
```
Panaetius, line 236: question: this would probably fail if it was
```suggestion
            "http://github.com/SwissDataScienceCenter/renku-data-services.git",
            "https://github.com/SwissDataScienceCenter/renku-data-services.git",
```
but I'm not sure if we want to handle this case. In a way it's not a duplicate, but I also can't think of a reason why a user would want both http and https for the same repo, so preventing an edge case for them could be nice

leafty: I think it's something that can be handled once we try to re-write repo URLs to use the canonical URL.

Panaetius (approved):

PR #720, opened 2025-03-11
fix: Kubernetes client tests

sambuc:
* Using the DummyK8sClient we don't need an actual K8s cluster to check our api calls end up where expected.
   * Implemented the strict minimum to pass the current tests
   * Everything else raises an exception.

 * Removed code in the test

 * Fixed warnings
   * Fixed missing call to `super().__init__()`
   * Removed unused fixtures
   * Removed dependency on `ClusterRequired`

### DIFF
--- a/components/renku_data_services/notebooks/api/classes/k8s_client.py
+++ b/components/renku_data_services/notebooks/api/classes/k8s_client.py
@@ -2,6 +2,7 @@
 
 import asyncio
 import base64
+import datetime
 import glob
 import json
 import os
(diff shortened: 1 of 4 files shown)


components/renku_data_services/notebooks/api/classes/k8s_client.py
```diff
@@ -945,47 +946,80 @@ async def delete_secret(self, name: str) -> None:
...
     ) -> _SessionType:
-        """Patch the user session."""
-        raise NotImplementedError()
+        """Pretend to Patch the user session."""
+        server = await self.get_server(server_name, safe_username)
+        if not server:
+            raise errors.MissingResourceError(
+                message=f"Cannot find server {server_name} for user {safe_username} in order to patch it."
+            )
+        return server
```
Panaetius, line 1012: you could probably actually do patching using `deepmerge`

sambuc: Thanks for the review, I'll keep this in mind as I extend the tests.

I also have to think hard if I don't want to change at which level I implement a dummy class for testing. With the current implementation we can't create unit tests for most of the code within this file, which I think is a problem. 

For now I am going to proceed forward as-is and I will probably rework this quite a bit as soon as I have a demo done.

PR #722, opened 2025-03-11
Fix 2 issues for switching off notebooks service

olevski:
These 2 changes we have made or fixed in the notebook service. But were wrong  in the data service.

In the PR to fully turn off the notebook service the problems showed up.

closes https://github.com/SwissDataScienceCenter/renku-notebooks/issues/2016

### DIFF
--- a/components/renku_data_services/notebooks/config/dynamic.py
+++ b/components/renku_data_services/notebooks/config/dynamic.py
@@ -206,7 +206,7 @@ class _CustomCaCertsConfig:
     def from_env(cls) -> Self:
         return cls(
             image=os.environ.get("NB_SESSIONS__CA_CERTS__IMAGE", "renku/certificates:0.0.2"),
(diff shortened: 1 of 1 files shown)


olevski:
The git clone mount path in the notebooks is different (and that works with saving session data). See here: https://github.com/SwissDataScienceCenter/renku-notebooks/blob/master/renku_notebooks/api/amalthea_patches/init_containers.py#L30

leafty (commented):
Should this be deployed and tested?

olevski:
> Should this be deployed and tested?

The errors are pretty obvious. But yeah I can deploy it in my PR where I turn off the renku notebooks. I will post a link and confirm.

olevski:
@leafty I tested this over at https://github.com/SwissDataScienceCenter/renku/pull/3861 and saving unsaved works from the ui works as expected with these fixes.

PR #734, opened 2025-03-18
fix: make project slug regex not have infinite recursion

### LINKED ISSUE #694: Pinned project validation fails when getting user preferences
```
String should match pattern '^[a-zA-Z0-9]([_.\-/]?[a-zA-Z0-9]+)*[a-zA-Z0-9]$' [type=string_pattern_mismatch, input_value='bethcg/renku-teaching-assignment-1', input_type=str]
```

From: https://sentry.dev.renku.ch/organizations/sentry/issues/9428/events/0862a32f453d47aca14b89161788112b/

The code where this happens is:
```
/app/env/lib/python3.12/site-packages/renku_data_services/users/blueprints.py in _get at line 245

        async def _get(_: Request, user: base_models.APIUser) -> JSONResponse:
            user_preferences: models.UserPreferences
            user_preferences = await self.user_preferences_repo.get_user_preferences(requested_by=user)
            return validated_json(apispec.UserPreferences, user_preferences)
```

Panaetius:
closes #694


olevski (dismissed):

olevski:
I think we have an extra escape in there...
Screenshot From 2025-03-21 11-54-10

components/renku_data_services/users/api.spec.yaml
```diff
@@ -531,7 +531,7 @@ components:
       example: "user/my-project"
       # limitations based on allowed characters in project slugs from Gitlab from here:
       # https://docs.gitlab.com/ee/user/reserved_names.html
-      pattern: "^[a-zA-Z0-9]([_.\\-/]?[a-zA-Z0-9]+)*[a-zA-Z0-9]$"
+      pattern: "^[a-zA-Z0-9]+([_.\\-/][a-zA-Z0-9]+)*[_.\\-/]?[a-zA-Z0-9]$"
```
olevski, line 534:
```suggestion
      pattern: "^[a-zA-Z0-9]+([_.\-/][a-zA-Z0-9]+)*[_.\-/]?[a-zA-Z0-9]$"
```

olevski: I tested what we had in the python repl and it did not complain and it seems to have worked just fine.

```
>>> re.match("^[a-zA-Z0-9]([_.\\-/]?[a-zA-Z0-9]+)*[a-zA-Z0-9]$", "test\fdf/fdsf")
>>> re.match("^[a-zA-Z0-9]([_.\\-/]?[a-zA-Z0-9]+)*[a-zA-Z0-9]$", "test/fdsf")
<re.Match object; span=(0, 9), match='test/fdsf'>
```

So it seems like we got lucky before. But we may as well fix it up now.

components/renku_data_services/users/apispec.py
```diff
@@ -57,7 +57,7 @@ class AddPinnedProject(BaseAPISpec):
         description="The slug used to identify a project",
         example="user/my-project",
         min_length=3,
-        pattern="^[a-zA-Z0-9]([_.\\-/]?[a-zA-Z0-9]+)*[a-zA-Z0-9]$",
+        pattern="^[a-zA-Z0-9]+([_.\\-/][a-zA-Z0-9]+)*[_.\\-/]?[a-zA-Z0-9]$",
```
olevski, line 60:
```suggestion
        pattern="^[a-zA-Z0-9]+([_.\-/][a-zA-Z0-9]+)*[_.\-/]?[a-zA-Z0-9]$",
```

olevski (changes requested):

components/renku_data_services/users/api.spec.yaml
olevski: Disregard this. The extra `\` is to escape the `\` in a regular python string. I got confused by this - not sure what I was thinking.

olevski (dismissed):
Sorry I got tripped up on the escape of `\` in the regular python double quoted string.

olevski (approved):

PR #735, opened 2025-03-19
feat: endpoint for listing project as namespace

olevski:
/deploy renku-ui=feat-data-connector-listing


components/renku_data_services/namespace/blueprints.py
```diff
@@ -236,3 +236,29 @@ async def _get_namespace(_: Request, user: base_models.APIUser, slug: Slug) -> J
             )
 
         return "/namespaces/<slug:renku_slug>", ["GET"], _get_namespace
+
+    def get_namespace_second_level(self) -> BlueprintFactoryResponse:
+        """Get user or group namespace by slug."""
```
Panaetius, line 241: issue: This docstring is wrong

components/renku_data_services/namespace/blueprints.py
```diff
@@ -236,3 +236,29 @@ async def _get_namespace(_: Request, user: base_models.APIUser, slug: Slug) -> J
...
+
+        @authenticate(self.authenticator)
+        async def _get_namespace_second_level(
+            _: Request, user: base_models.APIUser, first_slug: Slug, second_slug: Slug
+        ) -> JSONResponse:
+            path = ProjectPath.from_strings(first_slug.value, second_slug.value)
```
Panaetius, line 247: question: can this only be a project? it could be a data connector, no?

components/renku_data_services/project/orm.py
```diff
@@ -48,12 +48,15 @@ class ProjectORM(BaseORM):
...
     slug: Mapped["EntitySlugORM"] = relationship(
-        # NOTE: joined lazy relationship can cause duplicates to be returned
-        lazy="selectin",
+        lazy="joined",
         init=False,
         repr=False,
         viewonly=True,
         back_populates="project",
+        # NOTE: If the data_connector ID is not null below then multiple joins are possible here
+        # since an entity slug for data connector owned by a project will and an entity slug for a project
```
Panaetius, line 57: you accidentally a word
```suggestion
        # since an entity slug for data connector owned by a project and an entity slug for a project
```

At least I think this is right this way

Panaetius (changes requested):
nice work combining the two db repos, we are starting to go a bit overboard with them :slightly_smiling_face:

olevski:
> nice work combining the two db repos, we are starting to go a bit overboard with them 🙂

I am glad you agree. That was driving me a bit crazy tbh for a while now.

Panaetius:
there should be like one repo per module, and I think modules can be cleaned up a bit as well.
like I already did that with users and user_preferences (though they're still separate repors, but at least in the same module).  `crc` and `sessions` are kinda the same thing and could be combined. 

Project has quite a few (ProjectRepo, ProjectMemberRepo,ProjectSessionSecretRepo, ProjectMigrationRepo.

There's still the DataConnectorSecretRepo which I don't know if that needs to be separate?

And in storage there's an abstract BaseStorageRepo that is only inherited once in STorageRepository with one method, and the abstract base isn't depended on anywhere.

components/renku_data_services/namespace/blueprints.py
olevski: It can only be a project. Because "namespace" or "owner" is something that can hold other things. So only groups, users or projects are namespaces. And we only have 1 namespace that is at this second level. That is `user/project` or `group/project`. We do have paths like `group/dc` but those cannot be namespaces or owners - i.e. you cannot put other things inside `group/dc`. And `get_namespace_by_path` used in the handler can only return group, user or project namespaces. It will filter out anything with data connectors.

olevski:
I will fix the acceptance tests in a followup PR btw Ralf. I think they are broken because Jena or some other v1 part is acting up in this CI deployment.

PR #751, opened 2025-03-27
feat: product metrics build

Panaetius:
/deploy #notest renku=data-svc-posthog extra-values=posthog.enabled=true,posthog.apiKey=phc_9KXbWU5fPFWJeBQa6n53deAJJKfjqedLJHhMohutDS1,posthog.host=https://eu.i.posthog.com,posthog.environment=development/


components/renku_data_services/k8s_watcher/core.py
```diff
@@ -145,10 +151,65 @@ async def stop(self, timeout: timedelta = timedelta(seconds=10)) -> None:
...
-def k8s_object_handler(cache: K8sDbCache) -> EventHandler:
+async def collect_metrics(
+    previous_obj: K8sObject | None,
+    new_obj: APIObjectInCluster,
+    user_id: str,
+    metrics: MetricsService,
+    rp_repo: ResourcePoolRepository,
+) -> None:
+    """Track product metrics."""
+    user = APIUser(id=user_id)
```
olevski, line 162: suggestion(non-blocking): I think that the event type from the wathc would be useful here. But if we do this we should implement it in a follow up PR. If it is even possible.

olevski (changes requested):
Thanks Ralf this looks really good. Just a few minor suggestion. And if you want we can merge this as  is and do everything in a followup PR - not just the one comment where I specifically mentioned it.

components/renku_data_services/app_config/config.py
```diff
@@ -133,6 +135,20 @@ def from_env(cls, prefix: str = "") -> "SentryConfig":
...
+    """Configuration for posthog."""
+
+    enabled: bool
+
+    @classmethod
+    def from_env(cls, prefix: str = "") -> "PosthogConfig":
+        """Create posthog config from environment variables."""
+        enabled = os.environ.get(f"{prefix}POSTHOG_ENABLED", "false").lower() == "true"
+
+        return cls(enabled)
```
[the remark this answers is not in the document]
Panaetius: before we needed that, because that's what the posthog client uses for sending messages. Before we were sending messages in the background job in the main data services, but since sending now happens in the data_tasks, this value is only needed there. 

The main service just needs to know if metrics are enabled to put them in the staging table in postgres.

components/renku_data_services/k8s_watcher/core.py
Panaetius: I've added it to use DELETED for session stopped. Not sure it's useful for the other metrics

olevski (approved):

PR #754, opened 2025-04-01
test: force temporary db name to start with a letter

lorenzo-cavazzi:
This PR aims to force the temporary database names to start with a letter rather than a number. Please suggest a more reasonable solution if the solution here isn't acceptable.

The reason behind this change is the following: most Postgres clients use a syntax like the following for interacting with databases:

```
DROP DATABASE 01jqedjvej25v1c50jfba9m7w1_master
```

Any database name starting with a number is interpreted as the beginning of a numeric literal _in Postgres_ (citation needed) and requires proper escaping. Sadly, in all the clients I tried, that doesn't happen and I get some variation of the following error when trying to drop temporary DBs

> ERROR:  trailing junk after numeric literal at or near "01jqedjvej25v1c50jfba9m7w1_master"
LINE 1: drop database 01jqedjvej25v1c50jfba9m7w1_master

For convenience, it would be great to pick a naming scheme that avoids this very unique corner case

### DIFF
--- a/components/renku_data_services/base_api/error_handler.py
+++ b/components/renku_data_services/base_api/error_handler.py
@@ -29,7 +29,7 @@ class BaseError(Protocol):
 
 
 class BaseErrorResponse(Protocol):
-    """Porotocol for the error response class of an apispec module."""
+    """Protocol for the error response class of an apispec module."""
 
     error: BaseError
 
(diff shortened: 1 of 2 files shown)


olevski (approved):
Thanks Lorenzo that looks good. I did not know this was a problem. Somehow I have never come across it.

olevski:
@lorenzo-cavazzi this PR fixes the devcontainer build: https://github.com/SwissDataScienceCenter/renku-data-services/pull/755

Sorry this recently cropped up.

PR #756, opened 2025-04-01
feat: add option to strip prefix

olevski:
This adds the option on the environment to strip the URL prefix when requests reach the session.

This is needed for things like rstudio and other similar frontends that cannot run on a url path other than `/`.

/deploy


components/renku_data_services/notebooks/crs.py
```diff
@@ -26,12 +29,16 @@
     ImagePullSecret,
     Ingress,
     InitContainer,
+    Limits6,
+    Limits7,
```
mohammad-alisafaee, line 33: These aren't used anywhere. Do they do something specific when imported.

mohammad-alisafaee: Also, they are reimported below as `Limits` and `LimitsStr`

components/renku_data_services/notebooks/crs.py
```diff
@@ -26,12 +29,16 @@
...
     MatchExpression,
     NodeAffinity,
     NodeSelectorTerm,
     Preference,
     PreferredDuringSchedulingIgnoredDuringExecutionItem,
     ReconcileStrategy,
+    Requests6,
+    Requests7,
```
mohammad-alisafaee, line 41: Same as above, they are reimported below as `Requests` and `RequestsStr`

components/renku_data_services/session/orm.py
```diff
@@ -80,6 +80,7 @@ class EnvironmentORM(BaseORM):
         default=None,
     )
     build_parameters: Mapped["BuildParametersORM"] = relationship(lazy="joined", default=None)
+    strip_path_prefix: Mapped[bool] = mapped_column(default=False, server_default=false())
```
mohammad-alisafaee, line 83: Maybe add `nullable=False`

components/renku_data_services/migrations/versions/e117405fed51_add_strip_prefix_option_to_environments.py
```diff
@@ -0,0 +1,32 @@
...
+down_revision = "ca3731b65787"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column(
+        "environments",
+        sa.Column("strip_path_prefix", sa.Boolean(), server_default=sa.text("false"), nullable=False),
```
mohammad-alisafaee, line 23: I'm not sure if `default` is needed here or not. It's defined in the ORM.

olevski: It definitely does not hurt. If things are done directly in the DB and the DB default is not set we may get weird behaviour. This makes it nicer to do DB maintenance or change things as an admin. I think we  will never do this but in case we do it is a nice small thing that makes things easier.

olevski: Also this makes the addition of the new column in the table easy. The default is false for the column so the table can be just added to the table.

mohammad-alisafaee (approved):

Issue #758, opened 2025-04-02
Slug validation breaks the namespace listing endpoint


leafty:
See: https://sentry.dev.renku.ch/organizations/sentry/issues/9644/?query=is%3Aunresolved&referrer=issue-stream

PR #761, opened 2025-04-02
fix: lowercase namespace paths when serializing

olevski:
```
data-service Srv 0 2025-04-02 12:18:03 +0000 ERROR: Exception occurred while handling uri: 'https://renkulab.io/api/data/namespaces?kinds=group&kinds=u
ser&kinds=project&minimum_role=editor'
data-service Traceback (most recent call last):
data-service   File "/app/env/lib/python3.13/site-packages/renku_data_services/base_models/validation.py", line 25, in validate_and_dump
data-service     body = model.model_validate(data).model_dump(exclude_none=exclude_none, mode="json", **kwargs)
data-service            ~~~~~~~~~~~~~~~~~~~~^^^^^^
data-service   File "/app/env/lib/python3.13/site-packages/pydantic/main.py", line 627, in model_validate
data-service     return cls.__pydantic_validator__.validate_python(
data-service            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
data-service         obj, strict=strict, from_attributes=from_attributes, context=context
data-service         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
data-service     )
data-service     ^
data-service pydantic_core._pydantic_core.ValidationError: 4 validation errors for NamespaceResponseList
data-service 6.slug
data-service   String should match pattern '^(?!.*\.git$|.*\.atom$|.*[\-._][\-._].*)[a-z0-9][a-z0-9\-_.]*$' [type=string_pattern_mismatch, input_value=
'test-tasko-PRIVATE-2', input_type=str]
data-service     For further information visit https://errors.pydantic.dev/2.10/v/string_pattern_mismatch
data-service 6.path
data-service   String should match pattern '^(?!.*\.git$|.*\.atom$|.*[\-._][\-._].*)[a-z0-9][a-z0-9\-_.]*(?<!\.git)(?<!\.atom)(?:/[a-z0-9][a-z0-9\-_.]*
){0,1}$' [type=string_pattern_mismatch, input_value='tasko.olevski/test-tasko-PRIVATE-2', input_type=str]
data-service     For further information visit https://errors.pydantic.dev/2.10/v/string_pattern_mismatch
data-service 7.slug
data-service   String should match pattern '^(?!.*\.git$|.*\.atom$|.*[\-._][\-._].*)[a-z0-9][a-z0-9\-_.]*$' [type=string_pattern_mismatch, input_value=
'test-TASKO-PRIVATE-2', input_type=str]
```

This is likely affecting only me. Because only for a brief period we had one of two endpoints that did not have proper lowercase enforcement on slug values. So in production I have some namespaces that have uppercase characters for slugs.

### DIFF
--- a/components/renku_data_services/data_connectors/orm.py
+++ b/components/renku_data_services/data_connectors/orm.py
@@ -92,7 +92,7 @@ class DataConnectorORM(BaseORM):
         return models.DataConnector(
             id=self.id,
             name=self.name,
-            slug=self.slug.slug,
+            slug=self.slug.slug.lower(),
             namespace=self.slug.dump_namespace(),
             visibility=self._dump_visibility(),
             created_by=self.created_by_id,
--- a/test/bases/renku_data_services/data_api/test_projects.py
+++ b/test/bases/renku_data_services/data_api/test_projects.py
@@ -1189,14 +1189,14 @@ async def test_project_slug_case(
     # Get the project
     _, res = await sanic_client.get(f"/api/data/projects/{project_id}", headers=user_headers)
     assert res.status_code == 200
-    assert res.json.get("slug") == uppercase_slug
+    assert res.json.get("slug") == uppercase_slug.lower()
     etag = res.headers["ETag"]
     # Get it by the namespace
     _, res = await sanic_client.get(
         f"/api/data/namespaces/{group['slug']}/projects/{uppercase_slug}", headers=user_headers
     )
     assert res.status_code == 200
-    assert res.json.get("slug") == uppercase_slug
+    assert res.json.get("slug") == uppercase_slug.lower()
     # Patch the project
     new_name = "new-name"
     _, res = await sanic_client.patch(
@@ -1205,7 +1205,7 @@ async def test_project_slug_case(
         headers={"If-Match": etag, **user_headers},
     )
     assert res.status_code == 200
-    assert res.json["slug"] == uppercase_slug
+    assert res.json["slug"] == uppercase_slug.lower()
     assert res.json["name"] == new_name
 
 
@@ -1243,7 +1243,7 @@ async def test_project_copy_basics(
     assert copy_project == snapshot(exclude=props("id", "updated_at", "creation_date", "etag", "template_id"))
 
     _, response = await sanic_client.get(
-        f"/api/data/projects/{copy_project["id"]}", params={"with_documentation": True}, headers=user_headers
+        f"/api/data/projects/{copy_project['id']}", params={"with_documentation": True}, headers=user_headers
     )
     assert response.status_code == 200, response.text
     copy_project = response.json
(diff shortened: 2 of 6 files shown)


olevski:
closes #758

components/renku_data_services/data_connectors/orm.py
```diff
@@ -92,7 +92,7 @@ def dump(self) -> models.DataConnector:
         return models.DataConnector(
             id=self.id,
             name=self.name,
-            slug=self.slug.slug,
+            slug=self.slug.slug.lower(),
```
eikek, line 95: Only a question: it looks like we need to take care to always lower things, when inserting and retrieving. Would it make sense to make sure to always lower on insert, so we can be sure to have valid data in the db? It would require less discipline. I was wondering if we could have a `Mapped[Slug]` instead of `Mapped[str]` for example (I saw we do this for `ULID`) and use its canonical string representation.

eikek (approved):

components/renku_data_services/data_connectors/orm.py
leafty: The validation was added after some upper case slugs were already inserted. One solution could be to push a DB migration which would apply `lower()` on all existing slugs.

test/bases/renku_data_services/data_api/test_projects.py
```diff
@@ -1243,7 +1243,7 @@ async def test_project_copy_basics(
     assert copy_project == snapshot(exclude=props("id", "updated_at", "creation_date", "etag", "template_id"))
 
     _, response = await sanic_client.get(
-        f"/api/data/projects/{copy_project["id"]}", params={"with_documentation": True}, headers=user_headers
+        f"/api/data/projects/{copy_project['id']}", params={"with_documentation": True}, headers=user_headers
```
Panaetius, line 1246: I think your formatter is set up weird, using different quotes in this circumstance isn't necessary anymore in newer python (not that it matters, though)

Panaetius (commented):
question: this PR fixes it so values are always returned lower case. But is there an issue where an endpoint returns a lower-case value, that the client then picks up and queries another endpoint with, but fails with 404 because the other endpoint is querying the db with a now lower-case value but has the upper-case one in the db?

test/bases/renku_data_services/data_api/test_projects.py
olevski: Yeah this is a problem. And I dont want to get into writing a complicated migration right now. Because I cannot just set all slugs to lowercase in the migration. I have to also handle potential conflicts and rename some slugs so that when I make all lowercase I do not get conflicts with duplicates in the DB. So I will leave this for a later PR.

I will close this and reopen another PR where I just change the validation when dumping namespaces. We can do this without validation when we respond.

olevski:
Closed in favor of #766

PR #762, opened 2025-04-03
feat: add user preferences for dismiss project migration

andre-code:
PR to add user preference to dismiss project migration banner

/deploy renku=release-0.68.0


components/renku_data_services/users/apispec.py
```diff
@@ -300,3 +300,6 @@ class UserPreferences(BaseAPISpec):
         pattern="^[A-Za-z0-9]{1}[A-Za-z0-9-]+$",
     )
     pinned_projects: PinnedProjects
+    dismiss_project_migration_banner: Optional[bool] = Field(
+        None, description="Is project migration banned dismissed"
```
Panaetius, line 304:
```suggestion
        None, description="Is project migration banner dismissed"
```

components/renku_data_services/users/db.py
```diff
@@ -411,7 +411,9 @@ async def add_pinned_project(self, requested_by: APIUser, project_slug: str) ->
 
             if user_preferences is None:
                 new_preferences = UserPreferences(
-                    user_id=cast(str, requested_by.id), pinned_projects=PinnedProjects(project_slugs=[project_slug])
+                    user_id=cast(str, requested_by.id),
+                    pinned_projects=PinnedProjects(project_slugs=[project_slug]),
+                    dismiss_project_migration_banner=False,
```
Panaetius, line 416: I think this can just be a default on the class itself so it doesn't need to be set here. Just set it `dismiss_project_migration_banner: bool = False` on the `UserPreferences` class

components/renku_data_services/users/db.py
```diff
@@ -461,3 +463,40 @@ async def remove_pinned_project(self, requested_by: APIUser, project_slug: str)
...
+            )
+            user_preferences_orm = result.one_or_none()
+            if user_preferences_orm is None:
+                new_project_slugs: list[str] = []
+                pinned_projects = PinnedProjects(project_slugs=new_project_slugs).model_dump()
+                user_preferences_orm = UserPreferencesORM(
+                    user_id=cast(str, requested_by.id),
+                    pinned_projects=pinned_projects,
+                    dismiss_project_migration_banner=True,
+                )
```
Panaetius, line 482: `project_slugs` on `PinnedProjects` is allowed to be None in the type, so you could simplify this code a bit by just setting it to None. 

Also it's probably better to use `UserPreferencesORM.load` as in the code for pinned projects, for consistency.

Panaetius (changes requested):
Thank you for the changes, it's almost all good.

One overall suggestion I have would be to not have the flag be a negative, so call is `show_project_migration_banner`, have it default to `True` and set it to `False` when a user disables it. Then we wouldn't end up in double negative logic that is a bit awkward to read.

One other thing that came to mind when reading, though this was the case before your changes, is that the whole user preferences section is not really consistent with how the rest of our APIs work (it's more RPC than REST at the moment).
Ideally we'd just have `/user/preferences` with GET/POST/PATCH and make all preferences changes through that, without dedicated sub-paths for each preference entry. Otherwise the API could quickly get unwieldy with many endpoints when three endpoints would be enough. I'd probably also change it so GET doesn't return a 404 but just returns default values (pinned projects=[], show_banner=True), i.e. we always have preferences for every user, that'd probably make some of the logic on the UI simpler as well. Especially since knowing whether a user has set preferences or not is not not useful to know for us.

leafty:
@Panaetius 
> Ideally we'd just have /user/preferences with GET/POST/PATCH and make all preferences changes through that, without dedicated sub-paths for each preference entry.

I can try to refactor this to be REST-ish; the original API tried to deal with concurrent updates this way, but now we have ETags we can use to reject conflicting updates (though client-side resolution is still not there).

andre-code:
Thanks, @Panaetius , for your suggestions!
I'll definitely include the change to make the default value True, as well as the other suggestion you left in the code. As for simplifying the URLs, that will require some refactoring on the client side as well, so we’ll handle it in a separate PR to avoid blocking the current feature, if that’s okay with you.

andre-code:
Thanks @Panaetius for your review! I’ve applied the suggested changes in this commit: https://github.com/SwissDataScienceCenter/renku-data-services/pull/762/commits/789295906341b7a6a8d2b0eac9fd58b20d0f441b

Panaetius (approved):
thank you!

PR #763, opened 2025-04-03
chore: update the command to generate rclone_schema.autogenerated.json

leafty:
The use of `jq 'sort_by(.Name)'` will keep the entries sorted and keep diffs smaller when we update the schema.

Note that different versions of rclone may produce providers in a different order which is why this PR exists.


olevski (dismissed):
That is nice. Thanks @leafty !

components/renku_data_services/storage/README.md
```diff
@@ -2,4 +2,4 @@
 Contains APIspec generated schemas for cloud storage functionality.
 
 
-To create the RClone schema, run `rclone config providers > rclone_schema.autogenerated.json`.
+To create the RClone schema, run `rclone config providers | jq 'sort_by(.Name)' > rclone_schema.autogenerated.json`.
```
Panaetius, line 5: In my old PR for this, I also added `--indent 4` https://github.com/SwissDataScienceCenter/renku-data-services/pull/520/files#diff-3dde8d2cdc50450cd1f9375f6a003106be43cc320495f19d567f1e387a6216aaR5
It's been a while but I think that was to keep the diff smaller. And `-r` for raw output though I do not remember why I added that.

leafty: `jq` replaces unicode escape sequences with UTF-8 characters without the `--raw-output` option. Since python and data services can handle UTF-8 files, I don't think this would create an issue for us.

PR #770, opened 2025-04-04
feat: support env variables on session launchers

ciyer:
/deploy renku=release-0.68.0

@ciyer  I added the deploy above to make sure that we can merge into main without disrupting other work. I think we should be able to merge  this - but just wanted to check.


components/renku_data_services/session/api.spec.yaml
```diff
@@ -907,6 +913,15 @@ components:
...
       example: "succeeded"
+    EnvVariables:
+      description: Environment variables for the session pod
+      type: object
+      additionalProperties:
+        type: string
+        default: {}
+      example:
+        MY_VAR: value 1
+        VAR2: null
```
olevski, line 924: issue: You should add the validation here - this will give the nicest error message when it is not valid. The code generated from the apispec will produce nice, descriptive error messages that are also handled and returned to the caller. So you would have a schema for the key and a schema for the value. And you can combine the two into an Env variable.

When you do this apply the regex validation only for requests. For the responses we should just return str or str | null.

The same validation should remain in the `EnvVar` dataclass I proposed in the models.py file too. This is because if we relax this developers can easily forget and initialize and sneak in invalid values thorugh some other code path.

.devcontainer/devcontainer.json
```diff
@@ -51,7 +51,9 @@
 				"ms-python.isort",
 				"ms-python.vscode-pylance",
 				"ms-vscode.makefile-tools",
-				"matangover.mypy"
+				"matangover.mypy",
+				"ms-python.mypy-type-checker",
+				"charliermarsh.ruff"
```
olevski, line 56: Feel free to ignore this, but with the addition of ruff I am pretty sure you can remove flake, black and isort above. So if you dont mind you can try and if it works then clean up the extensions.

olevski (changes requested):
Thanks for adding this Sekhar. The code looks good. Lets talk about the patching changes we discussed that are also missing here.

.devcontainer/devcontainer.json
ciyer: Happy to do this. I didn't do it initially because I didn't know if it would disrupt other people's workflows.

components/renku_data_services/session/models.py
```diff
@@ -142,6 +144,50 @@ class EnvironmentPatch:
...
+    if env_variables is None:
+        return None
+
+    # TODO: Verify that these limits are compatible with k8s
+    MAX_NUMBER_ENV_VARIABLES = 32
+    MAX_LENGTH_ENV_VARIABLES_NAME = 256
+    MAX_LENGTH_ENV_VARIABLES_VALUE = 1000
+
+    # see https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_231
+    variable_regex = r"^[a-zA-Z_][a-zA-Z0-9_]*$"
```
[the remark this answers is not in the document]
ciyer: The link was off by a few entries, which I have now fixed. It should point to the "Name" section https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_235

components/renku_data_services/session/models.py
```diff
@@ -142,6 +144,50 @@ class EnvironmentPatch:
...
+        errors.append(f"Cannot have more than {MAX_NUMBER_ENV_VARIABLES} env variables.")
+    has_non_str_value = False
+    for value in env_variables.values():
+        if value is not None and not isinstance(value, str):
+            errors.append("Env variable values must be strings.")
+            has_non_str_value = True
+            break
+    if has_non_str_value:
+        return errors
+    for name, value in env_variables.items():
```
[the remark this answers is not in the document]
ciyer: I changed the type to only allow str values so the first loop is gone.

.devcontainer/devcontainer.json
olevski: Ok yeah up to you. As you suggest though it is probably better to not risk messing with the configuration now and messing up peoples workflows.

test/bases/renku_data_services/data_api/test_sessions.py
```diff
@@ -390,6 +392,27 @@ async def test_get_project_launchers(
...
+def test_env_variable_validation():
+    renku_name_env_variables = {
+        "RENKU_KEY_NUMBER_1": "a value",
+        "RENKULAB_THING": "another value",
+    }
+    try:
+        EnvVar.from_dict(renku_name_env_variables)
+        assert False  # should not reach this point
+    except errors.ValidationError as e:
+        assert e.message == "Env variable name 'RENKU_KEY_NUMBER_1' should not start with 'RENKU'."
```
olevski, line 404: Pytest can help you handle expected assertions without using try/except. Sorry I missed this on the first round.

https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions

olevski (approved):
Just one small change on the tests. You can do it in a follow up PR if you prefer.

test/bases/renku_data_services/data_api/test_sessions.py
ciyer: Fixed.

test/bases/renku_data_services/data_api/test_sessions.py
```diff
@@ -390,6 +392,27 @@ async def test_get_project_launchers(
...
+
+    non_posix_name_env_variables = {
```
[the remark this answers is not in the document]
ciyer: Fixed.

ciyer:
> Just one small change on the tests. You can do it in a follow up PR if you prefer.

The main branch had moved, so I had to rebase the PR anyway; I went ahead and made the suggested changes to the tests.

olevski (approved):

PR #772, opened 2025-04-04
fix: parsing k8s time durations with high values

olevski:
Fixes this issue from sentry:
https://sentry.dev.renku.ch/organizations/sentry/issues/9624/?environment=renkulab&project=13&referrer=project-issue-stream


Panaetius (dismissed):

leafty (changes requested):
Opinion: this should be fixed at the API level, not here. The duration returned by any k8s API should conform to `GEP-2257: Gateway API Duration Format`.
See: https://gateway-api.sigs.k8s.io/geps/gep-2257/
Also, see the `parse_duration()` implementation for reference: https://github.com/kubernetes-client/python/blob/master/kubernetes/utils/duration.py

olevski:
@leafty yeah I can probably use the reverse of parse_duration that exists in kubernetes/utils.py to fix this. I will take a look.

olevski:
@leafty I added the proper conversion from datetime to duration here. I will keep the `safe_parse_duration` just so we can handle the already existing improperly formatted sessions we probably have in prod.

I will also open a PR in amalthea so that the validation is fully/properly done.

olevski:
Here is the bug on the amalthea side: https://github.com/SwissDataScienceCenter/amalthea/issues/918

components/renku_data_services/notebooks/crs.py
```diff
@@ -304,3 +305,27 @@ class AmaltheaSessionV1Alpha1Patch(BaseCRD):
     def to_rfc7386(self) -> dict[str, Any]:
         """Generate the patch to be applied to the session."""
         return self.model_dump(exclude_none=True)
+
+
+def safe_parse_duration(val: Any) -> timedelta:
```
leafty, line 310: Note: I wonder if we should instead mirror the parseDuration code from golang. This is most likely the safest option when reading durations which do not conform to gep-2257.

leafty (approved):
LGTM

Panaetius (approved):

PR #779, opened 2025-04-09
fix: Start reprovisioning search index if a migration requires it

eikek:
If a solr migration requires a reindex, a background task is started on startup that reprovisions the search index.

/deploy

### DIFF
--- a/components/renku_data_services/search/reprovision.py
+++ b/components/renku_data_services/search/reprovision.py
@@ -1,5 +1,6 @@
 """Code for reprovisioning the search index."""
 
+from collections.abc import AsyncGenerator, Callable
 from datetime import datetime
 
 from sanic.log import logger
@@ -8,10 +9,13 @@ from renku_data_services.base_models.core import APIUser
 from renku_data_services.message_queue.db import ReprovisioningRepository
 from renku_data_services.message_queue.models import Reprovisioning
 from renku_data_services.namespace.db import GroupRepository
+from renku_data_services.namespace.models import Group
 from renku_data_services.project.db import ProjectRepository
+from renku_data_services.project.models import Project
 from renku_data_services.search.db import SearchUpdatesRepo
 from renku_data_services.solr.solr_client import DefaultSolrClient, SolrClientConfig
 from renku_data_services.users.db import UserRepo
+from renku_data_services.users.models import UserInfo
 
 
 class SearchReprovision:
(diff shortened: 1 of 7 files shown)


components/renku_data_services/search/reprovision.py
```diff
@@ -0,0 +1,101 @@
...
+        self._user_repo = user_repo
+        self._group_repo = group_repo
+        self._project_repo = project_repo
+
+    async def run_reprovision(self, requested_by: APIUser) -> None:
+        """Start a reprovisioning if not already running."""
+        reprovision = await self._reprovisioning_repo.start()
+        await self.init_reprovision(requested_by, reprovision)
+
+    async def acquire_reprovsion(self) -> Reprovisioning:
```
Panaetius, line 41:
```suggestion
    async def acquire_reprovision(self) -> Reprovisioning:
```

components/renku_data_services/search/reprovision.py
```diff
@@ -0,0 +1,101 @@
...
+            async for project_entity in all_projects:
+                await self._search_updates_repo.insert(project_entity, started)
+                counter += 1
+                log_counter(counter)
+            logger.info("Done adding project entities to search_updates table.")
+
+            logger.info(f"Inserted {counter} entities into the staging table.")
+
+        except Exception as e:
+            logger.error("Error while reprovisioning entities!", exc_info=e)
```
Panaetius, line 98: question: should the error handling be in the individual `async for` blocks, so if one entry fails it will still continue with the next one? as is, any error would stop the whole reprovisioning, right?

eikek: Yes, that is true and my comment is still there and was not addressed :-) I think, too, it is better to not fail the whole reprovisioning but skip failing entries.

eikek: @Panaetius How can I catch an exception that is thrown when async iterating over the list of entities (as in `async for user_entity in …`)? If I wrap the entire `async for` statement, it only catches the first exception, but I then can't proceed with the next element right?

Panaetius: I don't think you can from outside the generator. And inside the generator you'd probably have to change it so it returns an Either monad style thing with the error and the deal with that in the async for loop

eikek: Ok thanks 👍🏼, then I tried to be as close as possible wrapping what I can in a try-catch.

PR #783, opened 2025-04-09
fix: Updating search documents without optimistic concurrency control

eikek:
Fixes updates of search entries to always overwrite existing documents.

/deploy


olevski (approved):
Thanks Eike

PR #784, opened 2025-04-09
fix: copy dc links when copying projects

### LINKED ISSUE #781: Copying a public project with a public data connector (owned by a user) fails to link the data connector
In renku 0.67.1

To reproduce:
- Create a public project
- Make it a template
- Link a data connector which is public but owned by a user into this project
- Login as a different user - try to copy the project created above, the copy will be created but the data connector will not be linked into the copy

The following message show up:

Image

But this is not the right case for that message. The DC is public and the copy should succeed fully.

olevski:
fixes #781


leafty (approved):
LGTM

PR #791, opened 2025-04-11
chore: ignore other files in the .vscode folder

lorenzo-cavazzi:
The repo provides settings for VScode, but the folder might contain other config files that shouldn't be included in the repo


olevski (approved):

lorenzo-cavazzi:
@olevski yeah, that's excluded here `!.vscode/settings.json`
@leafty we only provide `settings.json`, all the other config files should be adjustable by users

leafty:
> @leafty we only provide `settings.json`, all the other config files should be adjustable by users

That's not a good argument, something like tasks or launch settings should be checked in if they are ever created.

lorenzo-cavazzi:
Can't we change the gitignore file too at that point?
I have my own local `launch.json`

olevski (approved):

PR #794, opened 2025-04-13
fix: apply memory limits when patching

olevski:
This adresses an issue reported by a user on discourse. Since we were updating the limits only on start - but not when patching, the session can fail to resume if you switch the resource class with one with a larger memory. This is because the patching was not updating the memory limit and the request would be > the limit after patching.

/deploy

### DIFF
--- a/components/renku_data_services/notebooks/core_sessions.py
+++ b/components/renku_data_services/notebooks/core_sessions.py
@@ -359,7 +359,7 @@ def resources_from_resource_class(resource_class: ResourceClass) -> Resources:
         "cpu": str(round(resource_class.cpu * 1000)) + "m",
         "memory": f"{resource_class.memory}Gi",
     }
-    limits: dict[str, str | int] = {}
+    limits: dict[str, str | int] = {"memory": f"{resource_class.memory}Gi"}
     if resource_class.gpu > 0:
         gpu_name = GpuKind.NVIDIA.value + "/gpu"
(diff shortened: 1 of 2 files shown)


components/renku_data_services/notebooks/core_sessions.py
```diff
@@ -359,7 +359,7 @@ def resources_from_resource_class(resource_class: ResourceClass) -> Resources:
         "cpu": str(round(resource_class.cpu * 1000)) + "m",
         "memory": f"{resource_class.memory}Gi",
     }
-    limits: dict[str, str | int] = {}
```
olevski, line 362: This is the line was the bug.

components/renku_data_services/notebooks/core_sessions.py
```diff
@@ -359,7 +359,7 @@ def resources_from_resource_class(resource_class: ResourceClass) -> Resources:
...
+    limits: dict[str, str | int] = {"memory": f"{resource_class.memory}Gi"}
```
olevski, line 362: This addresses the bug

olevski:
Here is the thread where the user reported the problem: https://matrix.to/#/!MDFzvUYnYsMOMSdFSC:gitter.im/$2O5DP1B80_flFPXmF9DVDUgXVM2BhqlSAwWH9rhKAnU?via=gitter.im&via=matrix.org&via=staffchat.ethz.ch

Panaetius (approved):

PR #800, opened 2025-04-14
chore: update actions to 1.18.1

olevski:
The dependabot PR I think is maybe missing some permissions and it  was failing.

/deploy


olevski:
This PR will replace #752

PR #801, opened 2025-04-14
feat: add jupyterlab as option for builder

olevski:
Adding jupyterlab buildpack and frontend

/deploy renku-ui=feat-add-jupyterlab-option-for-builders extra-values=dataService.imageBuilders.enabled=true,dataService.imageBuilders.outputImagePrefix=harbor.dev.renku.ch/tasko-dev/,dataService.imageBuilders.pushSecretName=renku-build-secret,dataService.imageBuilders.strategyName=renku-buildpacks-v2


components/renku_data_services/session/models.py
```diff
@@ -306,6 +307,27 @@ class ShipwrightBuildRunParams:
     annotations: dict[str, str] | None = None
 
 
+@dataclass(frozen=True, eq=True, kw_only=True)
+class ShipwrightBuildRunParamsV2:
```
leafty, line 311: Question: is it not possible to share or migrate the existing `ShipwrightBuildRunParams`?

olevski: I made ShipwrightBuildRunParamsV2 a child of ShipwrightBuildRunParams in https://github.com/SwissDataScienceCenter/renku-data-services/pull/801/commits/f597a47b2428fe4449fd7c2fe0d200f74d7f5fea

olevski:
FYI the connection to the old search and redis queue got messed up and the search was not showing new things. The code I changed has nothing to do with this. So I just wiped and reinstalled the deployment.

This means any projects or other things in the ci deployment in renku were wiped.

components/renku_data_services/session/models.py
leafty: 1. I don't really understand why we need to have `ShipwrightBuildRunParams` and `ShipwrightBuildRunParamsV2` at the same time in the code. There will only ever exist `ShipwrightBuildRunParamsV2` loaded in renku-data-services once this PR is merged. So instead, the existing dataclass should simply be updated to reflect the new fields.
2. I think I would prefer having an update on the strategy file, so that we can easily diff it. But it should still be deployed with a new name on k8s so that we can deploy the strategy first, then upgrade the renku release so that the new builds do not break.

olevski: Applied in https://github.com/SwissDataScienceCenter/renku-data-services/pull/801/commits/fb0d713b130e0fc7e78ad32791de1f33e9fa5ec2

leafty (dismissed):
LGTM, works really well 👏

PR #803, opened 2025-04-15
feat: env_variable overrides on session start

ciyer:
/deploy renku=release-0.68.0


components/renku_data_services/notebooks/core_sessions.py
```diff
@@ -311,6 +313,24 @@ async def request_dc_secret_creation(
...
+
+    def env_var_item(name: str, launcher_value: str | None) -> SessionEnvItem | None:
+        if env_overrides.get(name):
+            return SessionEnvItem(name=name, value=env_overrides[name])
+        if not launcher_value:
+            return None
+        return SessionEnvItem(name=name, value=launcher_value)
+
+    env_vars = [env_var_item(env.name, env.value) for env in launcher.env_variables]
+    return [env_var for env_var in env_vars if env_var is not None]
```
olevski, line 331: suggestion: We can have  a regular loop and just one more list comprehension, instead of 3 list comprehensions. I think the logic in the regular loop is a bit easier to follow.

```suggestion
def get_launcher_env_variables(launcher: SessionLauncher, body: apispec.SessionPostRequest) -> list[SessionEnvItem]:
    """Get the environment variables from the launcher, with overrides from the request."""
    output: list[SessionEnvItem] = []
    env_overrides = {i.name: i.value for i in body.env_variables or []}
    for env in launcher.env_variables or []:
        if env.name in env_overrides:
            output.append(SessionEnvItem(name=env.name, value=env_overrides[env.name]))
        else:
            output.append(SessionEnvItem(name=env.name, value=env.value))
    return output

```

olevski: Also a separate but a related concern...

question: What do we do if the variable is in the overrides but not saved in the session launcher? In this case we just silently drop it? In this case I think we should either:
- A: error out with 422 letting the user know they are overriding a variable that is not in the launcher
- B: accept it as a new var

components/renku_data_services/notebooks/api.spec.yaml
```diff
@@ -865,6 +886,8 @@ components:
           type: integer
         cloudstorage:
           $ref: "#/components/schemas/SessionCloudStoragePostList"
+        env_variables:
```
olevski, line 889: suggestion: If this is only an override (not an upsert) than the name of this variable should have "override" in the name imo. As it stands right now it makes me think it is an upsert.

olevski (changes requested):
Thanks Sekhar. The code looks good. I just have a few suggestions/clarifications.

Almost all are about whether we upsert or just override env vars. And that we should decide and clarify that.

components/renku_data_services/notebooks/core_sessions.py
ciyer: As implemented, env variables that are not declared in the session launcher are silently ignored.

I think we do not want to do B, and only allow overrides for env variables that have been declared in the session launcher, but we could treat it as an error rather than just dropping the unknown variables.

I think the decision is a UX/prod decision -- we can wait until the next sync meeting to decide how to handle it.

olevski (dismissed):

olevski (approved):

PR #805, opened 2025-04-16
build: bump rclone to version 1.69.1

leafty:
Providers diff:
```
- amazon cloud drive
+ azurefiles
+ cloudinary
+ filescom
+ gofile
+ iclouddrive
+ imagekit
+ linkbox
+ pixeldrain
+ quatrix
+ ulozto
```

See: https://github.com/SwissDataScienceCenter/renku-ui/pull/3650.

All acceptance tests pass on the corresponding `renku-ui` deployment.


.devcontainer/rclone/install.sh
```diff
@@ -0,0 +1,15 @@
...
+mkdir -p /tmp/rclone
+cd /tmp/rclone
+git clone "${RCLONE_REPOSITORY}"
+cd rclone
+git checkout "${RCLONE_REF}"
+
+echo "Building rclone"
+make rclone
+cd $HOME
+rm -rf /tmp/rclone
```
olevski, line 15: suggestion: It will be much faster if you dont to build rclone from source. Rclone publishes executables in their github releases. https://github.com/rclone/rclone/releases. So you can just download and extract that.

components/renku_data_services/data_connectors/api.spec.yaml
```diff
@@ -683,6 +683,9 @@ components:
               provider:
                 type: string
                 description: The provider this value is applicable for. Empty if valid for all providers.
+            required:
+              - value
+              - help
```
olevski, line 688: question: This and the other change in the apispec.yaml are breaking changes for API. Now this by itself is usually not a problem since we do these often from time to time. But in this case we use this API to validate responses (i.e. when we list data connectors I think or when we list the config for a data connector). So it can happen that we have stuff in the DB that was validated against the old API that breaks this API. 

So I think that we either try to do a migration for this. OR we simply remove the validation on listing things - as long as the ui can handle the uncertainty this may represent (or the ui does not touch these properties that much).

olevski (changes requested):
Just a two things. If I am not mistaken this has the possibility of breaking listing data connectors (and a few other methods) that pull from the DB and validate with the new API we just updated.

.devcontainer/rclone/install.sh
leafty: The build step is not too long, at least much faster than the SOLR step anyway. This is also very simple to do to test fixes and changes on `rclone`, which we will soon have (see #741).

components/renku_data_services/data_connectors/api.spec.yaml
leafty: These fields ARE NOT coming from the DB but from the rclone schema. As such, they should be following the rclone schema. The `sensitive_fields` are copied from the schema by matching the corresponding fields, so this will pass validation (the schema endpoint passes validation).

olevski (approved):

PR #806, opened 2025-04-16
Feat/k8s client merge

sambuc:
This PR mostly improves upon the K8s Watcher client wrapper by merging the client related classes and type within the `k8s` package.

This prepares merging the feature of the external resource build, and prevents duplication of kr8s client wrapper within our code base.

This moves around code, without modifications outside of updating the import lines.


components/renku_data_services/k8s_watcher/__init__.py
```diff
@@ -1 +1,100 @@
 """K8s watcher and k8s clients."""
+
+from __future__ import annotations
```
Panaetius, line 3: since this is business logic, I'd but it in core.py to be consistent. __init__.py is a good place for package level initialization code (as it's processed on every import) and for exposing easier imports via __all__, but things like business logic should not be in __init__.py.

Panaetius (changes requested):
general suggestion, in the spirit of not having hundreds of modules, why not go all the way and just put all the code from components/k8s_watcher into components/k8s? it's all k8s related anyways?

components/renku_data_services/k8s_watcher/__init__.py
olevski: @Panaetius we discussed this but I did not want to make even larger refactoring changes during the build. We will do what you propose for sure during cooldown.

olevski:
> general suggestion, in the spirit of not having hundreds of modules, why not go all the way and just put all the code from components/k8s_watcher into components/k8s? it's all k8s related anyways?

@Panaetius we discussed this but we did not want to do it so that the changes are not that extreme. We will do this during cooldown.

components/renku_data_services/k8s_watcher/__init__.py
olevski: @sambuc when we discussed this what I had in mind is to just import `K8sWatcher` in `__init__.py`. Then it can be imported from the module. See for example this: https://github.com/SwissDataScienceCenter/renku-data-services/blob/main/components/renku_data_services/errors/__init__.py#L3

Sor you can import errors just like `from renku_data_services.errors import BaseError` rather than `from renku_data_services.errors.errors import BaseError`

sambuc: I have moved the content of `__init__.py` `core.py` and added import + re-export so that public objects appear at the main module level.

I also fixed a couple of thing to be more explicit about testing `None` values and small refactors.

I will submit later in the day the refactor to use an enum instead of lower case strings for the `Kinds` being cached.

olevski (approved):

PR #808, opened 2025-04-17
fix: improve efficiency of the K8s Entity Cache

sambuc:
Avoid one or more call to `lower()` on each K8s API call.

We cannot rely on just using `kind`, as the K8s API sometimes returns it in lower case and sometimes requires the exact case to be used.

Instead, we use `singular` which is set and initiaized as `kind` in lower case, thus calling `lower()` only on instantation of objects.

Also switch from a list of kinds to cache to a set.


Panaetius (approved):
thanks you, this looks much cleaner

Issue #792, opened 2025-04-11
follow up syncronization issues for k8s watcher in data services


Panaetius:
We have sync issues that we need to follow up on around:
- Hibernated status. I.e. patching for hibernation updates the hibernated field in the spec but for the status to be hibernated it takes another reconciliation cycle until it's available on the CR.
- Deletion. I.e. we can have cases where child resources of a session are being still deleted but the session has been deleted from the db. In this case recreating same or similar thing fails. We'll need to have something in place to not delete the session from the db and mark it as terminated before it actually is removed in k8s (finalizers and all)
- Creation timestamp on newly created sessions or similar sources. In some cases we cannot have the data service api respond with a null creation timestamp from a successful creation. But the first response after creation from amalthea does not have a creation timestamp. -> use `common.k8s_objects` data base entry `creation_date` instead in responses if it is not present on the CR

Panaetius:
the issue with the creation timestamp should be fixed in #811

PR #811, opened 2025-04-23
fix: get updated k8s state when creating objects in k8s cache

Panaetius:
related to #792 

we had this [in the old code](https://github.com/SwissDataScienceCenter/renku-data-services/blob/0caedf776c7a69fcd61e230775b69455e59a5976/components/renku_data_services/notebooks/api/classes/k8s_client.py#L127-L135) but it was missed when implementing the new cache


Panaetius:
closed since changes were integrated into base pr

PR #813, opened 2025-04-24
chore: make error handler work with sanic-ext 24.12

Panaetius:


bases/renku_data_services/data_api/app.py
```diff
@@ -39,6 +44,25 @@ def str_to_slug(value: str) -> Slug:
...
+    sanic_ext does not return contained exceptions as errors anymore, instead it returns a string.
+    This undoes that change.
+    """
+    try:
+        return validator(model, body)
+    except VALIDATION_ERROR as e:
+        raise ValidationError(
+            f"Invalid request body: {model.__name__}. Error: {e}",
+            extra={"exception": e},
+        ) from None
```
leafty, line 63: Out of curiosity: why not `raise ... from e`?

Panaetius: this is just the code how it was on their end before they made it even worse. But good point, if we monkepatch it, we might as well handle errors properly

leafty (approved):
LGTM

leafty (commented):
Comment: should we look at other validation methods? Is sanic_ext doing the same thing when validating query parameters, path parameters, headers?

Panaetius:
> Comment: should we look at other validation methods? Is sanic_ext doing the same thing when validating query parameters, path parameters, headers?

query use the same machinery so is also patched. we don't use sanic validation for headers. Path parameters aren't validated using sanic_ext (they're part of routing code and don't "fail", just a route doesn't match and then you get a 404)

PR #816, opened 2025-04-28
Build/external resources cluster entities

sambuc:
/deploy renku=release-0.68.0


components/renku_data_services/crc/db.py
```diff
@@ -391,6 +393,8 @@ async def update_resource_pool(self, api_user: base_models.APIUser, id: int, **k
                 kwargs["idle_threshold"] = None
             if kwargs.get("hibernation_threshold") == 0:
                 kwargs["hibernation_threshold"] = None
+            if kwargs.get("cluster_id") == "":
+                kwargs["cluster_id"] = None
```
olevski, line 397: reminder: Lets double check there is way to remove the cluster associated with a resource pool. I think with the code as written we cannot do this. Because when patching things are str | None and they default to None. And with this statement in the code below if the value of `kwargs["cluster_id"]` is None then no changes are made - we have the continue statement. I think you want to remove that `continue` statement below in the code. The problem is that you have to make sure you dont accidentally get `kwargs["cluster_id"]` being None because of the patch body or something like that that defaults to None if not set.

In some cases it is ambiguous what None means when you want to "remove" a value like this. For this purpose we have a special type to indicate resetting/removal. I.e.  search for `validate_project_patch` and look for the `RESET` type. But I think in this case you dont need this RESET type.

components/renku_data_services/crc/db.py
```diff
@@ -858,3 +866,72 @@ async def update_user(self, api_user: base_models.APIUser, keycloak_id: str, **k
...
+
+    @_only_admins
+    async def select_all(self, api_user: base_models.APIUser) -> list[Cluster]:
+        """Get cluster descriptions from the database."""
+
+        async with self.session_maker() as session:
+            r = await session.scalars(select(ClusterORM))
+            clusters = r.fetchall()
+
+            return [c.to_cluster() for c in clusters]
```
olevski, line 885: suggestion: Make this an async iterator and use sqlalchemy streams. See for example `get_all_users` in `db.py` in the users component.

components/renku_data_services/crc/db.py
```diff
@@ -858,3 +866,72 @@ async def update_user(self, api_user: base_models.APIUser, keycloak_id: str, **k
...
+    @_only_admins
+    async def update(self, api_user: base_models.APIUser, cluster: SavedCluster, cluster_id: ULID) -> Cluster:
+        """Updates a cluster description."""
+
+        async with self.session_maker() as session, session.begin():
+            saved_cluster = (await session.scalars(select(ClusterORM).where(ClusterORM.id == cluster_id))).one_or_none()
+            if saved_cluster is None:
+                raise errors.MissingResourceError(message=f"Cluster definition id='{cluster.id}' does not exist.")
+
+            saved_cluster.id = cluster.id
```
olevski, line 920: issue: We should not be changing the primary key.

components/renku_data_services/crc/orm.py
```diff
@@ -133,6 +136,24 @@ def dump(self, matching_criteria: models.ResourceClass | None = None) -> models.
         )
 
 
+class ClusterORM(BaseORM):
+    """Cluster definition."""
+
+    __tablename__ = "clusters"
+    id: Mapped[ULID] = mapped_column("id", ULIDType, primary_key=True, default_factory=lambda: str(ULID()), init=False)
+    name: Mapped[str] = mapped_column(String(40), unique=True, index=True)
```
olevski, line 144: question: Do we care if this is unique? This `name` field is only for the benefit of users. We have the primary key to identify clusters with - that is unique. But I think this does not have to be, right?

components/renku_data_services/crc/orm.py
```diff
@@ -133,6 +136,24 @@ def dump(self, matching_criteria: models.ResourceClass | None = None) -> models.
...
+    config_name: Mapped[str] = mapped_column(String(40), unique=True, index=True)
+
+    def to_cluster(self) -> SavedCluster:
+        """Create a cluster model from the ORM object."""
+        return SavedCluster(id=self.id, name=self.name, config_name=self.config_name)
+
+    @classmethod
+    def from_model(cls, cluster: Cluster) -> "ClusterORM":
```
olevski, line 152: nitpick: In all other instances we call these methods dump and load.

components/renku_data_services/crc/models.py
```diff
@@ -242,16 +274,23 @@ def update(self, **kwargs: Any) -> "ResourcePool":
...
         elif "quota" in data and isinstance(data["quota"], Quota):
             quota = data["quota"]
+
         if "classes" in data and isinstance(data["classes"], set):
             classes = [ResourceClass.from_dict(c) if isinstance(c, dict) else c for c in list(data["classes"])]
         elif "classes" in data and isinstance(data["classes"], list):
             classes = [ResourceClass.from_dict(c) if isinstance(c, dict) else c for c in data["classes"]]
+
+        if "cluster" in data and isinstance(data["cluster"], Cluster):
+            cluster = Cluster.from_dict(data["cluster"]) if isinstance(data["cluster"], dict) else data["cluster"]
```
olevski, line 292:
```suggestion
        if "cluster" in data:
            cluster = Cluster.from_dict(data["cluster"]) if isinstance(data["cluster"], dict) else data["cluster"]
```

olevski (commented):
Thank you Lionel this looks really good! I was really picky in the review so there are several minor issues.

components/renku_data_services/crc/db.py
olevski: No need for the RESET type. Just update the logic to be like the "hibernated_threshold" property. And check if the cluster pointed by by the cluter_id really exists - if it does not fail with 404.

components/renku_data_services/crc/db.py
olevski: In the update method make the cluster parameter be of type "UnsavedCluster" which does not contain an id.

components/renku_data_services/crc/orm.py
olevski: Leave as is.

components/renku_data_services/crc/api.spec.yaml
```diff
@@ -834,6 +970,40 @@ paths:
 
 components:
   schemas:
+    Cluster:
+      type: object
+      additionalProperties: false
+      properties:
+        name:
+          $ref: "#/components/schemas/Name"
+        config_name:
```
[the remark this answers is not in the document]
olevski: Lets keep the same name for the property but change the description to explain that:
- it is a secret in k8s
- the secret contains kubeconfig 
- it is in the same namespace as the renku services

components/renku_data_services/crc/apispec_base.py
```diff
@@ -10,3 +16,41 @@ class Config:
...
+    def serialize_environment_id(cls, environment_id: str | ULID | None) -> str | None:
+        """Custom serializer that can handle ULIDs."""
+        if environment_id is None:
+            return None
+        return str(environment_id)
+
+    @field_validator("environment_kind", mode="before", check_fields=False)
+    @classmethod
+    def serialize_environment_kind(cls, environment_kind: models.EnvironmentKind | str) -> str:
+        """Custom serializer that can handle ULIDs."""
```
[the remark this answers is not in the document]
sambuc: done

components/renku_data_services/crc/api.spec.yaml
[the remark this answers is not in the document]
sambuc: done, please tell me if the description is clear enough for you.

components/renku_data_services/crc/api.spec.yaml
```diff
@@ -9,310 +9,334 @@ info:
...
+            "application/json":
+              schema:
+                $ref: "#/components/schemas/ClustersWithId"
+        default:
+          $ref: "#/components/responses/Error"
+      tags:
+        - clusters
     post:
-      summary: Create a new resource pool
+      summary: Create a new cluster description
```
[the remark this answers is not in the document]
sambuc: applied everywhere where this was relevent

components/renku_data_services/crc/blueprints.py
```diff
@@ -566,3 +568,101 @@ async def _post_put(
...
+        """Update the cluster descriptions."""
+
+        @authenticate(self.authenticator)
+        @only_admins
+        @validate(json=apispec.Cluster)
+        async def _handler(
+            _request: Request, user: base_models.APIUser, cluster_id: ULID, body: apispec.Cluster
+        ) -> HTTPResponse:
+            # We allow the cluster_id to be modified. The operation might fail if the new id already exists in the DB.
+            cluster = SavedCluster(id=cluster_id, name=body.name, config_name=body.config_name)
```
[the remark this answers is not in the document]
sambuc: done

components/renku_data_services/crc/blueprints.py
```diff
@@ -566,3 +568,101 @@ async def _post_put(
...
+        """Patch the cluster descriptions."""
+
+        @authenticate(self.authenticator)
+        @only_admins
+        @validate(json=apispec.ClusterPatch)
+        async def _handler(
+            _request: Request, user: base_models.APIUser, cluster_id: ULID, body: apispec.ClusterPatch
+        ) -> HTTPResponse:
+            # We allow the cluster_id to be modified. The operation might fail if the new id already exists in the DB.
+            old = await self.repo.select(user, cluster_id)
```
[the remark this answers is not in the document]
sambuc: done

components/renku_data_services/crc/db.py
sambuc: done

components/renku_data_services/crc/db.py
sambuc: done

components/renku_data_services/crc/db.py
sambuc: done

components/renku_data_services/crc/orm.py
sambuc: renamed

components/renku_data_services/crc/models.py
```diff
@@ -173,12 +175,41 @@ def generate_id(self) -> "Quota":
...
+
+    @classmethod
+    def from_dict(cls, data: dict) -> "Cluster":
+        """Instantiate a SavedCluster from the dictionary."""
+        cls._check_keys(["name", "config_name"], data)
+
+        if "id" in data:
+            return SavedCluster(**{**data})
+        else:
+            return UnsavedCluster(**{**data})
```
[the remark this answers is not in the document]
sambuc: - imported __future__
- removed check_keys
- simplified from_dict & fixed up type usage Cluster vs SavedCluster vs UnsavedCluster

components/renku_data_services/crc/models.py
```diff
@@ -193,7 +224,8 @@ class ResourcePool:
     hibernation_threshold: Optional[int] = None
     default: bool = False
     public: bool = False
-    cluster: Optional[KubeClusterSettings] = None
+    cluster_id: Optional[ULID] = None
+    cluster: Optional[Cluster] = None
```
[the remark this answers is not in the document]
sambuc: done

components/renku_data_services/crc/models.py
sambuc: Done, I also had to add the case to handle None.

components/renku_data_services/crc/api.spec.yaml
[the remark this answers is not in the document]
olevski: Yes that is clear.

PR #822, opened 2025-05-05
feat: move shipwright clients over to use multi cluster cached client

Panaetius:


components/renku_data_services/k8s/config.py
```diff
@@ -0,0 +1,111 @@
...
+        kubeconfig: str | None = None,
+        current_context_name: str | None = None,
+        ns: str | None = None,
+        sa: str | None = None,
+        url: str | None = None,
+    ) -> None:
+        self._kubeconfig = kubeconfig
+        self._ns = ns
+        self._current_context_name = current_context_name
+        self._sc = sa
```
olevski, line 28: suggestion: This is a leftover from Lionel's PR. I think he misunderstood my comment. We should rename `self._sc` to `self._sa` because this is the service account. And the short name for service account is `sa` not `sc`. `sc` is actually the short name for storage class.

components/renku_data_services/session/k8s_client.py
```diff
@@ -395,4 +279,35 @@ async def get_image_build_logs(self, buildrun_name: str, max_log_lines: int | No
         pod_name = taskrun.status.podName if taskrun.status else None
         if not pod_name:
             raise errors.MissingResourceError(message=f"The buildrun {buildrun_name} has no pod to retrieve logs from.")
-        return await self.base_client.get_pod_logs(name=pod_name, max_log_lines=max_log_lines)
+        return await self.get_pod_logs(name=pod_name, max_log_lines=max_log_lines)
+
+    async def get_pod_logs(self, name: str, max_log_lines: int | None = None) -> dict[str, str]:
```
olevski, line 284: suggestion: I think we should make this private. It is not used anywhere else. And I dont see that anyone would call this directly. In this case we care about the logs of the image build.

olevski (changes requested):
Thanks Ralf this looks really good. Just a few minor changes.

components/renku_data_services/k8s/config.py
Panaetius: could also be a rebase mistake. I'll fix it

olevski: That is a good point. Thanks!

olevski (dismissed):

olevski (approved):

PR #831, opened 2025-05-09
fix: make 404 errors quiet by default

### LINKED ISSUE #687: We log exceptions that are expected with full traces
A 401 or 404 will result in a full trace in the logs. This should not happen.

We have the "quiet" field in the errors that suppresses this. But it would be nice if we made the default to be to not print traces when an expected error is raised on purpose in the code.

leafty:
Closes #687.

Also, remove `quiet=True` where it is already the default.


olevski (approved):
Nice!

PR #834, opened 2025-05-12
fix: disable k8s cache watching image builders if they aren't enabled

Panaetius:
/deploy renku=fix-cache-image-builders
/AzureDeploy renku=fix-cache-image-builders


leafty:
Can you test on Azure? The CRDs are not deployed on `azure-dev`.

leafty (approved):
LGTM

PR #838, opened 2025-05-12
fix: sync k8s caches at startup and every 30 mins

olevski:
/deploy renku=release-0.68.0


olevski:
I know why the tests fail. I will fix this.

components/renku_data_services/k8s/clients.py
```diff
@@ -316,6 +317,28 @@ def __init__(self, cluster: Cluster, cache: K8sDbCache, kinds_to_cache: list[str
...
+
+    async def __sync(self) -> None:
+        """Upsert K8s objects in the cache and remove deleted objects from the cache."""
+        for kind in self.__kinds_to_cache:
+            fltr = K8sObjectFilter(kind=kind, namespace=self.get_cluster().namespace)
+            # Upsert new / updated objects
+            async for obj in super().list(fltr):
+                await self.__cache.upsert(obj)
+            # Remove objects that have been deleted from k8s but are still in cache
+            async for cache_obj in self.__cache.list(fltr):
```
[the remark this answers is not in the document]
olevski: I fixed this as well.

components/renku_data_services/k8s/clients.py
```diff
@@ -316,6 +317,28 @@ def __init__(self, cluster: Cluster, cache: K8sDbCache, kinds_to_cache: list[str
         super().__init__(cluster)
         self.__cache = cache
         self.__kinds_to_cache = set(k.lower() for k in kinds_to_cache)
+        self.__sync_period_seconds = 1800
+        self.__sync_task = asyncio.create_task(self.__periodic_sync())
+
+    async def __sync(self) -> None:
+        """Upsert K8s objects in the cache and remove deleted objects from the cache."""
+        for kind in self.__kinds_to_cache:
+            fltr = K8sObjectFilter(kind=kind, namespace=self.get_cluster().namespace)
```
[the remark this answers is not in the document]
olevski: @Panaetius I added a GVK class and added group, version and kind explicitly everywhere.

I did not notice this in the review but I found it pretty scary that what went into the db for `version` was what kr8s puts together in their weird logic. So I separated things out. This way we have group, version and kind fields in the db. We have case insensitive matching for these fields in the db and we do not let the way kr8s defines these to leak into the DB.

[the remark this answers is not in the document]
olevski: So I addressed this comment and then added more stuff.

components/renku_data_services/k8s_watcher/orm.py
```diff
@@ -23,6 +25,23 @@ class BaseORM(MappedAsDataclass, DeclarativeBase):
     registry = COMMON_ORM_REGISTRY
 
 
+class CaseInsensitiveComparator(Comparator[str]):
```
Panaetius, line 28: nice!

bases/renku_data_services/data_api/main.py
```diff
@@ -148,6 +148,13 @@ async def setup_rclone_validator(app: Sanic) -> None:
         validator = RCloneValidator()
         app.ext.dependency(validator)
 
+    @app.before_server_start
+    async def start_k8s_cache_synchronization(app: Sanic) -> None:
+        sync_coroutines = config.nb_config.k8s_client.get_synchronization_coroutines()
+        if config.builds_config.enabled and config.builds_config.shipwright_client is not None:
+            sync_coroutines.extend(config.builds_config.shipwright_client.client.get_synchronization_coroutines())
+        app.add_task(asyncio.gather(*sync_coroutines), name="k8s_cache_sync")
+
```
Panaetius, line 157: question: why run this in main data services instead of k8s watcher or data_tasks jobs? This will run on each horizontally scaled data-svc deployment, not sure that's what we want.

I'd personally just add it in the loop here: https://github.com/SwissDataScienceCenter/renku-data-services/blob/main/components/renku_data_services/k8s_watcher/core.py#L103 (gated by an `if datetime.now() <last_sync + interval:` kinda thing) then there's no concurrency issue or anything like that. and code handling each kind can be kinda self-contained

olevski: Yeah I will move it. I thought I had a reason but I cannot think of it now.

olevski: done - you can look at the 2nd of the two new commits to see the changes

Panaetius (dismissed):

components/renku_data_services/k8s_watcher/core.py
```diff
@@ -7,80 +7,19 @@
...
 from renku_data_services.base_models.core import APIUser, InternalServiceAdmin, ServiceAdminId
 from renku_data_services.base_models.metrics import MetricsService
 from renku_data_services.crc.db import ResourcePoolRepository
-from renku_data_services.errors import errors
 from renku_data_services.k8s.clients import K8sClusterClient
-from renku_data_services.k8s.models import GVK, Cluster, ClusterId, K8sObject, K8sObjectFilter, K8sObjectMeta
+from renku_data_services.k8s.models import GVK, K8sObject, K8sObjectFilter
 from renku_data_services.k8s_watcher.db import K8sDbCache
 from renku_data_services.notebooks.crs import State
-from renku_data_services.session.constants import DUMMY_TASK_RUN_USER_ID
```
olevski, line 25: All of this (i.e. `APIObjectInCluster`) just got moved to another file.

Panaetius (approved):

PR #832, opened 2025-05-12
fix: Upgrade kr8s to 20.7

sambuc:
* async generators for `list()` and `get()`
 * might help with refreshing the service account credentials

Deployment on Azure: https://github.com/SwissDataScienceCenter/renku/pull/4007
 
 /deploy renku=release-0.68.0

PR #840, opened 2025-05-13
fix: Upgrade kr8s to 20.7 (#832)

olevski:
This re-introduces the changes that were reverted in #837 

/deploy

This will need to be tested on Azure as well because that is where we saw problems with then newest version of kr8s.


olevski:
This upgrade was already done in some other pr.

PR #841, opened 2025-05-13
fix: add default for show_project_migration_banner

olevski:
/deploy renku=master


olevski:
I will open another test PR to replicate the error conditions with a pre-existing users/rows in the table

components/renku_data_services/users/orm.py
```diff
@@ -79,7 +79,11 @@ class UserPreferencesORM(BaseORM):
     pinned_projects: Mapped[dict[str, Any]] = mapped_column("pinned_projects", JSONVariant)
     """Pinned projects."""
 
-    show_project_migration_banner: Mapped[bool] = mapped_column("show_project_migration_banner", Boolean)
+    show_project_migration_banner: Mapped[bool] = mapped_column(
+        "show_project_migration_banner",
+        Boolean,
+        server_default=true(),
```
leafty, line 85:
```suggestion
        default=True,
        server_default=true(),
```

leafty (changes requested):
See comments

olevski:
I checked on a test ci deployment where I deployed 0.67.2 , made some users and projects, then applied the latest version of renku with the fix here. And the migrations worked.

leafty (approved):

PR #844, opened 2025-05-14
refactor: separate configs and dependency management

Panaetius:
Splits data services and secrets storage config into separate dependency management and config. Moves the top level config and the dependency management into bases/ instead of app_config.

Note that there's still other places that need to be cleaned up, e.g. SyncConfig, and this is just a first step for the main services.

the commits in this PR should be split in a way that makes it easier to review if going from commit to commit, in order.

/deploy


components/renku_data_services/app_config/config.py
```diff
@@ -9,120 +9,43 @@
 instantiated multiple times without creating multiple database connections.
```
olevski, line 9: issue: This whole docstring is not really applicable now. As far as I can see we dont do any of the `@property` stuff now in this specific file.

components/renku_data_services/app_config/config.py
```diff
@@ -9,120 +9,43 @@
...
-)
+
+@dataclass
+class KeycloakConfig:
+    """Configuration values for keycloak."""
+
+    keycloak_url: str
+    keycloak_realm: str
+    client_id: str
+    client_secret: str
```
olevski, line 25: Also make sure to import `field` from dataclasses at the top.


```suggestion
    client_secret: str = field(repr=False)
```

bases/renku_data_services/data_tasks/config.py
```diff
@@ -20,12 +20,14 @@ class PosthogConfig:
     environment: str
```
olevski, line 20: I know it is like that from before and you did not touch it here. But lets make the api_key have `field(repr=False)`

bases/renku_data_services/data_api/config.py
```diff
@@ -0,0 +1,89 @@
...
+from renku_data_services.secrets.config import PublicSecretsConfig
+from renku_data_services.session.config import BuildsConfig
+from renku_data_services.solr.solr_client import SolrClientConfig
+from renku_data_services.users.config import UserPreferencesConfig
+
+
+@dataclass
+class Config:
+    """Application configuration."""
+
```
olevski, line 23: This cleaned up really nicely!!!

olevski (changes requested):
Thanks Ralf. This looks really good. Just a several really small suggestion.

olevski (approved):

PR #864, opened 2025-05-16
chore: add pr templates

olevski:


olevski:
Actually we can probably fully eliminate the bugfix tempalte. And for bugfixes just use the commit message.

.github/pull_request_template/bugfix.md
```diff
@@ -0,0 +1,18 @@
...
+## Internal Changes
+
+**Bug Fixes**
+
+- **Data services**:
+
+# Internal/Miscellaneous
+
+Add a deploy string here or other things that are not relevant for the release notes
+or commit messages.
```
Panaetius, line 18:
```suggestion
<!-- Add a deploy string here or other things that are not relevant for the release notes
or commit messages. -->
```

also for other comments. Then you don't need to delete but it won't be rendered

olevski: Ah I thought there was some trick like that but wasn't sure. Thanks Ralf I will make the change.

olevski: Done. I also removed the bugfix template./

Panaetius (approved):

PR #865, opened 2025-05-19
fix: delete data connector when its slug is deleted

### LINKED ISSUE #859: Deleting a group does not delete the data connectors in the group
To replicate:
1. Create a group
2. Create a DC in the group
3. Delete the group

This is causing us to have a bunch of data connectors that have no namespaces in the db.

While we are here we should also make sure that deletion propagates for data connectors for all the other cases:
- deleting a project should delete all data connectors that are in the project namespace
- deleting a user should also delete all data connectors that are in the user namespace

leafty:
Closes #859.

The migration adds a trigger and deletes the straggler data connectors which should have been deleted previously.

Manual tests:
1. On this deployment:
    1. Create a group
    2. Create a data connector in the group
    3. Delete the group
    4. Check that listing all data connectors does not crash data services (HTTP 500)
2. On https://ci-renku-4017.dev.renku.ch/v2:
    1. Deploy before this change
    2. Create a group
    3. Create a data connector in the group
    4. Delete the group
    5. Listing all data connectors will crash data services (HTTP 500)
    6. Upgrade to this version
    7. Check that listing all data connectors does not crash data services (HTTP 500)

/deploy

### DIFF
--- a/components/renku_data_services/migrations/versions/f4ad62b7b323_cleanup_data_connector_when_slug_is_.py
+++ b/components/renku_data_services/migrations/versions/f4ad62b7b323_cleanup_data_connector_when_slug_is_.py
@@ -1,0 +1,50 @@
+"""Cleanup data connector when slug is removed
+
+Revision ID: f4ad62b7b323
+Revises: dcb9648c3c15
+Create Date: 2025-05-19 07:15:11.989650
+
+"""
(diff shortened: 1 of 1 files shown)


olevski (approved):
That is great! Thanks Flora!

PR #875, opened 2025-05-27
fix: remove hardcoded defaults for args and command

olevski:
/deploy

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,3 +15,4 @@ compute resources.
 ## Developing
 
 See the [Development documentation for more details](/DEVELOPING.md)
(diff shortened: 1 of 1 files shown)


olevski:
Note some of the acceptance tests were failing because I was not targeting the release branch.

sambuc:
Does this still allow the users to set a non-default entry point in a launcher?

olevski:
@sambuc this is strictly for images built from code. And for these users are not allowed to modify anything like that. We set that for them.

sambuc (approved):

PR #878, opened 2025-05-27
fix: allow anonymous users to get build details

lorenzo-cavazzi:
Anonymous users can start sessions, but the UI cannot show the proper buttons without knowing the status of the last build for launchers based on code repositories.

This PR relaxes the requirements to invoke the builds related PRs. It will be used in https://github.com/SwissDataScienceCenter/renku-ui/pull/3707

/deploy

### DIFF
--- a/bases/renku_data_services/data_api/__init__.py
+++ b/bases/renku_data_services/data_api/__init__.py
@@ -1,1 +1,5 @@
 """Renku data service."""
+
+import renku_data_services.app_config.logging as logging
+
+logging.configure_logging()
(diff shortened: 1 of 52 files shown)


olevski (changes requested):
Thanks for the PR Lorenzo.

For the logs we should allow only logged in users to see them. I think this is safer and more conservative. They may contain information sensitive to the user.

And for `get_build` you can fix the error simply by calling `self._refresh_build` if the user.id is not None. That part is on line 775.

olevski (approved):

PR #879, opened 2025-05-28
chore: turn on flake8-bugbear linter

olevski:
I was looking for a linter that can catch default mutable values assigned to function args.

And this one has that rule. It also has some other rules that are pretty neat imo. And it also forces us to think about whether we do or dont want to chain exceptions. Which is nice.

So I added all the rules from it and made the fixes. Most were about the exceptions chaining.

https://docs.astral.sh/ruff/rules/#flake8-bugbear-b

/deploy


Panaetius (commented):
just a minor comment, otherwise all good (and found some real bugs!)

PR #882, opened 2025-06-02
fix: upgrade buildpacks to 0.0.6

olevski:
This fixes a bug with not having proper shell completion and .bashrc sourcing.


leafty (approved):
LGTM

Note: checked that image exists, not that the image works as intended.

PR #887, opened 2025-06-06
fix: Skip authz if admin is requesting data connectors

eikek:
For reprovisioning it is necessary to get all entities. When selecting data connectors from the db, there was always a call to authz to first get all identifiers of resources the requesting user has read access to. For internal service admins, this doesn't work, because they are not modelled in authz. For admins in general the call should be prevented, because admins can see everything. 

This change skips the authz call if the requesting user is an admin. Additionally the search reprovisioning code uses an internal service admin for doing its work and not the user passed through from the endpoint call. 

/deploy

### DIFF
--- a/bases/renku_data_services/data_api/main.py
+++ b/bases/renku_data_services/data_api/main.py
@@ -17,7 +17,6 @@ from sentry_sdk.integrations.sanic import SanicIntegration, _context_enter, _con
 import renku_data_services.solr.entity_schema as entity_schema
 from renku_data_services.app_config import logging
 from renku_data_services.authz.admin_sync import sync_admins_from_keycloak
-from renku_data_services.base_models.core import APIUser
 from renku_data_services.data_api.app import register_all_handlers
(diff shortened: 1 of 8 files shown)


leafty:
> @leafty this is related to the special InternalServiceAdmin entities that are not in authzed.

Maybe we should consider having the service account in authz. In the meantime we can have a bypass for this case.

eikek:
@olevski ok, I then revert the second case. I don't think it's too bad, though tbh. One could argue that some admin triggered the reprovisioning and the reprovisioning process itself is run by an internal account. In my opinion, there is no need for that code (deep down doing the reprovisioing) to know anything about a calling user. just my opinion and I happily change it as advised.

@leafty I would also be in favor to have one way to determine "admin" status. If we have a flag, that's fine - but if we have a service admin in authzed, perhaps we can remove the flag or have it initialized by authzed so it can be trusted.

olevski (approved):

PR #888, opened 2025-06-11
fix: add autheticated property to events

olevski:
Some metric events were missing the required authenticated property to indicate if the user that triggered the event was anonymous or not.

/deploy


mohammad-alisafaee (approved):
LGTM, thank you!

PR #897, opened 2025-06-13
feat: add metrics for copied and migrated projects

rokroskar:
Adds missing metrics for copied and migrated projects. It uses the same `project_created` metric with a new metadata field that specifies `project_creation_kind` to be either `NEW`, `MIGRATED` or `COPIED`.


rokroskar:
For some reason the optional avro schema test is failing 🤷‍♂️

PR #908, opened 2025-06-27
chore: reduce default db pool sizes

olevski:
We use a default pool size of 10 that is fully hardcoded. But this is not necessary in all the auxiliary services like the watcher, tasks and secrets. So I reduced it to 4 in those places.

/deploy


eikek (approved):
I think `pool-size` is the only data not read from an environment variable. Is that on purpose? Wouldn't it be nicer (more consistent, more honoring to the name `from_env`, configurable) to also read that from an env variable?

leafty (approved):
LGTM

It would be even better to have it as `os.environ.get("DB_POOL_SIZE") or "4"` and configure values in the renku helm chart.

olevski:
>You can add ENV DB_POOL_SIZE=10 to the data service Dockerfile. Thus the default can be different for each service but still configurable (later) from the helm chart.

Ah yeah I did not think of adding it in the dockerfile. I will do that.

leafty (approved):

PR #911, opened 2025-06-30
feat: remove the message queue code

olevski:
/deploy 

Removes all code related to:
- message queue
- redis

Will do a followup PR to move the little bit of code in the message queue component that the search uses for reprovisioning. This code will be moved into the search component.


olevski:
p.s. not sure what to call this (chore with breaking changes, refactor?).

test/bases/renku_data_services/data_api/test_projects.py
```diff
@@ -1247,7 +1204,7 @@ async def test_project_copy_basics(
     assert copy_project == snapshot(exclude=props("id", "updated_at", "creation_date", "etag", "template_id"))
 
     _, response = await sanic_client.get(
-        f"/api/data/projects/{copy_project["id"]}", params={"with_documentation": True}, headers=user_headers
+        f"/api/data/projects/{copy_project['id']}", params={"with_documentation": True}, headers=user_headers
```
eikek, line 1207: Are these single quotes on purpose here?

eikek (dismissed):

test/bases/renku_data_services/data_api/test_projects.py
olevski: Ah that is my autoformatting doing weird things. I will remove that.

eikek (dismissed):
Another thought: do we need redis for something else? If not (I think the whole message queue code is gone), perhaps then the redis client dependency could also be removed? But, of course, can be a later PR.

olevski:
@eikek yes you are right I can remove the redis client from the package dependencies. I forgot about this.

olevski:
@eikek this is now fully ready for a final approval. A little bit of cleanup and small fixes to tests were added since your last review.

eikek (dismissed):
👍🏼 sorry for the late response here!

PR #933, opened 2025-07-21
feat: add support for git revision and context dir options to image builds

leafty:
wip


components/renku_data_services/session/orm.py
```diff
@@ -203,13 +203,21 @@ class BuildParametersORM(BaseORM):
 
     frontend_variant: Mapped[str] = mapped_column("frontend_variant", String(99))
 
+    repository_revision: Mapped[str | None] = mapped_column(
+        "repository_revision", String(500), nullable=True, default=None
+    )
+
+    context_dir: Mapped[str | None] = mapped_column("context_dir", String(500), nullable=True, default=None)
```
eikek, line 210: just a nitpick: I personally wouldn't restrict the db varchars to some arbitrary length (assuming it is arbitrary here :)). There is no gain from postgres side and it would error for directories of 501 length for probably no good reason. But we do it everywhere, so maybe just for consistency :-)

eikek (approved):
👍🏼

components/renku_data_services/session/orm.py
leafty: I would not expect any usual use-case to it the limit for a folder path. We have to restrict the length of strings we store in the DB, otherwise it can be used to store arbitrary binary data.

PR #938, opened 2025-07-24
fix: update ports for the git proxy

### LINKED ISSUE #921: Session API should not accept non-usable session ports
See: https://github.com/SwissDataScienceCenter/renku-ui/issues/3757

Note that session services should use a pre-defined range.

The API should be updated to reject ports <= 1024 and ports in the reserved range(s).

leafty:
Closes #921.

Details:
* Moves the git proxy port inside the port range reserved for renku sessions
* Requires an update to `git-clone` -> https://github.com/SwissDataScienceCenter/renku-notebooks/pull/2032, https://github.com/SwissDataScienceCenter/renku-notebooks/pull/2035
* Deploying renku with this update will affect paused sessions -> the new git proxy will not listen on port `8080` while the git configuration will still point at the old port. -> Ask users to shut down then restart?
* The port can be configured in the helm chart -> https://github.com/SwissDataScienceCenter/renku/pull/4120

/deploy renku=leafty/update-git-proxy-ports extra-values=enableV1Services=false,postgresql.primary.persistence.enabled=true,notebooks.gitClone.image.name=leafty/test,notebooks.gitClone.image.tag=git-clone-9c4ffbd


components/renku_data_services/notebooks/config/dynamic.py
```diff
@@ -119,8 +119,8 @@ def from_env(cls) -> Self:
             renku_client_secret=os.environ["NB_SESSIONS__GIT_PROXY__RENKU_CLIENT_SECRET"],
             renku_client_id=os.environ.get("NB_SESSIONS__GIT_PROXY__RENKU_CLIENT_ID", "renku"),
             sentry=_SentryConfig.from_env(prefix="NB_SESSIONS__GIT_PROXY__"),
```
leafty, line 121: just noticed this, is this correct?

olevski (approved):

PR #939, opened 2025-07-24
feat: don't allow change of resource class if it means change of cluster

### LINKED ISSUE #935: Changing clusters when patching a session should not be allowed
See: https://github.com/SwissDataScienceCenter/renku-data-services/blob/8be58eb7918017a7a8323aac53e35b6ef4fd5c2c/components/renku_data_services/notebooks/core_sessions.py#L527-L544

The patch code does not raise an error when a user tries to switch resource classes which would also switch clusters. This needs to be prevented as this operation would fail: the session volume and other session resources (e.g. secrets) cannot be transported across clusters.

sgaist:
## Describe your changes

This PR adds a check to ensure that resource class changes do not happen if they are configured with a target cluster that is different than the one currently in use by the session.

## Issue ticket number and link

Fixes #935

### DIFF
--- a/components/renku_data_services/notebooks/core_sessions.py
+++ b/components/renku_data_services/notebooks/core_sessions.py
@@ -525,6 +525,11 @@ async def patch_session(
 
     # Resource class
     if body.resource_class_id is not None:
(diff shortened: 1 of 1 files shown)


leafty (approved):
🚀

olevski (approved):

PR #954, opened 2025-08-06
chore: Simplify Makefile, rename api.spec.yaml

sambuc:


Makefile
```diff
@@ -55,20 +54,6 @@ API_SPECS := \
...
-components/renku_data_services/project/apispec.py: components/renku_data_services/project/api.spec.yaml
-components/renku_data_services/session/apispec.py: components/renku_data_services/session/api.spec.yaml
-components/renku_data_services/namespace/apispec.py: components/renku_data_services/namespace/api.spec.yaml
-components/renku_data_services/secrets/apispec.py: components/renku_data_services/secrets/api.spec.yaml
-components/renku_data_services/connected_services/apispec.py: components/renku_data_services/connected_services/api.spec.yaml
-components/renku_data_services/repositories/apispec.py: components/renku_data_services/repositories/api.spec.yaml
-components/renku_data_services/notebooks/apispec.py: components/renku_data_services/notebooks/api.spec.yaml
-components/renku_data_services/platform/apispec.py: components/renku_data_services/platform/api.spec.yaml
-components/renku_data_services/data_connectors/apispec.py: components/renku_data_services/data_connectors/api.spec.yaml
-components/renku_data_services/search/apispec.py: components/renku_data_services/search/api.spec.yaml
```
leafty, line 70: We can just remove this and not do the file renaming and everything will still work.

leafty (changes requested):
We don't need to rename the files 👇

Makefile
sambuc: done

PR #957, opened 2025-08-07
chore: fix the acceptance-tests.yml workflow

leafty:
Most importantly, this fixes not running the acceptance tests.

### DIFF
--- a/.github/workflows/acceptance-tests.yml
+++ b/.github/workflows/acceptance-tests.yml
@@ -1,5 +1,6 @@
 
 name: Acceptance tests
+
 permissions:
   contents: read
   pull-requests: read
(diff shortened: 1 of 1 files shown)


.github/workflows/acceptance-tests.yml
```diff
@@ -120,39 +122,6 @@ jobs:
           amalthea_sessions: "${{ needs.check-deploy.outputs.amalthea-sessions }}"
           extra_values: "${{ needs.check-deploy.outputs.extra-values }}"
 
-  legacy-selenium-acceptance-tests:
```
leafty, line 123: Note: removed because we should not run this from this repo.

PR #964, opened 2025-08-11
fix: remove unsafe STFP options for rclone

### LINKED ISSUE #963: Remove `key_file` option from SFTP/SSH rclone provider
When performing the connection test, this will attempt to read the file path on the `renku-data-services` worker, which could leak data.

Also, remove any similar option referencing a local file for other providers.

Similarly, the `ssh binary` option needs to be removed, as the user can then try to execute any binary present on the server.

leafty:
Closes #963.

Also, some minor cleanup for the rclone options.

/deploy extra-values=enableV1Services=false,postgresql.primary.persistence.enabled=true

### DIFF
--- a/test/bases/renku_data_services/data_api/test_storage.py
+++ b/test/bases/renku_data_services/data_api/test_storage.py
@@ -249,6 +249,21 @@ async def storage_test_client(
             422,
             "",
         ),
+        (
+            {
+                "project_id": "123456",
+                "name": "mystorage",
+                "configuration": {
+                    "type": "sftp",
+                    "host": "myhost",
+                    "ssh": "ssh",  # passing in banned option
+                },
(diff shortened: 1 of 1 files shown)


components/renku_data_services/storage/rclone.py
```diff
@@ -369,6 +377,8 @@ def validate_config(
             raise errors.ValidationError(message=f"The following fields are required but missing:\n{missing_str}")
 
         for key in keys:
+            self.check_unsafe_option(key)
```
leafty, line 380: Note: Code below (L390) basically allows any option to be provided by hand (advanced config). If some other options are unsafe, they need to be rejected by `check_unsafe_option()`. This means that the OAuth2 options we hide can in fact be used, though I'm not sure what would happen then. At least, those should not execute harmful code paths.

PR #969, opened 2025-08-13
fix: set default cluster name back to renkulab

olevski:
/deploy

Note: the target branch is not master. This is just to fix the problem with conflicting names in the k8s objects table. The bug occurs because we updated the cluster name but did not migrate the postgres table. And the k8s objects table requires the name of all resources objects to be unique (regardless of which cluster they belong to). So when we changed the default cluster name without migrating the existing (old) cluster name we have a conflict when we try to upgrade existing resources that are in the db table with the old cluster name.


olevski:
In master we need to do this: https://github.com/SwissDataScienceCenter/renku-data-services/issues/970 to properly address this.

leafty (approved):

PR #997, opened 2025-09-01
feat: support for redirecting URLs

ciyer:


components/renku_data_services/platform/api.spec.yaml
```diff
@@ -67,6 +161,45 @@ components:
...
+        source_url:
+          description: The protocol-relative URL to redirect
+          type: string
+        target_url:
+          description: The protocol-relative URL to redirect to
+          type: string
+      required:
+        - source_url
+        - target_url
+      additionalProperties: false
```
olevski, line 202: A few things:
1. I am not sure what "protocol-relative" means exactly. Like the protocol is not needed in the target url?
2. Providing examples on the Post and on the get response is a good idea imo.


```suggestion
    UrlRedirectPlanPost:
      description: Schema for creating a new redirect info
      type: object
      properties:
        source_url:
          description: The protocol-relative URL to redirect
          type: string
        target_url:
          description: The protocol-relative URL to redirect to
          type: string
      required:
        - source_url
        - target_url
      additionalProperties: false
      example:
        source_url: "https://gitlab.renkulab.io/user/project"
        target_url: "https://github.com/user/project"
```

components/renku_data_services/platform/orm.py
```diff
@@ -50,3 +52,42 @@ def dump(self) -> models.PlatformConfig:
...
             updated_at=self.updated_at,
         )
+
+
+class UrlRedirectsORM(BaseORM):
+    """The url redirects."""
+
+    __tablename__ = "url_redirects"
+
+    id: Mapped[ULID] = mapped_column("id", ULIDType, primary_key=True, default_factory=lambda: str(ULID()), init=False)
```
olevski, line 62: suggestion: it is even better if you add the server-side default for the id so that postgres will just make a ulid. It helps if you do mgirations or other changes. You will need to import `text` from `sqlalchemy` too at the top.

```suggestion
    id: Mapped[ULID] = mapped_column("id", ULIDType, primary_key=True, default_factory=lambda: str(ULID()), init=False, server_default=text("generate_ulid()"))
```

components/renku_data_services/platform/core.py
```diff
@@ -6,3 +6,11 @@
...
     """Validate the update to the platform configuration."""
     return models.PlatformConfigPatch(incident_banner=patch.incident_banner)
+
+
+def validate_url_redirect_post(post: apispec.UrlRedirectPlanPost) -> models.UnsavedUrlRedirectConfig:
+    """Validate the creation of a URL redirect."""
+    return models.UnsavedUrlRedirectConfig(
+        source_url=post.source_url,
+        target_url=post.target_url,
+    )
```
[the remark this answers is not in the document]
olevski: Also these validations should just be done on creation of redirects and on patching. For listing / serializing the responses we should not validate. We assume if it came from the DB it is good.

components/renku_data_services/platform/db.py
```diff
@@ -56,3 +61,112 @@ async def update_config(
...
+            stmt = select(schemas.UrlRedirectsORM).where(schemas.UrlRedirectsORM.source_url == source_url)
+            result = await session.execute(stmt)
+            url_redirect_orm: schemas.UrlRedirectsORM | None = result.scalars().first()
+
+            if not url_redirect_orm:
+                raise errors.MissingResourceError(
+                    message=f"A redirect for '{source_url}' does not exist or you do not have access to it."
+                )
+
+            # We do not currently check if the user is allowed to access the project, but we could...
```
olevski, line 116: I think we should not bother. Only if we have specific entity ids we can try to use this. With the current state we should just provide the service at the target url is responsible for doing authentication and authorization.

components/renku_data_services/platform/orm.py
```diff
@@ -50,3 +52,42 @@ def dump(self) -> models.PlatformConfig:
...
+
+    source_url: Mapped[str] = mapped_column("source_url", String(), default="", unique=True, index=True)
+    """The source URL for the redirect."""
+
+    target_url: Mapped[str] = mapped_column("target_url", String(), default="", index=True)
+    """The target URL for the redirect."""
+
```
olevski, line 69: Not sure what is the point with the defaults. Both of these have to be provided and they cannot be "" or NULL.


```suggestion
    source_url: Mapped[str] = mapped_column("source_url", String(), unique=True, index=True)
    """The source URL for the redirect."""

    target_url: Mapped[str] = mapped_column("target_url", String(), index=True)
    """The target URL for the redirect."""

```

components/renku_data_services/platform/api.spec.yaml
```diff
@@ -46,6 +46,100 @@ paths:
...
+            description: The url-encoded (original) source URL
+            schema:
+              type: string
+        requestBody:
+          required: true
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/UrlRedirectPlanPost"
+        responses:
```
olevski, line 132: A few issues here:
1. You are accepting both the source and target url for the patch. And they both have to be specified. What if you want to change just one of them. The convention we have been following is that every field of a patch request is optional. So we shuld have a specific schema for the patch request as well as a patch model and a dedicated validation function for the patching.
2. This is a bit related to 1. Currently the endpoint can be used like this:

```
PATCH /api/data/platform/redirects/<url1>
request body: {source_url: <url2>, target_url: <url3>}
```
This is really awkward. You are changing the identifier of the redirect. I would use a specific id for the redirect here and not have this awkwardness and to have a schema that allows me to change one or both of the parameters.

OR if we stick to passing the source url in the path for the patch request then I would only accept the target url in the payload.

```
PATCH /api/data/platform/redirects/<url1>
request body: { target_url: <url3>}
```

There is a 3rd alternative that may be clener, in that case the endpoints are like this:
- GET /api/data/platform/redirects - gets all redirects, have a query parameter that can specify a source_url and returns the redirects that are exact match on the source, if you are not an admin you have to provide the query parameter otherwise you get a 401 and if you are not an admin you only get 1 result in the list in the response
- POST /api/data/platform/redirects - create a new redirect (unchanged)
- PATCH /api/data/platform/redirects/<redirect-id>
- GET /api/data/platform/redirects/<redirect-id> (may not be needed)

components/renku_data_services/platform/blueprints.py
```diff
@@ -68,3 +73,90 @@ async def _patch_singleton_configuration(
             )
 
         return "/platform/config", ["PATCH"], _patch_singleton_configuration
+
+
+@dataclass(kw_only=True)
+class PlatformUrlRedirectBP(CustomBlueprint):
```
olevski, line 79: issue: We need to implement `DELETE`. Admins should be able to delete redirects. It will be likely rarely used but we should have it.

components/renku_data_services/platform/db.py
```diff
@@ -2,12 +2,17 @@
...
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from renku_data_services import base_models, errors
+from renku_data_services.authz.authz import Authz
+from renku_data_services.base_api.pagination import PaginationRequest
 from renku_data_services.platform import models
 from renku_data_services.platform import orm as schemas
+from renku_data_services.project.db import ProjectRepository
+
+# from renku_data_services.utils.core import with_db_transaction
```
olevski, line 15: question: leftover from other code? If not needed we should fully remove it.

components/renku_data_services/platform/db.py
```diff
@@ -56,3 +61,112 @@ async def update_config(
...
+        """Create a new URL redirect config."""
+        if user.id is None:
+            raise errors.UnauthorizedError(message="You do not have the required permissions for this operation.")
+        if not user.is_admin:
+            raise errors.ForbiddenError(message="You do not have the required permissions for this operation.")
+
+        async with self.session_maker() as session, session.begin():
+            existing = await session.scalar(
+                select(schemas.UrlRedirectsORM).where(schemas.UrlRedirectsORM.source_url == post.source_url)
+            )
```
olevski, line 131: question: Why not just reuse `get_redirect_config_by_source_url` in this and 1 or 2 other cases in ths same class?

test/bases/renku_data_services/data_api/test_platform_config.py
```diff
@@ -48,3 +50,160 @@ async def test_patch_platform_config_unauthorized(
...
+
+    assert res.status_code == 200, f"status code {res.status_code} != 200"
+    assert res.json is not None
+    assert res.headers["page"] == "1"
+    assert res.headers["per-page"] == "5"
+    assert res.headers["total"] == "0"
+    assert res.headers["total-pages"] == "0"
+
+    # redirects = res.json
+    # assert redirects.get("etag") != ""
```
olevski, line 69: reminder: remove or enable before merging

components/renku_data_services/platform/api.spec.yaml
ciyer: Fixed in 8448063860f8b36a0bc2423e161a2a2f78ef45f7

components/renku_data_services/platform/core.py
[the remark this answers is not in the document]
ciyer: I thought we could offer a bit more flexibility here, since only admins are allowed to register redirects. In particular, for project redirects, I was expecting the old and new URL references to

components/renku_data_services/platform/blueprints.py
ciyer: I was intending to do this in a follow-up PR when the admin UI for managing redirects is implemented, but I can do it here if that is preferred.

test/bases/renku_data_services/data_api/test_platform_config.py
ciyer: Fixed in c98ae3fbf297010f5a183dda2b27a20576e4f932

components/renku_data_services/platform/db.py
ciyer: Fixed in 4e2dd792d47e41c892c74dac61d7ef90940a9bac

PR #1001, opened 2025-09-02
feat: private images

eikek:
Adds support for private images from GitHub (GitLab was already working). 

- The image check at the repository api is done via a HEAD request
- Additional adds an endpoint to delete a connection (aka disconnect) from a client/provider
- use `auth:<base64 of user:token>` for pull secret format
- Rephrase error message when parsing an image url is not the expected result
- tested on github, gitlab.com and gitlab.datascience.ch

/deploy renku-ui=lorenzo/connected-private-images


leafty (commented):
Looks good, some minor comments here

leafty (approved):
LGTM

PR #1007, opened 2025-09-05
feat: add generic OIDC integration type

leafty:
Add a new possible value for the field `kind` on the `OAuth2Client` class in `connected_services`.

This allows Renku to connect to third-party platforms which implement OAuth 2.0 via OpenID Connect, e.g. with Keycloak.

This change is made so that we can connect RenkuLab to the FirecREST API at cscs.ch. The new type of client has also been tested here -> https://staging.dev.renku.ch/integrations.

Other changes:
* Update `models.py` so that we do not import from `apispec.py` into `db.py` or `orm.py`.

/deploy


components/renku_data_services/connected_services/core.py
```diff
@@ -20,16 +20,22 @@ def validate_oauth2_client_patch(patch: apispec.ProviderPatch) -> models.OAuth2C
...
 
 def validate_unsaved_oauth2_client(clnt: apispec.ProviderPost) -> models.UnsavedOAuth2Client:
     """Validate the the creation of a new OAuth2 Client."""
     if clnt.image_registry_url is not None:
         validate_image_registry_url(clnt.image_registry_url)
+    if clnt.oidc_issuer_url and models.ProviderKind(clnt.kind.value) != models.ProviderKind.generic_oidc:
+        raise errors.ValidationError(
+            message=f"The field 'oidc_issuer_url' can only be set when kind is set to {models.ProviderKind.generic_oidc.value}.",  # noqa E501
+            quiet=True,
+        )
```
olevski, line 35: issue: I think you need another check. If provider kind is oidc and the oidc issuer url is None.

olevski: Also please run oidc issuer url through url parse and check that we have at least a scheme (i.e. https) and domain name.

components/renku_data_services/connected_services/provider_adapters.py
```diff
@@ -232,12 +235,86 @@ def api_validate_account_response(self, response: Response) -> models.ConnectedA
...
+        request_url = urljoin(issuer_url, ".well-known/openid-configuration")
+        res = self.__get_httpx_client().get(request_url, headers=[("Accept", "application/json")])
+        if res.status_code != 200:
+            raise errors.ProgrammingError(message=f"Could not read OIDC issuer configuration from {request_url}")
+        res_json = res.json()
+        return {
+            "authorization_endpoint": res_json["authorization_endpoint"],
+            "token_endpoint": res_json["token_endpoint"],
+            "userinfo_endpoint": res_json["userinfo_endpoint"],
+        }
```
olevski, line 298: question: have you checked if we already use a library that can just do this instead of us having to code it ourselves?

leafty: I would rather not, considering that the OAuth 2.0 library we use has lead to spaghetti code and race conditions.

This is relatively simple to interpret and check against the relevant documentation. I can add a link to it here so that it is easy to maintain in the future.

olevski (dismissed):

leafty:
@olevski sorry, I pushed a small change here.

Also, I validated the alembic migration, it can be applied, reverted, then re-applied without issues.

PR #1008, opened 2025-09-05
chore: run sanic in single process mode

olevski:
This should significantly reduce memory consumption.

/deploy

### DIFF
--- a/bases/renku_data_services/data_api/main.py
+++ b/bases/renku_data_services/data_api/main.py
@@ -6,7 +6,6 @@ from os import environ
 from typing import TYPE_CHECKING, Any
 
 import sentry_sdk
-import uvloop
 from sanic import Request, Sanic
 from sanic.response import BaseHTTPResponse
 from sanic.worker.loader import AppLoader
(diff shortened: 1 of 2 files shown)


olevski:
The memory seems to have gone down from ~700MB to ~400MB.

bases/renku_data_services/data_api/main.py
```diff
@@ -200,4 +208,7 @@ async def logging_setup2(app: Sanic) -> None:
     loader = AppLoader(factory=create_app)
     app = loader.load()
     app.prepare(**args)
-    Sanic.serve(primary=app, app_loader=loader)
+    if os.name == "posix":
+        Sanic.serve(primary=app)
+    else:
+        Sanic.serve(primary=app, app_loader=loader)
```
leafty, line 214: Why is this dependent on the `os` and not on the value of `single_process`? With this PR, it seems that the user has no choice but to use `--single-process` to run the server.

leafty: Which is why we should set `spawn` or `fork` depending on the value of `--single-process`.

olevski: Sanic recommends somewhere in the docs to change the start method as early as possible right after you import Sanic. So we will be breaking that recommendation. But I can try.

PR #1012, opened 2025-09-12
Changes to support redirects from decommissioned URLs to live ones

ciyer:


components/renku_data_services/platform/api.spec.yaml
```diff
@@ -67,6 +179,64 @@ components:
         incident_banner:
           $ref: "#/components/schemas/IncidentBanner"
       additionalProperties: false
+    SourceUrl:
+      description: |
+        The URL that should be redirected. Must begin with `/projects` or
+        `https://gitlab.renkulab.io/`
+      type: string
+      examples:
```
olevski, line 187:
```suggestion
      example:
```

PR #1014, opened 2025-09-18
feat: update resource pools to support remote sessions

leafty:
Add support for resource pools which can support remote sessions on HPC.

Details:
* Add the `remote` field on the Resource Pool schemas in the API spec.
  * When present, the `remote` field contains the configuration used to start remote sessions
  * If the `remote` field is absent, then the sessions are started locally (`location: local` in Amalthea)
* This field is split into `remote_provider_id` and `remote_json` in the database
  * `remote_provider_id` field, type `str`. This allows the resource pool to be linked with an integration.
  * `remote_json` field, type `dict[str, Any]`. This field contains the rest of the remote configuration object.

Important: this PR only contains changes to the resource pool module, the changes related to sessions will be made in another PR.

Merging into #988

PR Stack:
* (base) #988
    * (this) #1014

/deploy renku=leafty/exp-dev-naming amalthea-sessions=leafty/git-repositories


components/renku_data_services/crc/core.py
```diff
@@ -38,3 +41,50 @@ def validate_cluster_patch(patch: apispec.ClusterPatch) -> models.ClusterPatch:
...
+        partition=body.partition,
+    )
+
+
+def validate_firecrest_api_url(url: str) -> None:
+    """Validate the URL to the FirecREST API."""
+    parsed = urlparse(url)
+    if not parsed.netloc:
+        raise errors.ValidationError(
+            message=f"The host for the image registry url {url} is not valid, expected a non-empty value.",
```
olevski, line 82:
```suggestion
            message=f"The host for the firecrest api url {url} is not valid, expected a non-empty value.",
```

components/renku_data_services/crc/core.py
```diff
@@ -38,3 +41,50 @@ def validate_cluster_patch(patch: apispec.ClusterPatch) -> models.ClusterPatch:
...
+            quiet=True,
+        )
+    accepted_schemes = ["https"]
+    if parsed.scheme not in accepted_schemes:
+        raise errors.ValidationError(
+            message=f"The scheme for the image registry url {url} is not valid, expected one of {accepted_schemes}",
```
olevski, line 88:
```suggestion
            message=f"The scheme for the firecrest api url {url} is not valid, expected one of {accepted_schemes}",
```

components/renku_data_services/crc/api.spec.yaml
```diff
@@ -1325,6 +1332,12 @@ components:
           $ref: "#/components/schemas/PublicFlag"
         default:
           $ref: "#/components/schemas/DefaultFlag"
+        remote:
+          $ref: "#/components/schemas/RemoteFlag"
+        remote_provider_id:
+          $ref: "#/components/schemas/RemoteProviderId"
+        remote_configuration:
+          $ref: "#/components/schemas/RemoteConfigurationPatch"
```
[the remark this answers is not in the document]
olevski: This way the `remote_configuration` can have a `configuration` or similar field that will be JSON and a dedicated field like for the oauth provider id.

olevski (changes requested):
Just a few questions and suggestions about the API

components/renku_data_services/crc/api.spec.yaml
```diff
@@ -1325,6 +1332,12 @@ components:
           $ref: "#/components/schemas/PublicFlag"
         default:
           $ref: "#/components/schemas/DefaultFlag"
+        remote:
+          $ref: "#/components/schemas/RemoteFlag"
+        remote_provider_id:
+          $ref: "#/components/schemas/RemoteProviderId"
```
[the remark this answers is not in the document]
leafty: Because it's not nice to have foreign keys in a JSON field.

components/renku_data_services/crc/api.spec.yaml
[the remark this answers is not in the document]
leafty: I don't want to create another level of indirection, especially when this is a one-to-one relation, not a one-to-many relation.

I will update the API to make the fields nicer to the API consumer.

components/renku_data_services/crc/models.py
```diff
@@ -390,3 +413,68 @@ def get_default_resource_class(self) -> ResourceClass | None:
...
+        res = asdict(self)
+        res["kind"] = self.kind.value
+        return res
+
+
+@dataclass(frozen=True, eq=True, kw_only=True)
+class RemoteConfigurationPatchReset:
+    """Model used to unset a remote configuration."""
+
+    pass
```
olevski, line 459: You dont need this. We already have a dedicated type such a case. Check `RESET` in `components/renku_data_services/base_models/core.py`. Instead of making this custom type just pass in this RESET object and use that.

olevski (changes requested):
Thanks Flora. Just one more thing about the reseting of the field. We already have a type for this purpose. So just reuse that.

components/renku_data_services/crc/models.py
leafty: Done in 68db9ef0.

PR #1017, opened 2025-09-24
chore: better error messages on http client errors

eikek:
Try for better error message for failing http calls

Unfortunately, the error message in the cause exception is not much
better. So there is now a `HttpClient` extending `AsyncClient` which
injects the request url into the error message and re-raises the
original error.

/deploy


components/renku_data_services/base_api/error_handler.py
```diff
@@ -132,6 +133,16 @@ def default(self, request: Request, exception: Exception) -> HTTPResponse:
                 )
             case jwt.exceptions.InvalidTokenError():
                 formatted_exception = errors.InvalidTokenError()
+
+            case httpx.RequestError():
+                formatted_exception = errors.BaseError(message=f"Error on remote connection: {exception}")
```
leafty, line 138: We don't need to create a new class from the httpx client.

```suggestion
            case httpx.RequestError():
                formatted_exception = errors.BaseError(message=f"Error while performing {exception.request.method} {exception.request.url}: {exception}")
```

This way we can log the request method and URL without having to modify the httpx classes.

Reference: https://www.python-httpx.org/exceptions/

leafty (changes requested):

components/renku_data_services/base_api/error_handler.py
eikek: Great, thank you. Somehow missed that page :-|

leafty: Then we have to manually handle the error.
1. By default we say there was an error performing an HTTP request. In the logs and in Sentry we can log the method and URL (and more if needed).
2. For endpoints were we know we are not leaking information we can try/catch the exception to give the user the method and URL.

leafty: But the client instances could be passed around, which can be hard to track precisely. Explicit handling at the caller site can look redundant but cannot escape to other parts of the code.

components/renku_data_services/base_api/error_handler.py
```diff
@@ -132,6 +133,17 @@ def default(self, request: Request, exception: Exception) -> HTTPResponse:
                 )
             case jwt.exceptions.InvalidTokenError():
                 formatted_exception = errors.InvalidTokenError()
+
+            case httpx.RequestError():
+                req_uri = exception.request.url
+                formatted_exception = errors.BaseError(message=f"Error on remote connection to {req_uri}: {exception}")
```
leafty, line 139: Note: should we also include the method here?

leafty (dismissed):

PR #1019, opened 2025-09-25
feat: Attempt to support dockerhub private images

eikek:
/deploy

### DIFF
--- a/components/renku_data_services/migrations/helper.py
+++ b/components/renku_data_services/migrations/helper.py
@@ -1,7 +1,5 @@
 """Helper functions for writing migrations."""
 
-from typing import cast
-
 from alembic import op
 from psycopg import sql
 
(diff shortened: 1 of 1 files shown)


components/renku_data_services/connected_services/api.spec.yaml
```diff
@@ -109,6 +109,35 @@ paths:
           $ref: "#/components/responses/Error"
       tags:
         - oauth2
+  /oauth2/providers/{provider_id}/simple_connect:
```
leafty, line 112: As a general remark: I think this API endpoint should be relocated somewhere else. It should not have `oauth2` in the path when this is to handle third-party platforms which DO NOT implement OAuth 2.0.

Some suggestions: `/integrations/...`, `/non-oauth2/...`, `/providers/...`, etc.

PR #1028, opened 2025-09-30
Check for image_registry_url when getting GH installations

eikek:
Also add some logging if things go wrong.

This adds a quick fix to allow private images from GitHub container registry. Unfortunately, GitHub doesn't allow access via tokens generated by "GitHub App". See details [here](https://github.com/orgs/community/discussions/24636) and the corresponding [open issue](https://github.com/github/roadmap/issues/558). Luckily it does work using a "GitHub OAuth App". The solution will be to add two integrations for GitHub - one is the GitHub App we are already using for accessing repositories. Then there is one GitHub OAuth App for accessing (private) images. In order to make this work, the user must still connect at github - so in case an integration is added with the ghcr.io domain, we must configure the provider adapter to connect to github.com for authorization. Additionally, there is no notion of "installations" for GitHub OAuth Apps, so the if is amended to check whether we have the "GitHub App" otherwise there will be an error returned from GitHub.

Implications for adding integrations:
- one with `url=https://github.com` which *must be* a GitHub App, no `image_registry_url`
- one with `url=https://github.com` which *must be* an _GitHub OAuth App_ with an `image_registry_url`

---

/deploy renku-ui=lorenzo/connected-private-images renku=ciyer/redirect-test-fixes

### DIFF
--- a/components/renku_data_services/connected_services/db.py
+++ b/components/renku_data_services/connected_services/db.py
@@ -466,7 +466,7 @@ class ConnectedServicesRepository:
             #       if no image_registry_url is specified
             if (
                 connection.client.kind == models.ProviderKind.github
-                and connection.client.url == "https://github.com"
+                and not connection.client.image_registry_url
                 and isinstance(adapter, GitHubAdapter)
             ):
                 request_url = urljoin(adapter.api_url, "user/installations")
--- a/components/renku_data_services/connected_services/db.py
+++ b/components/renku_data_services/connected_services/db.py
@@ -466,7 +466,7 @@ class ConnectedServicesRepository:
             #       if no image_registry_url is specified
             if (
                 connection.client.kind == models.ProviderKind.github
-                and not connection.client.image_registry_url
+                and connection.client.url == "https://github.com"
                 and isinstance(adapter, GitHubAdapter)
             ):
                 request_url = urljoin(adapter.api_url, "user/installations")
--- a/components/renku_data_services/connected_services/db.py
+++ b/components/renku_data_services/connected_services/db.py
@@ -474,7 +474,7 @@ class ConnectedServicesRepository:
                 try:
                     response = await oauth2_client.get(request_url, params=params, headers=adapter.api_common_headers)
                 except OAuthError as err:
-                    logger.warning(f"Error getting installations: {err}")
+                    logger.warning(f"Error getting installations at {request_url}: {err}")
                     if err.error == "bad_refresh_token":
                         raise errors.InvalidTokenError(
                             message="The refresh token for the connected service has expired or is invalid.",
@@ -484,7 +484,9 @@ class ConnectedServicesRepository:
                     raise
 
                 if response.status_code > 200:
-                    logger.warning(f"Could not get installations: {response.status_code} - {response.text}")
+                    logger.warning(
+                        f"Could not get installations at {request_url}: {response.status_code} - {response.text}"
+                    )
                     raise errors.UnauthorizedError(message="Could not get installation information.")
 
                 return adapter.api_validate_app_installations_response(response)
--- a/components/renku_data_services/connected_services/db.py
+++ b/components/renku_data_services/connected_services/db.py
@@ -461,13 +461,20 @@ class ConnectedServicesRepository:
             connection,
             adapter,
         ):
-            # NOTE: App installations are only available from GitHub
-            if connection.client.kind == models.ProviderKind.github and isinstance(adapter, GitHubAdapter):
+            # NOTE: App installations are only available from GitHub when using a "GitHub App"
+            #       it doesn't work for "OAuth App". Currently we guess handling a "Github App"
+            #       if no image_registry_url is specified
+            if (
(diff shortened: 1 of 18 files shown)


components/renku_data_services/connected_services/db.py
```diff
@@ -461,13 +461,20 @@ async def get_oauth2_app_installations(
...
             adapter,
         ):
-            # NOTE: App installations are only available from GitHub
-            if connection.client.kind == models.ProviderKind.github and isinstance(adapter, GitHubAdapter):
+            # NOTE: App installations are only available from GitHub when using a "GitHub App"
+            #       it doesn't work for "OAuth App". Currently we guess handling a "Github App"
+            #       if no image_registry_url is specified
+            if (
+                connection.client.kind == models.ProviderKind.github
+                and connection.client.url == "https://github.com"
```
leafty, line 469: This will break for GitHub Enterprise (the URL of the GitHub instance is customized for GitHub Enterprise).

I think it is more robust solution here is to test again `image_registry_url`.

We have one GitHub Enterprise integration working in production:
Image
```json
{
        "id": "github.zhaw.ch",
        "kind": "github",
        "app_slug": "zhaw-github",
        "client_id": "redacted",
        "client_secret": "redacted",
        "display_name": "ZHAW GitHub",
        "scope": "",
        "url": "https://github.zhaw.ch",
        "use_pkce": false
    }
```

leafty (changes requested):
Discussion: Why don't we use the `image_registry_url` as a temporary marker for the difference between the GitHub App and the OAuth App?

For GitHub.com, we would have two providers configured:
* One with `url=https://github.com` and no `image_registry_url`
* One with `url=https://github.com` and `image_registry_url=https://ghcr.io`

Maybe the long term direction is to add a new provider type, e.g. `github_oauth_app` so that knowing the difference is simpler.

eikek:
> Maybe the long term direction is to add a new provider type, e.g. `github_oauth_app` so that knowing the difference is simpler.

Oh yes, I would definitely favor such a solution for later. It is very awkward to base these decisions on a empty or non-empty `image_registry_url`.

eikek:
Yes, we really need something to capture this intent. For now, you suggest to rather do the checks in the other places (repositories) - not only in the connected_services part of the code? I thought it might be nicer to keep it as close together as possible.

leafty:
> Yes, we really need something to capture this intent. For now, you suggest to rather do the checks in the other places (repositories) - not only in the connected_services part of the code? I thought it might be nicer to keep it as close together as possible. I would rather overwrite the `client_url` and not touch any other code to make up for that "fix". But it's not a strong opinion either. (sorry for the many updates on that comment 😅)

The `client_url` will also not work if we create an OAuth App on a GitHub Enterprise server, as the registry URL may not match `ghcr.io`.

eikek:
> The `client_url` will also not work if we create an OAuth App on a GitHub Enterprise server, as the registry URL may not match `ghcr.io`.

Right, we can have many GitHubs with different urls 🤦🏼. ok, then we need to use same URLs but once with registry and once without. And making sure the correct one is an oauth app. Then `provider_kind==github and image_registry_url` means OAUth App.

leafty:
It should simple enough to add the new fields to the `OAuth2Provider` class used to parse the json response.

olevski:
Should we just switch to using the github oauth app for both code and image pulls? And not use the "regular" github app at all? Assuming that the github oauth app can handle both code pulls/pushes and image pulls?

leafty:
> Should we just switch to using the github oauth app for both code and image pulls? And not use the "regular" github app at all? Assuming that the github oauth app can handle both code pulls/pushes and image pulls?

I think we should not:
* https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/rate-limits-for-oauth-apps -> Rate limits are much better for GitHub Apps
* https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/differences-between-github-apps-and-oauth-apps#git-access -> using the OAuth app tokens for git is more complex

components/renku_data_services/connected_services/utils.py
```diff
@@ -2,9 +2,41 @@
...
+logger = logging.getLogger(__name__)
 
 
 def generate_code_verifier(size: int = 48) -> str:
     """Returns a randomly generated code for use in PKCE."""
     rand = random.SystemRandom()
     return base64.b64encode(rand.randbytes(size)).decode()
+
+
+class GitHubProviderType(StrEnum):
```
leafty, line 22: Should we add a TODO here that says "TODO: add a new type of integration 'github_oauth' instead of running this logic".

This would remove the need for `get_github_provider_type` and instead just check the integration type.

leafty (commented):
Looks, good some comment below.

Also, check that some commented out code is properly removed or has a comment attached.

components/renku_data_services/connected_services/utils.py
eikek: I added this todo to the function. Hope that is fine. I'd like to do the mentioned change later.

PR #1029, opened 2025-10-01
chore: add Salim and Samuel to codeowners

olevski:
This allows their reviews on PRs to protected branches to count.

I cannot add them to the python maintainers group because then I will have to add them to the SDSC org fully. And I think that causes other permissions to be assigned to them - i.e. like k8s access.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,3 @@
-*           @SwissDataScienceCenter/renku-python-maintainers
+*	@SwissDataScienceCenter/renku-python-maintainers
+*	@sgaist
+*	@SalimKayal


.github/CODEOWNERS
```diff
@@ -1 +1,3 @@
-*           @SwissDataScienceCenter/renku-python-maintainers
+*	@SwissDataScienceCenter/renku-python-maintainers
+*	@sgaist
+*	@SalimKayal
```
leafty, line 3: Is this possible?

```suggestion
*	@SwissDataScienceCenter/renku-python-external-maintainers
```

leafty (approved):
Can we have

PR #1034, opened 2025-10-02
wip: let admins retrieve all migrations

ciyer:


ciyer:
Retested, and @leafty is correct. I'm guessing the situation I saw was due to some information being out of sync immediately after promoting a user to admin. Closing.

PR #1059, opened 2025-10-13
fix: patching of session custom resources

olevski:
/deploy

### DIFF
--- a/components/renku_data_services/base_models/core.py
+++ b/components/renku_data_services/base_models/core.py
@@ -7,8 +7,9 @@ import unicodedata
 from dataclasses import dataclass, field
 from datetime import datetime
 from enum import Enum, StrEnum, auto
-from typing import ClassVar, Never, Optional, Protocol, Self, TypeVar, overload
+from typing import Any, ClassVar, Never, Optional, Protocol, Self, TypeVar, overload
 
+from pydantic_core import core_schema
 from sanic import Request
 
 from renku_data_services.errors import errors
@@ -421,6 +422,18 @@ class ResetType(Enum):
 
     def __str__(self) -> str:
         return "RESET"
+
+    @classmethod
+    def __get_pydantic_core_schema__(cls, *args: Any, **kwargs: Any) -> core_schema.AfterValidatorFunctionSchema:
+        """This enables pydantic to serialize this as None/null."""
+        return core_schema.no_info_after_validator_function(
+            cls,
+            core_schema.int_schema(),
+            serialization=core_schema.plain_serializer_function_ser_schema(
+                lambda _: None,
+                when_used="always",
+            ),
+        )
 
 
 RESET = ResetType.Reset
--- a/components/renku_data_services/notebooks/crs.py
+++ b/components/renku_data_services/notebooks/crs.py
@@ -14,7 +14,7 @@ from pydantic import BaseModel, Field, field_serializer, field_validator, model_
 from pydantic.types import HashableItemType
 from ulid import ULID
 
-from renku_data_services.base_models.core import ResetType
+from renku_data_services.base_models.core import RESET, ResetType
 from renku_data_services.errors import errors
 from renku_data_services.notebooks import apispec
 from renku_data_services.notebooks.constants import AMALTHEA_SESSION_GVK, JUPYTER_SESSION_GVK
@@ -171,6 +171,8 @@ class CullingDurationParsingMixin(BaseCRD):
     def __serialize_duration_field(self, val: Any, nxt: Any, _info: Any) -> Any:
         if isinstance(val, timedelta):
             return format_duration(val)
+        if val is RESET:
+            return None
         return nxt(val)
 
     @field_validator("*", mode="wrap")


components/renku_data_services/base_models/core.py
```diff
@@ -400,9 +401,11 @@ async def authenticate(self, access_token: str, request: Request) -> AnyAPIUser:
         ...
 
 
-ResetType = NewType("ResetType", object)
+__ResetType = NewType("__ResetType", object)
+ResetType = Annotated[__ResetType, PlainSerializer(lambda _: None, return_type=None)]
```
leafty: Why not?
```python
ResetType = Annotated[NewType("__ResetType", object), PlainSerializer(lambda _: None, return_type=None)]
```

test/bases/renku_data_services/data_api/utils.py
```diff
@@ -142,3 +143,14 @@ def setup_amalthea(install_name: str, app_name: str, version: str, cluster: Kind
             break
     else:
         raise AssertionError("Timeout waiting on amalthea to run") from None
+
+
+def create_api_user_from_user_info(user: UserInfo, access_token: str, is_admin: bool = False) -> APIUser:
```
leafty: Shouldn't this method be kept in the test folders if it is only used in tests?

components/renku_data_services/notebooks/core_sessions.py
```diff
@@ -1030,29 +1073,39 @@ async def patch_session(
...
+        except (errors.MissingResourceError, errors.UnauthorizedError, errors.ForbiddenError):
+            old_rp = None
         rc = rp.get_resource_class(body.resource_class_id)
         if not rc:
             raise errors.MissingResourceError(
                 message=f"The resource class you requested with ID {body.resource_class_id} does not exist"
             )
-        # TODO: reject session classes which change the cluster
+        if old_rp is not None and rp.cluster != old_rp.cluster:
+            raise errors.ValidationError(message="Changing resource pools with different clusters is not allowed.")
```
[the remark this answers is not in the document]
olevski: Removed.

components/renku_data_services/base_models/core.py
leafty: I guess it's one of those cases where assigning does things.

components/renku_data_services/notebooks/crs.py
leafty: Just a note: I wonder if we could write some form of metaclass to derive a class where all fields are reset-able.

```python
class Example:
    """Generated from the OpenAPI"""
    some_field: str | None = None

class Resetable:
    """Some magic meta class"""
    pass

ResetableExample = Resetable(Example)
# Would be equivalent to:
# class ResetableExample:
#     some_field: str | ResetType | None = None
```

components/renku_data_services/notebooks/utils.py
```diff
@@ -128,6 +130,21 @@ def node_affinity_from_resource_class(
     return affinity
 
 
+def node_affinity_patch_from_resource_class(
```
leafty: Note: I think this may not work when there is no default affinity for sessions.

components/renku_data_services/notebooks/utils.py
```diff
@@ -128,6 +130,21 @@ def node_affinity_from_resource_class(
...
+    resource_class: crc_models.ResourceClass, default_affinity: Affinity
+) -> AffinityPatch:
+    """Create a patch for the session affinity."""
+    affinity = node_affinity_from_resource_class(resource_class, default_affinity)
+    patch = AffinityPatch(nodeAffinity=RESET, podAffinity=RESET, podAntiAffinity=RESET)
+    if affinity.nodeAffinity:
+        patch.nodeAffinity = affinity.nodeAffinity
```
leafty: Node affinity is not cleared when the service is deployed without a default node affinity.

leafty (changes requested):

components/renku_data_services/notebooks/utils.py
leafty: Tested here: https://github.com/SwissDataScienceCenter/renku/pull/4235

components/renku_data_services/notebooks/crs.py
olevski: Yeah probably. But l dont want to do this now. We can add this in the future.

components/renku_data_services/notebooks/crs.py
```diff
@@ -375,36 +384,61 @@ def base_url(self) -> str | None:
...
-    annotations: dict[str, str] | None = None
+    annotations: dict[str, str | ResetType] | ResetType | None = None
+
+
+class AffinityPatch(BaseCRD):
+    """Patch for the affinity of a session."""
+
+    nodeAffinity: NodeAffinity | ResetType | None = None
+    podAffinity: PodAffinity | ResetType | None = None
+    podAntiAffinity: PodAntiAffinity | ResetType | None = None
```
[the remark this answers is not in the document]
olevski: I propagated the patching for affinities all the way through now.

components/renku_data_services/notebooks/utils.py
olevski: I propagated the affinity patching all the way through now. So it should work. I will test it tomorrow.
see https://github.com/SwissDataScienceCenter/renku-data-services/pull/1059/commits/08e3941556960c2cba99e2f3bacad8b46358c72a

components/renku_data_services/notebooks/config/dynamic.py
```diff
@@ -456,8 +461,8 @@ def _for_testing(cls) -> Self:
         )
 
     @property
-    def affinity_model(self) -> Affinity:
-        return Affinity.model_validate(self.affinity)
+    def affinity_model(self) -> Affinity | None:
+        return Affinity.model_validate(self.affinity) if self.affinity is not None else None
```
leafty: Why not return the empty `Affinity()` instance when there is no specific affinity configured? This way you do not have to update all the types in this commit.

leafty (approved):
LGTM, just see the comment above as a suggestion.

leafty (approved):

Issue #1060, opened 2025-10-13
Session secret is missing (sometimes)


olevski:
This happens on some sessions or intermittently. It prevents the session from starting back up.

The session pod keeps restarting and this shows up in the events for the pod:
```
Events:                                                                                                                                             
│   Type     Reason       Age                    From     Message                                                                                     
│   ----     ------       ----                   ----     -------                                                                                     
│   Warning  FailedMount  4m12s (x72 over 134m)  kubelet  MountVolume.SetUp failed for volume "<session-id>-secrets-volume" : secret "<session-id>-secrets" not found
```

This secret contains encrypted secrets that are then decrypted. This decryption happens in the `init-user-secrets` init container. The decrypted secrets are then mounted in the session in the `/secrets` path.

Potentially this can occur if someone:
1. Start a session which has user secrets
2. Hibernate the session
3. Remove all users secrets
4. Restart the session

olevski:
Will close this in favor of concurrent execution guard on the POST endpoint for sessions.

PR #1063, opened 2025-10-14
fix: make secret service patch on conflict

olevski:
/deploy

This should partially help with https://github.com/SwissDataScienceCenter/renku-data-services/issues/1060

I think that the secret service should be idempotent when it comes to creating secrets that  are requested from it. Currently if the secret already exists it will just fail with a 409.

An even simpler alternative to this is to not fail if the secret exists but then also not patch. I think this is risky because if we dont patch then we could be stumbling upon a very old secret whose data is completely useless. This way there is a higher chance that the data in the secret will be valid.


components/renku_data_services/k8s/models.py
```diff
@@ -161,6 +161,30 @@ def to_v1_secret(self) -> V1Secret:
...
+            )
+        if "annotations" in self.manifest.metadata:
+            patch.append(
+                {"op": "replace", "path": "/metadata/annotations", "value": self.manifest.metadata.annotations},
+            )
+        if "ownerReferences" in self.manifest.metadata:
+            patch.append(
+                {"op": "replace", "path": "/metadata/ownerReferences", "value": self.manifest.metadata.ownerReferences},
+            )
+        return patch
```
leafty:
```suggestion
        # We never create 'finalizers' nor 'managedFields', so we do not patch them.
        return patch
```

components/renku_data_services/secrets/blueprints.py
```diff
@@ -44,6 +45,14 @@ async def _post(_: Request, user: base_models.APIUser, body: apispec.K8sSecret)
...
             try:
                 result = await self.client.create_secret(secret)
+            except kr8s.ServerError as e:
+                if not e.response:
+                    raise errors.SecretCreationError(
+                        message=f"An error occurred creating secrets: {str(type(e))}"
+                    ) from None
+                if e.response.status_code == 409:
+                    # NOTE: It means that the secret already exists, so we try to patch
+                    result = await self.client.patch_secret(secret, patch=secret.to_patch())
```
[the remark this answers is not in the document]
olevski: I added the line to do this. But sentry is just not added at all for the secret service. So it would not work. I will do followup prs for adding sentry in all the other components - i.e. the k8s watcher and the secret svc.

components/renku_data_services/k8s/models.py
olevski: done in https://github.com/SwissDataScienceCenter/renku-data-services/pull/1063/commits/4ffba7bea2c90b9fc71472abeb75bb341d8f8d36

leafty (approved):

leafty (approved):

PR #1065, opened 2025-10-15
feat: allow data connectors to be skipped when launching sessions

### LINKED ISSUE #1009: Data connectors in sessions: allow skip
Change the behavior of the sessions launch API:

* Add a `skipped` field which specifies if the data connector should be mounted (defaults to `false`).
* This makes the overrides much easier to handle -> if `skipped` then do not mount the DC
* Still mount the project's DCs by default

leafty:
Closes #1009.

⚠️ **Includes API breaking changes** ⚠️

Allows data connectors to be skipped when launching a session.

UI update: https://github.com/SwissDataScienceCenter/renku-ui/pull/3873.

Note: more related changes can be found in #1068.

/deploy renku-ui=leafty/feat-allow-dc-skip


components/renku_data_services/notebooks/blueprints.py
```diff
@@ -219,9 +220,11 @@ async def _handler(
             internal_gitlab_user: APIUser,
             body: apispec.SessionPostRequest,
         ) -> JSONResponse:
+            launch_request = validate_session_post_request(body=body)
             session, created = await start_session(
                 request=request,
-                body=body,
+                # body=body,
+                launch_request=launch_request,
```
olevski:
```suggestion
                launch_request=launch_request,
```

PR #1070, opened 2025-10-20
chore: re-enable notebooks tests and use kind

olevski:
/deploy

### DIFF
--- a/test/bases/renku_data_services/data_api/test_search.py
+++ b/test/bases/renku_data_services/data_api/test_search.py
@@ -28,6 +28,8 @@ from test.bases.renku_data_services.data_api.conftest import (
 )
 
 
+# TODO: figure out how to run search tests fully parallel
+@pytest.mark.xdist_group("search")
 @pytest.mark.asyncio
 async def test_direct_member_search(
     create_user: CreateUserCall,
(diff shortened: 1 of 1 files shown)


test/bases/renku_data_services/data_api/conftest.py
```diff
@@ -727,3 +736,53 @@ def __make_headers(user: UserInfo, admin: bool = False) -> dict[str, str]:
...
+def cluster_name():
+    return f"k8s-cluster-{str(ULID()).lower()}"
+
+
+@pytest.fixture(scope="session")
+def cluster(cluster_name, monkeysession):
+    cluster = KindCluster(cluster_name)
+    monkeysession.setenv("KUBECONFIG", cluster.kubeconfig)
+    with cluster:
+        yield cluster
```
sgaist: Would this be a bit more Pythonic ?
```suggestion
    monkeysession.setenv("KUBECONFIG", cluster.kubeconfig)
    with KindCluster(cluster_name) as cluster:
        yield cluster
```

olevski: I dont understand why we are so hung up on this issue which has zero consequence. 

The `KindCluster` copies the env variables at initialization, stores that in `self.env` at initialization and then uses `self.env` when it runs commands. So if I do not have `KUBECONFIG` set to the right thing at `KindCluster` intialization things will not work.]


This is what I did with the addition of a separate fixture that returns the kubeconfig and sets the env variable.:
```
with KindCluster(cluster_name, kubeconfig=kubeconfig) as cluster:
        yield cluster
```

PR #1074, opened 2025-10-23
build: set fixed postgres version in devcontainer

leafty:
This fixes the current issue where the devcontainer is not starting in GitHub actions.

Version 16 is used, as this is what we deploy Renku with.


sambuc (approved):
LGTM

PR #1089, opened 2025-10-29
fix: namespace-is  searches in namespace and its children

eikek:
When searching `namespace:renku` only entities directly attached to that namespace where returned. Now it searches also the child namespaces. This fixes the issue on the group page, where data connectors inside projects (where projects are inside the group) were not listed.

/deploy


leafty (approved):
LGTM, thanks @eikek ! 👏

PR #1092, opened 2025-10-31
fix: do not apply hardcoded culling defaults

olevski:
We were adding a default for culling (hardcoded from the old notebooks config) - even when culling should not happen at all according to the resource pool. With this change we respect explicitly what the resource pool says. And if a threshold is not set in the pool then the sessions will not be culling for that scenario.

It also adds the `RESET` type for the culling thresholds in a resource pool. Although these worked fine even without its addition. But this way things are more in line with the rest of the code base.

/deploy

### DIFF
--- a/.github/workflows/acceptance-tests.yml
+++ b/.github/workflows/acceptance-tests.yml
@@ -64,7 +64,7 @@ jobs:
           password: ${{ secrets.RENKU_DOCKER_PASSWORD }}
 
       - name: Find deployment url
-        uses: peter-evans/find-comment@v3
+        uses: peter-evans/find-comment@v4
         id: deploymentUrlMessage
         with:
           issue-number: ${{ github.event.pull_request.number }}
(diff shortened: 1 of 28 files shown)


components/renku_data_services/notebooks/core_sessions.py
```diff
@@ -509,21 +509,24 @@ def get_culling(
...
         # it is hibernated.
-        hibernation_threshold_seconds = 1
+        hibernation_threshold: timedelta | None = timedelta(seconds=1)
     else:
-        hibernation_threshold_seconds = (
-            resource_pool.hibernation_threshold or nb_config.sessions.culling.registered.hibernated_seconds
+        hibernation_threshold = (
+            timedelta(seconds=resource_pool.hibernation_threshold)
+            if resource_pool.hibernation_threshold is not None
+            else None
```
olevski: This was the main problem. We should not be falling back to `nb_config.culling.x` if the resource pool does not set a limit.

leafty (approved):
LGTM

PR #1094, opened 2025-11-03
feat: add user alerts

wesjdj:
# Release Notes

## User-Facing Changes

**🌟 New Features**

- 🐸 **Data services**:

**✨ Improvements**

- 🐸 **Data services**:

**🐞 Bug Fixes**

- **Data services**:

## Internal Changes

**Improvements**

- **Data services**:

**Bug Fixes**

- **Data services**:

# Internal/Miscellaneous

/deploy #notest

### DIFF
--- a/components/renku_data_services/migrations/versions/5ec28ea89e0a_add_alerts_table.py
+++ b/components/renku_data_services/migrations/versions/5ec28ea89e0a_add_alerts_table.py
@@ -1,7 +1,7 @@
 """add alerts table
 
 Revision ID: 5ec28ea89e0a
-Revises: d437be68a4fb
+Revises: 42049656cdb8
 Create Date: 2025-11-05 13:41:25.972261
 
 """


components/renku_data_services/notifications/api.spec.yaml
```diff
@@ -0,0 +1,99 @@
...
+      type: object
+      properties:
+        status:
+          $ref: "#/components/schemas/AlertStatus"
+        alerts:
+          type: array
+          description: The list of alerts.
+          items:
+            type: string
+          minItems: 1
```
leafty: This part of the API is not conventional: a POST should create a single object. Here we can create multiple alerts at once.

components/renku_data_services/notifications/api.spec.yaml
```diff
@@ -0,0 +1,99 @@
...
+      description: Data required to create an alert.
+      type: object
+      properties:
+        status:
+          $ref: "#/components/schemas/AlertStatus"
+        alerts:
+          type: array
+          description: The list of alerts.
+          items:
+            type: string
```
leafty: Note: This payload seems too minimalist for now. It can be updated as we get a better understanding of what is an alert object.

components/renku_data_services/notifications/api.spec.yaml
```diff
@@ -0,0 +1,99 @@
...
+          type: string
+          format: date-time
+          description: The time the alert started.
+        endsAt:
+          type: string
+          format: date-time
+          description: The time the alert ended.
+        fingerprint:
+          type: string
+          description: The fingerprint of the alert.
```
leafty: What is a fingerprint? In `data-services` we identify objects using the `id` field, which is usually a ULID.

components/renku_data_services/notifications/api.spec.yaml
```diff
@@ -0,0 +1,99 @@
...
+        type: string
+    AlertAnnotations:
+      type: object
+      description: Annotations associated with the alert.
+      additionalProperties:
+        type: string
+    Alert:
+      description: An alert sent to a user.
+      type: object
+      properties:
```
leafty: Potentially missing properties:
* `event_type`: the `EventType` of an alert (or notification). Some examples: `session_close_to_memory_limit`, `session_out_of_memory`, `session_cannot_start`, etc.
* `title`: the alert title, a short title of the alert, e.g. "Session is out of memory".
* `message`: message to display to the user, giving more information about the alert. E.g. "Session 'xyz' is close to running out of memory. Try to free up memory or the session may be forcefully restarted."
* `session_name`: the name (identity) of the session related to the alert (field can be `null`)

components/renku_data_services/notifications/api.spec.yaml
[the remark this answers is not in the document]
wesjdj: Thank you @leafty, that all makes sense to me

leafty: Good start :+1:

components/renku_data_services/notifications/orm.py
```diff
@@ -0,0 +1,62 @@
...
+    id: Mapped[ULID] = mapped_column("id", ULIDType, primary_key=True, default=lambda: str(ULID()), init=False)
+    """ID of the alert."""
+
+    title: Mapped[str] = mapped_column("title", String(), nullable=False)
+    """Title of the alert."""
+
+    message: Mapped[str] = mapped_column("message", String(), nullable=False)
+    """Message of the alert."""
+
+    user_id: Mapped[str] = mapped_column("user_id", String(), nullable=False)
```
leafty: If possible we should have a foreign key constraint for `user_id`. Otherwise we can add alerts for users that do not exist.


```suggestion
    user_id: Mapped[str] = mapped_column(ForeignKey(UserORM.keycloak_id, ondelete="CASCADE"), nullable=False, index=True)
```

with

```python
from renku_data_services.users.orm import UserORM
```

components/renku_data_services/notifications/orm.py
```diff
@@ -0,0 +1,62 @@
...
+    """ID of the user the alert is for."""
+
+    session_name: Mapped[Optional[str]] = mapped_column("session_name", String(), default=None)
```
leafty:
```suggestion
    session_name: Mapped[str | None] = mapped_column("session_name", String(), default=None)
```

components/renku_data_services/notifications/api.spec.yaml
```diff
@@ -0,0 +1,204 @@
...
+    Ulid:
+      description: ULID identifier
+      type: string
+      minLength: 26
+      maxLength: 26
+      pattern: "^[0-7][0-9A-HJKMNP-TV-Z]{25}$" # This is case-insensitive
+    Alert:
+      description: An alert sent or displayed to a user.
+      type: object
+      properties:
```
leafty: Note, can be updated in a later PR: I think that maybe we could have a `status: firing/resolved` field. This way, patching an alert would use clean REST semantics (that is patching an existing field with an allowed value).

components/renku_data_services/notifications/db.py
```diff
@@ -0,0 +1,98 @@
...
+            raise errors.ForbiddenError(message="You do not have the required permissions for this operation.")
+
+        async with self.session_maker() as session:
+            query = (
+                select(schemas.AlertORM)
+                .where(schemas.AlertORM.user_id == alert.user_id)
+                .where(schemas.AlertORM.title == alert.title)
+                .where(schemas.AlertORM.message == alert.message)
+                .where(schemas.AlertORM.session_name == alert.session_name)
+                .where(schemas.AlertORM.resolved_at.is_(None))
```
leafty: Discussion: with this logic, if I post an alert with just a slightly different message, then I can add as many alerts on the same session as I want.

To avoid this, I think it could be useful to have a field called `event_type` which is used to specify the type of alert, e.g. `session_oom` or `session_memory_limit`, etc. Then we can de-duplicate based on `user_id, event_type, session_name`.

components/renku_data_services/notifications/db.py
```diff
@@ -0,0 +1,102 @@
...
+        async with self.session_maker() as session, session.begin():
+            user_exists = await session.scalar(select(UserORM.keycloak_id).where(UserORM.keycloak_id == alert.user_id))
+            if user_exists is None:
+                raise errors.ValidationError(message=f"User with ID '{alert.user_id}' does not exist.")
+
+            query = (
+                select(schemas.AlertORM)
+                .where(schemas.AlertORM.user_id == alert.user_id)
+                .where(schemas.AlertORM.event_type == alert.event_type)
+                .where(schemas.AlertORM.session_name == alert.session_name)
```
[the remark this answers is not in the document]
leafty: OK, it's working as intended.

PR #1095, opened 2025-11-04
Feat: Improve repository response 

eikek:
Improve working with repositories:

- remove endpoints accepting the url as a url path segment and use a query parameter. Additional to the weird looking urls, there was also a problem with route matching in sanic. If a url was passed as a argument with a trailing slash (even when percent encoded), sanic would try to find a route, which of course resulted in a 404
- The `/probe` endpoint has been subsumed by the `repositories?url=…` endpoint and is therefore removed
- When servicing the repository metadata, data that we know (a provider, a connection) is still returned so the ui is able to present different (error) scenarios
- includes PR #1114 fixing refreshing tokens
- introduce a `git_url.py` module doing more checks on urls denoting git repositories

/deploy renku-ui=lorenzo/code-intergations renku=build/code-integrations

### DIFF
--- a/components/renku_data_services/connected_services/oauth_http.py
+++ b/components/renku_data_services/connected_services/oauth_http.py
@@ -247,7 +247,6 @@ class _SafeAsyncOAuthClient(AsyncOAuth2Client):  # type: ignore[misc]
                     self.token = updated_token
 
 
-
 class DefaultOAuthClient(OAuthHttpClient):
     """The default oauth-http client."""
 


components/renku_data_services/connected_services/oauth_http.py
```diff
@@ -0,0 +1,611 @@
...
+    invalid_user = "invalid_user"
+    no_client = "no_client"
+    no_state = "no_state"
+
+
+class OAuthHttpError(StrEnum):
+    """Errors possible when using the client."""
+
+    oauth_error = "oauth_error"  # nosec: B105
+    unauthorized = "unauthorized"  # nosec: B105
```
olevski: I just reran the checks for this without the comments and it passed.

```suggestion
    oauth_error = "oauth_error"
    unauthorized = "unauthorized"
```

components/renku_data_services/connected_services/oauth_http.py
```diff
@@ -0,0 +1,611 @@
...
+    ) -> models.OAuth2TokenSet | None:
+        """Check the database for a recently updated token and return it."""
+        ...
+
+    def get_lock(self, id: ULID) -> ConnectionLock:
+        """Return a lock scoped by a connection id."""
+        ...
+
+
+class _SafeAsyncOAuthClient(AsyncOAuth2Client):  # type: ignore  # nosec: B107
```
olevski: Nosec is not needed, and the type ignore can be more restrictive. Tested this with mypy locally. It works.

```suggestion
class _SafeAsyncOAuthClient(AsyncOAuth2Client):  # type: ignore[misc]
```

components/renku_data_services/connected_services/oauth_http.py
```diff
@@ -0,0 +1,611 @@
...
+        client_secret: str | None = None,
+        token_endpoint_auth_method=None,
+        revocation_endpoint_auth_method=None,
+        scope=None,
+        redirect_uri=None,
+        token=None,
+        token_placement="header",  # nosec: B107
+        leeway=60,
+        **kwargs,
+    ):
```
olevski: We can add type hints here, also the top level nosec is not needed. Tested locally with mypy and bandit (for nosec)


```suggestion
    def __init__(
        self,
        client_id: str,
        connection_id: ULID,
        token_check: _TokenCheck,
        client_secret: str | None = None,
        token_endpoint_auth_method: str | None = None,
        revocation_endpoint_auth_method: str | None = None,
        scope: str | None = None,
        redirect_uri: str | None = None,
        token: OAuth2TokenSet | None = None,
        token_placement: str = "header",  # nosec: B107
        leeway: int = 60,
        **kwargs: Any,
    ):
```

components/renku_data_services/repositories/api.spec.yaml
```diff
@@ -8,18 +8,23 @@ info:
 servers:
   - url: /api/data
 paths:
-  /repositories/{repository_url}:
+  /repositories:
```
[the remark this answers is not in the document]
eikek: I agree. I'll update the path.

components/renku_data_services/repositories/apispec_base.py
```diff
@@ -15,13 +15,13 @@ class Config:
...
+    @field_validator("id", mode="before", check_fields=False)
     @classmethod
-    def serialize_connection_id(cls, connection_id: str | ULID | None) -> str | None:
+    def serialize_connection_id(cls, id: str | ULID | None) -> str | None:
         """Custom serializer that can handle ULIDs."""
-        if connection_id is None:
+        if id is None:
             return None
-        return str(connection_id)
+        return str(id)
```
leafty: We have a more generic version of this which handles ULIDs in any field.

components/renku_data_services/repositories/blueprints.py
```diff
@@ -32,45 +29,64 @@ def get_one_repository(self) -> BlueprintFactoryResponse:
...
             internal_gitlab_user: base_models.APIUser,
-            repository_url: str,
             etag: str | None,
         ) -> JSONResponse | HTTPResponse:
-            repository_url = unquote(repository_url)
-            RepositoryParams.model_validate(dict(repository_url=repository_url))
+            query_args: dict[str, str] = req.get_args() or {}
+            repository_url = query_args.get("url")
+            if repository_url is None:
+                return HTTPResponse(status=404)
```
leafty: Why is this not a 422? The request is essentially invalid in this case.

test/components/renku_data_services/repositories/test_git_url.py
```diff
@@ -0,0 +1,71 @@
...
+    "abc",
+    "http://",
+    "localhost/repo",
+    "ftp://test.com",
+    "http://localhost:3000/my/repo",
+    "http://127.0.0.1:3000/my/repo",
+    "http://127.0.0.1/my/repo",
+    "http://localhost/my/repo",
+]
+good_urls = [
```
leafty: valid_urls

test/bases/renku_data_services/data_api/test_repositories.py
```diff
@@ -111,20 +114,20 @@ async def test_get_one_repository(oauth2_test_client: SanicASGITestClient, user_
...
-    assert result.get("provider_id") == "provider_1"
-    assert result.get("repository_metadata") is not None
-    repository_metadata = result["repository_metadata"]
-    assert repository_metadata.get("git_http_url") == repository_url
-    assert repository_metadata.get("permissions") is not None
-    permissions = repository_metadata["permissions"]
-    assert permissions.get("pull")
-    assert not permissions.get("push")
+    assert result["connection"]["id"] == connection["id"]
+    assert result["provider"]["id"] == "provider_1"
```
leafty: You should use `.get` so that it's the `assert` which fails, not the get attribute on NoneType or similar if the response is malformed.

components/renku_data_services/repositories/apispec_base.py
eikek: I think, this was already there - can you tell me where to look for this generic version? Or not sure what you mean.

components/renku_data_services/repositories/git_url.py
```diff
@@ -0,0 +1,91 @@
...
+        """Parse a string into a GitUrl."""
+        return cls.from_parsed(urlparse(url.strip()))
+
+    @classmethod
+    def from_parsed(cls, url: ParseResult) -> GitUrlError | GitUrl:
+        """Tests a parsed url if it is a valid GitUrl."""
+
+        if url.scheme == "":
+            return GitUrlError.no_url_scheme
+        if url.scheme not in ["http", "https"]:
```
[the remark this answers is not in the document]
leafty: OK, maybe we can have an issue to change it later.

components/renku_data_services/repositories/git_url.py
```diff
@@ -0,0 +1,91 @@
...
+            return GitUrlError.invalid_url_scheme
+        if url.netloc == "":
+            return GitUrlError.no_url_host
+        # Don't allow localhost
+        if url.netloc.startswith("localhost:") or url.netloc == "localhost":
+            return GitUrlError.invalid_git_url
+        if url.netloc.startswith("127.0.0.1:") or url.netloc == "127.0.0.1":
+            return GitUrlError.invalid_git_url
```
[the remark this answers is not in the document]
eikek: Yes, thanks. Always forget that we have millions of loop-back addresses…

test/bases/renku_data_services/data_api/test_repositories.py
eikek: Ok, thanks. I thought it doesn't matter what exactly fails in the tests. But I assume we have better error messages then?

components/renku_data_services/connected_services/oauth_http.py
```diff
@@ -0,0 +1,611 @@
...
+                    self._connection_id, current_token, self.leeway
+                )
+                if updated_token is None:
+                    raise errors.InvalidTokenError(
+                        message="The refresh token for the connected service has expired or is invalid.",
+                        detail=f"Please reconnect your integration for the service with ID {str(self._connection_id)} "
+                        "and try again.",
+                    ) from err
+                else:
+                    self.token = updated_token
```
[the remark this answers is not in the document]
olevski: Ok I will submit a new suggestion,

components/renku_data_services/connected_services/oauth_http.py
```diff
@@ -0,0 +1,611 @@
...
```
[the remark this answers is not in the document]
eikek: Thanks!

PR #1098, opened 2025-11-05
fix: handle session secret slots when copying a project

### LINKED ISSUE #1096: Copying a project should copy its secret slots
Copying a project should copy the secret slots found in the project. Also, if secrets are provided for the source project, provide the same secrets in the copied project.

leafty:
Closes #1096.

/deploy renku-ui=main

### DIFF
--- a/.github/workflows/acceptance-tests.yml
+++ b/.github/workflows/acceptance-tests.yml
@@ -73,7 +73,7 @@ jobs:
 
       - name: Create comment deployment url
         if: steps.deploymentUrlMessage.outputs.comment-id == 0
-        uses: peter-evans/create-or-update-comment@v4
+        uses: peter-evans/create-or-update-comment@v5
         with:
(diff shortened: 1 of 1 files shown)


components/renku_data_services/project/core.py
```diff
@@ -95,6 +96,15 @@ async def copy_project(
...
 
+    # NOTE: Copy session secret slots
+    secret_slots = await session_secret_repo.get_all_session_secret_slots_from_project(
+        user=user, project_id=source_project_id
+    )
+    for secret_slot in secret_slots:
+        await session_secret_repo.copy_session_secret_slot(
+            user=user, project_id=project.id, session_secret_slot=secret_slot
+        )
+
```
eikek: Could this be done in the db entirely maybe? with `insert … select …` or something? not blocking, just wondering. could be quicker and perhaps easier to write as well.

eikek (approved):

PR #1100, opened 2025-11-05
refactor: improve typing of RESET

leafty:
Before this change: There is no guard against having multiple instances of the `ResetType` and the Python type-checkers know this. Developers have to write type guards where this should not be necessary when handling values which can be `RESET`.

After this change: There is only on single instance of the `ResetType` and the Python type-checkers know this. Code can be written using `is RESET` or `is not RESET` and no additional type guards are needed.

Example of unnecessary type guard flagged after this change:
Screenshot 2025-11-05 at 13 16 47

As another example, consider:
```python
from renku_data_services.base_models import RESET, ResetType

def my_upper_function(value: str | ResetType) -> str:
    if value is RESET:
        return ""
    else:  # mypy knows that value can only be of type 'str' here
        return f"value: {value.upper()}"
```


eikek (approved):
Thank you!

PR #1101, opened 2025-11-05
feat: indicate supported platforms when checking a session image

leafty:
Add a new field `platforms` returned when checking a session image with `GET /api/data/sessions/images`. The field may be returned when the image is accessible and will contain the list of supported platforms. If gathering this information fails, the field will simply not be present in the response.

Example:
```bash
curl 'https://<hostname>/api/data/sessions/images?image_url=node:24' | jq
```

```json
{
  "accessible": true,
  "platforms": [
    {
      "architecture": "amd64",
      "os": "linux"
    },
    {
      "architecture": "arm64",
      "os": "linux",
      "variant": "v8"
    },
    {
      "architecture": "ppc64le",
      "os": "linux"
    },
    {
      "architecture": "s390x",
      "os": "linux"
    }
  ]
}
```

Implementation notes:
* The current code related to handling of OCI images is split between `renku_data_services.notebooks.api.classes.image` and `renku_data_services.notebooks.image_check`. This will be consolidated in the future inside a single Python module.
* The new `renku_data_services.notebooks.oci` may seem redundant at present, but the existing image handling code will be refactored to be contained only in the `oci` module.

/deploy

### DIFF
--- a/components/renku_data_services/notebooks/core_sessions.py
+++ b/components/renku_data_services/notebooks/core_sessions.py
@@ -5,7 +5,7 @@ import json
 import os
 import random
 import string
-from collections.abc import AsyncIterator, Sequence
+from collections.abc import AsyncIterator, Mapping, Sequence
 from datetime import timedelta
 from pathlib import PurePosixPath
 from typing import Protocol, TypeVar, cast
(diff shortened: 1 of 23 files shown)


components/renku_data_services/notebooks/core_sessions.py
```diff
@@ -983,7 +984,8 @@ async def patch_session(
     project_session_secret_repo: ProjectSessionSecretRepository,
     rp_repo: ResourcePoolRepository,
     session_repo: SessionRepository,
-    connected_svcs_repo: ConnectedServicesRepository,
+    # connected_svcs_repo: ConnectedServicesRepository,
```
leafty:
```suggestion
```

components/renku_data_services/notebooks/core_sessions.py
```diff
@@ -1119,7 +1121,8 @@ async def patch_session(
         image=image,
         server_name=server_name,
         nb_config=nb_config,
-        connected_svcs_repo=connected_svcs_repo,
+        # connected_svcs_repo=connected_svcs_repo,
```
leafty:
```suggestion
```

PR #1108, opened 2025-11-12
fix: use notebooks config fallback for culling

olevski:
/deploy

### DIFF
--- a/components/renku_data_services/notebooks/core_sessions.py
+++ b/components/renku_data_services/notebooks/core_sessions.py
@@ -527,24 +527,36 @@ def get_culling(
     user: AuthenticatedAPIUser | AnonymousAPIUser, resource_pool: ResourcePool, nb_config: NotebooksConfig
 ) -> Culling:
     """Create the culling specification for an AmaltheaSession."""
-    if user.is_anonymous:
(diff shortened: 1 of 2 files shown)


components/renku_data_services/notebooks/core_sessions.py
```diff
@@ -527,24 +527,36 @@ def get_culling(
...
+    match (user.is_anonymous, resource_pool.hibernation_threshold):
+        case True, _:
+            # NOTE: Anonymous sessions should not be hibernated at all, but there is no such option in Amalthea
+            # So in this case we set a very low hibernation threshold so the session is deleted quickly after
+            # it is hibernated.
+            hibernation_threshold = timedelta(seconds=1)
+        case False, int():
+            hibernation_threshold = timedelta(seconds=resource_pool.hibernation_threshold)
+        case False, None:
+            hibernation_threshold = timedelta(seconds=nb_config.sessions.culling.registered.hibernated_seconds)
```
olevski: This was removed in a previous change I made. And we actually fallback to this often.

olevski: The similar nb_config fallback was also removed for idling.

PR #1109, opened 2025-11-13
feat: add alertmanager webhook compatibility

wesjdj:
/deploy #notest

### DIFF
--- a/components/renku_data_services/notifications/core.py
+++ b/components/renku_data_services/notifications/core.py
@@ -37,7 +37,8 @@ def transform_alert_to_unsaved_alert(alert: apispec.AlertmanagerAlert) -> models
 
     session_name = labels.get("statefulset")
     if not session_name:
-        session_name = None
+        # We discard alerts without a session for now
+        return None
 
     title = annotations.get("title")
     if not title:


components/renku_data_services/notifications/api.spec.yaml
```diff
@@ -82,8 +82,140 @@ paths:
           $ref: "#/components/responses/Error"
       tags:
         - alerts
+  /alerts/webhook/alertmanager:
```
leafty: I think I would prefer that the path is not under `/alerts`: we can add support for any types of webhooks, what they trigger will depend on the internal logic.


```suggestion
  /webhooks/alertmanager:
```

leafty (commented):
Looks good 👍

components/renku_data_services/notifications/core.py
```diff
@@ -19,3 +23,63 @@ def validate_alert_patch(patch: apispec.AlertPatch) -> models.AlertPatch:
...
+    return models.UnsavedAlert(
+        title=title,
+        message=message,
+        event_type=event_type,
+        user_id=user_id,
+        session_name=session_name,
+    )
+
+
+def transform_alertmanager_webhook(
```
sgaist: The function name is not really intuitive to grasp what it is going to do.

maybe `webhook_to_alerts` or `transform_webhook_to_alerts`

components/renku_data_services/notifications/core.py
```diff
@@ -19,3 +23,63 @@ def validate_alert_patch(patch: apispec.AlertPatch) -> models.AlertPatch:
...
+    annotations = alert.annotations
+
+    user_id = labels.get("safe_username")
+    if not user_id:
+        logger.warning("Alert is missing 'safe_username' label, skipping: %s", alert)
+        return None
+
+    session_name = labels.get("statefulset")
+    if not session_name:
+        session_name = None
```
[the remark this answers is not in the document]
leafty: No, but we do not have a use-case for capturing non-session alerts at the moment, so the processing should stop and `return None`.

leafty: @wesjdj 

```suggestion
        # We discard alerts without a session for now
        return None
```

PR #1119, opened 2025-11-19
feat: add envidat data connectors support

olevski:
/deploy renku-ui=olevski/feat-modify-global-dc-display

Initial support for envidat datasets.

The following changes are made:
- adds the ability to mount a data connector from http://envidat.ch
- adds the data connector description, publisher name, doi and keywords to the search
- makes the doi, publisher name and keywords case insesitive when searching
- makes project keywords case insensitive when searching
- adds doi and publisher_name as parameters for search query parser

### DIFF
--- a/components/renku_data_services/search/query_manual.md
+++ b/components/renku_data_services/search/query_manual.md
@@ -217,6 +217,38 @@ silently ignored.
 This content will be replaced by the output of the code block above.
 <!-- OUTPUT:END -->
 
+## DOIs
+
+These are applicable only for data connectors created from DOIs.
+The DOI value provided is not changed at all by the query parser.
+Note that we store all DOI values not as full URLs
+(i.e. `https://www.doi.org/10.16904/envidat.714`) but
+only as the value (i.e. `10.16904/envidat.714`). Future improvements
+will probably remove these limitations and handle full URLs transparently.
+The search will look up this term in a case-insensitive manner.
+
+<!-- CODE:START -->
+<!-- print(f"""``` -->
+<!-- {Segments.doi_is("10.16904/envidat.714").render()} -->
+<!-- ```""") -->
+<!-- CODE:END -->
+<!-- OUTPUT:START -->
+This content will be replaced by the output of the code block above.
+<!-- OUTPUT:END -->
+
+## Publisher names
+
+These are applicable only for data connectors created from DOIs.
+The search will look up this term in a case-insensitive manner.
+
+<!-- CODE:START -->
+<!-- print(f"""``` -->
+<!-- {Segments.publisher_name_is("EnviDat").render()} -->
+<!-- ```""") -->
+<!-- CODE:END -->
+<!-- OUTPUT:START -->
+This content will be replaced by the output of the code block above.
+<!-- OUTPUT:END -->
 
 ## Dates
 
--- a/test/components/renku_data_services/search/test_user_query_parser.py
+++ b/test/components/renku_data_services/search/test_user_query_parser.py
@@ -15,6 +15,7 @@ from renku_data_services.search.user_query import (
     CreatedByIs,
     DateTimeCalc,
     DirectMemberIs,
+    DoiIs,
     IdIs,
     InheritedMemberIs,
     KeywordIs,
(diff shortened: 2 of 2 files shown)


components/renku_data_services/data_connectors/doi/schema_org.py
```diff
@@ -0,0 +1,77 @@
...
+        # NOTE: PurePosixPath("/test") / "/subfolder" == /subfolder, so we still have to strip /
+        return (PurePosixPath("/") / PurePosixPath(self.bucket) / self.prefix.lstrip("/")).as_posix()
+
+
+def get_rclone_config(dataset: SchemaOrgDataset, provider: DatasetProvider) -> S3Config:
+    """Parse the dataset into an rclone configuration."""
+    match provider:
+        case DatasetProvider.envidat:
+            return __get_rclone_s3_config_envidat(dataset)
+        # TODO: Add scicat here
```
eikek: is this left on purpose?

components/renku_data_services/search/user_query.py
```diff
@@ -812,6 +864,16 @@ def role_is(cls, role: Role, *args: Role) -> Segment:
         """Return role-is query segment."""
         return RoleIs(Nel(role, list(args)))
 
+    @classmethod
+    def doi_is(cls, doi: str, *args: str) -> Segment:
+        """Return slug-is query segment."""
```
eikek: The comment is revealing the copy source :-) (below the same)

components/renku_data_services/solr/entity_documents.py
```diff
@@ -287,7 +289,7 @@ def _add_tzinfo(cls, v: datetime) -> datetime:
 
     @classmethod
     def from_dict(cls, d: dict[str, Any]) -> DataConnector:
-        """Create a Project from a dictionary."""
+        """Create a data connector from a dictionary."""
```
eikek: Thanks 🙏🏼

components/renku_data_services/data_connectors/doi/schema_org.py
olevski: Yes we will add another provider Scicat from PSI soon.

components/renku_data_services/data_connectors/core.py
```diff
@@ -182,15 +267,18 @@ async def validate_unsaved_global_data_connector(
 
     # Assign user-friendly target_path if possible
     target_path = data_connector.slug
-    with contextlib.suppress(errors.ValidationError):
-        name_slug = base_models.Slug.from_name(name).value
-        target_path = base_models.Slug.from_name(f"{name_slug[:30]}-{target_path}").value
+    if metadata:
```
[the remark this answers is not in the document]
olevski: Yes it is confusing. Basically if there is no metadata the slug is the same as the name. So when we build the target path with the slug and the name it looks ugly because it is the same thing repeated twice.

I changed the logic and if statement. I think now it makes more sense.

components/renku_data_services/data_connectors/core.py
```diff
@@ -378,3 +469,52 @@ def transform_secrets_for_front_end(
...
+    if len(doi) == 0:
+        raise errors.ValidationError(
+            message="Cannot get configuration for Envidat data connector because the doi is a string with zero length."
+        )
+    doi = DOI(doi)
+
+    new_config = payload.model_copy(deep=True)
+    new_config.configuration = {}
+
+    envidat_url = "https://envidat.ch/converters-api/internal-dataset/convert/jsonld"
```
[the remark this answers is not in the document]
olevski: It was only needed in one place. I removed one instance.

components/renku_data_services/search/solr_user_query.py
```diff
@@ -344,6 +346,14 @@ async def visit_slug_is(self, ft: SlugIs) -> None:
         """Process slug-is segment."""
         self.__append(st.field_is_any(Fields.slug, ft.values.map(st.from_str)))
 
+    async def visit_doi_is(self, ft: DoiIs) -> None:
```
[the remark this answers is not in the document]
olevski: done in https://github.com/SwissDataScienceCenter/renku-data-services/pull/1119/commits/7c7211842bf677a0c101fe60a6458c4760c6977b

Issue #1129, opened 2025-12-10
Renaming a group does not update the search


olevski:
To reproduce:
- create a group
- create a project in the group
- go to the groups search page - you should see the project
- rename the group
- go to the group search page, the project will not be there

The problem is that the change of the group slug/name does not update the search.

PR #1151, opened 2025-12-19
fix: updating group slug did not update search

mohammad-alisafaee:
Fixes https://github.com/SwissDataScienceCenter/renku-data-services/issues/1129

/deploy extra-values=enableInternalGitlab=false

### DIFF
--- a/components/renku_data_services/connected_services/oauth_http.py
+++ b/components/renku_data_services/connected_services/oauth_http.py
@@ -14,6 +14,7 @@ import sqlalchemy as sa
 import sqlalchemy.orm as sao
 from authlib.integrations.base_client import InvalidTokenError, OAuthError
 from authlib.integrations.httpx_client.oauth2_client import AsyncOAuth2Client
+from cryptography.fernet import InvalidToken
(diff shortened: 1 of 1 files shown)


components/renku_data_services/namespace/db.py
```diff
@@ -262,12 +265,17 @@ async def update_group(
...
         if patch.name is not None:
             group.name = patch.name
         if patch.description is not None:
             group.description = patch.description if patch.description else None
 
-        return group.dump()
+        new_group = group.dump()
+
+        if slug_changed:
+            return models.GroupUpdate(old=old_group, new=new_group)
```
eikek: I think I would then always return `GroupUpdate`. It is a bit confusing that `update_group(…)` returns only an `GroupUpdate` when the slug changed, otherwise only the new group. Since we have the data now anyways, I think it is more consistent in general to always return the `GroupUpdate` no matter what changed. Wdyt?

eikek (approved):

eikek (approved):

PR #1161, opened 2026-01-07
feat: use current global images in new deployments

leafty:
Closes #1161.

* Removes the old images from database migrations
* Add the new global session environments with a task in `data-tasks`

/deploy extra-values=enableInternalGitlab=false

### DIFF
--- a/components/renku_data_services/namespace/blueprints.py
+++ b/components/renku_data_services/namespace/blueprints.py
@@ -97,7 +97,7 @@ class GroupsBP(CustomBlueprint):
         ) -> JSONResponse:
             group_patch = validate_group_patch(body)
             res = await self.group_repo.update_group(user=user, slug=slug, patch=group_patch)
-            return validated_json(apispec.GroupResponse, res)
+            return validated_json(apispec.GroupResponse, data=res.new)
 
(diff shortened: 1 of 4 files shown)


bases/renku_data_services/data_tasks/dependencies.py
```diff
@@ -56,13 +59,21 @@ def from_env(cls, cfg: Config | None = None) -> "DependencyManager":
...
+            builds_config=None,  # type: ignore
+        )
         syncer = UsersSync(
             cfg.db.async_session_maker,
             group_repo=group_repo,
             user_repo=user_repo,
             metrics=metrics,
             authz=authz,
         )
+        session_tasks = SessionTasks(session_repo=session_repo)
```
eikek: As a "mid-term" thing, we could introduce a `protocol` that exposes only the two methods required by `SessionTasks`. It wouldn't remove the `# type: ignore` statements above, but makes it harder to accidentally use unsupported methods in later changes to `SessionTasks`. Some special constructor/method on `SessionRepository` could return that `protocol`, for example, so it is type-safe after the point of creation.

eikek (approved):

PR #1169, opened 2026-01-14
fix: use ingress class name for local cluster

olevski:
/deploy

This fixes a problem that @aledegano is seeing when trying to deploy Renku.


olevski:
This is not a problem in production because if we do not see the ingress class name in its dedicated field then we look for it in the ingress annotations. And in production we specify this annotation `kubernetes.io/ingress.class: webapprouting.kubernetes.azure.com`

PR #1185, opened 2026-01-22
refactor: remove v1 legacy code

olevski:
/deploy

### DIFF
--- a/bases/renku_data_services/data_api/dependencies.py
+++ b/bases/renku_data_services/data_api/dependencies.py
@@ -86,6 +86,14 @@ default_resource_pool = crc_models.UnsavedResourcePool(
             default=True,
         ),
         crc_models.UnsavedResourceClass(
+            name="medium",
+            cpu=0.75,
+            memory=1,
+            max_storage=20,
+            gpu=0,
+            default=False,
+        ),
+        crc_models.UnsavedResourceClass(
             name="large",
             cpu=1.0,
             memory=2,


components/renku_data_services/notebooks/api/amalthea_patches/git_proxy.py
```diff
@@ -15,7 +15,7 @@
 
 if TYPE_CHECKING:
     # NOTE: If these are directly imported then you get circular imports.
-    from renku_data_services.notebooks.api.classes.server import UserServer
+    pass
```
leafty: Just remove the `if TYPE_CHECKING:` block.

olevski: This is some weird part of my autoformatting doing this. I will remove it.

components/renku_data_services/notebooks/api/classes/data_service.py
```diff
@@ -50,84 +49,32 @@ async def validate_class_storage(
...
 
 
 @dataclass
 class DummyCRCValidator:
     """Dummy validator for resource pools and classes."""
 
-    options: ServerOptions = field(default_factory=lambda: ServerOptions(0.5, 1, 0, 1, "/lab", False, True))
-
-    async def validate_class_storage(self, user: APIUser, class_id: int, storage: int | None = None) -> ServerOptions:
+    async def validate_class_storage(self, user: APIUser, class_id: int, storage: int | None = None) -> None:
```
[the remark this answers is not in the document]
olevski: It used to also return this `ServerOptions` object that was only needed in V1. But it is still being called from the session starting code for v2 and the output in the v2 code is not used at all.

PR #1187, opened 2026-01-22
feat: consistent logging format and json support

mohammad-alisafaee:
* Replace usage of Python's logging module with our own for consistency
* Include `trace_id` in errors to be displayed to users in UI

/deploy #notest renku-gateway=build/consistent-logging-format extra-values=enableInternalGitlab=false,dataService.sentry.enabled=true,dataService.sentry.environment=renku-ci-ds-1187,dataService.sentry.dsn=https://5ec3c75875797b564a9f4654d5fc46d5@o4509039310995456.ingest.de.sentry.io/4509044015169616,dataService.sentry.sampleRate=0.0001,gateway.sentry.enabled=true,gateway.sentry.environment=renku-ci-ds-1187,gateway.sentry.dsn=https://b4747210a4b6a37a4ccc33beee2dbea7@o4509039310995456.ingest.de.sentry.io/4509513007693904,gateway.sentry.sampleRate=0.0001

### DIFF
--- a/bases/renku_data_services/data_api/dependencies.py
+++ b/bases/renku_data_services/data_api/dependencies.py
@@ -28,7 +28,7 @@ from renku_data_services.connected_services.db import ConnectedServicesRepositor
 from renku_data_services.connected_services.oauth_http import DefaultOAuthHttpClientFactory, OAuthHttpClientFactory
 from renku_data_services.crc import models as crc_models
 from renku_data_services.crc.constants import DEFAULT_RUNTIME_PLATFORM
-from renku_data_services.crc.db import ClusterRepository, ResourcePoolRepository, UserRepository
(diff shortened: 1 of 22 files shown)


pyproject.toml
```diff
@@ -275,6 +275,10 @@ module = [
 ]
 ignore_missing_imports = true
 
+[[tool.mypy.overrides]]
+module = ["test.*"]
+disable_error_code = ["no-untyped-def", "var-annotated", "import-untyped"]
+
```
mohammad-alisafaee: This is for PyCharm. It passes files individually to `mypy` that ignores `mypy` settings in general but only considers these overrides. Having `mypy` checks in tests is useful sometimes but it can be noisy if we don't disabled these checks.

PR #1195, opened 2026-01-28
feat: add support for mounting data connectors using OAuth 2.0 credentials

leafty:
_Merging into a feature branch_

Details:
* Add DataSourceRepository class to handle OAuth 2.0 credentials when mounting data connectors of type `drive` and `dropbox`.
*  When a session start is requested, insert `token` and `token_url` into the `rclone` configuration.
* When a session resume is requested, read the `rclone` configuration from k8s secrets and update the `token` and `token_url` fields.
* Add methods to read k8s secrets (create, patch, delete already exist).
* Add a `POST /oauth2/connections/<connection_id:ulid>/token_endpoint` endpoint: it conforms to [RFC 6749 - Section 6: Refreshing an Access Token](https://datatracker.ietf.org/doc/html/rfc6749#section-6). This is what is used as the `token_url` in the `rclone` configurations. This means that data services handles OAuth 2.0 token refreshes which is critical for platforms which issue single-use refresh tokens.

PR stack:
* (base) #1201
  * (this) #1195
    * (tip) #1200


components/renku_data_services/notebooks/core_sessions.py
```diff
@@ -348,6 +356,83 @@ async def get_data_sources(
...
+                ulid = name[len(secret_prefix) :]
+                try:
+                    mounted_dcs.append((ULID.from_str(ulid.upper()), name))
+                except ValueError:
+                    logger.warning(f"Could not parse {ulid.upper()} as a ULID.")
+    async for dc in data_connectors_stream:
+        if not data_source_repo.is_patching_enabled(dc.data_connector):
+            continue
+        dc_id = dc.data_connector.id
+        mounted_dc = next(filter(lambda tup: tup[0] == dc_id, mounted_dcs), None)
```
olevski: suggestion(non-blocking): if mounted_dcs is a `dict[ULID, str]` then you can find a dc that is mounted a lot easily without having to filter.

olevski (approved):

PR #1196, opened 2026-01-29
feat: endpoints for data deposits

olevski:
Note: the target branch is a feature branch.


components/renku_data_services/connected_services/oauth_http.py
```diff
@@ -106,6 +105,12 @@ async def for_user_connection_raise(self, user: APIUser, connection_id: ULID) ->
         """Same as `for_user_connection` but throws on error."""
         ...
 
+    async def for_user_provider_access_token(
```
leafty: I'm not sure why we would need to add this method when it is possible to use `for_user_connection`. That method returns a configured OAuth 2.0 client, which will also handle token refresh if needed. By returning just a token, you do not make sure that the token is not expired and that it will be refreshed and saved back in the database.

PR #1205, opened 2026-02-10
feat: add capacity reservation

wesjdj:
/deploy renku=add-capacity-reservation


components/renku_data_services/crc/db.py
```diff
@@ -1099,6 +1099,31 @@ async def update(self, api_user: base_models.APIUser, cluster: ClusterPatch, clu
...
+            return ClusterId(result)
+
+    async def get_resource_class_by_id(self, class_id: int) -> models.ResourceClass | None:
+        """Return the resource class with the given ID, including quota from the parent resource pool."""
+        async with self.session_maker() as session:
+            stmt = select(schemas.ResourceClassORM).where(schemas.ResourceClassORM.id == class_id)
+            result = await session.scalar(stmt)
+            if result is None:
+                return None
+            return result.dump()
```
olevski: I am a bit worried because this is the only unprotected method in a class where all other similar methods are protected (or return trivial info like ClusterId). We have a constant somwehere in the repo that indentifies an internal service user. You can pass that in if you need to use this method but the call is not triggered by the user.

components/renku_data_services/capacity_reservation/core.py
```diff
@@ -0,0 +1,200 @@
...
+    reservation: models.CapacityReservation,
+    occurrence: models.Occurrence,
+    active_sessions: int,
+    now: datetime,
+) -> int:
+    """Calculate the target number of placeholder replicas for an active occurrence."""
+    provisioning = reservation.provisioning
+
+    if provisioning.scale_down_behavior == models.ScaleDownBehavior.NONE:
+        raise ValueError("calculate_target_replicas called for NONE scale down behavior")
```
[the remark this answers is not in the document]
olevski: question: Do you even need to raise an error here? If the scale down behaviour is none then wouldn't you just return the provisioning.placeholder_count? So even if someone calls this and the scale down is none they will just get the right behaviour.

components/renku_data_services/capacity_reservation/tasks.py
```diff
@@ -0,0 +1,214 @@
...
+        for expired_occurrence in expired_occurrences:
+            logger.info(f"Deactivating occurrence {expired_occurrence.id} as its end time has passed.")
+            if expired_occurrence.deployment_name is None:
+                logger.warning(f"Expired occurrence {expired_occurrence.id} has no deployment name, skipping delete.")
+            else:
+                reservations = await self.capacity_reservation_repo.get_capacity_reservations_by_ids(
+                    [expired_occurrence.reservation_id]
+                )
+                if reservations:
+                    await self.k8s_client.delete_deployment(expired_occurrence.deployment_name, reservations[0])
```
olevski: Same as above. Why not loop over reservations?

components/renku_data_services/capacity_reservation/k8s_client.py
```diff
@@ -0,0 +1,258 @@
...
+    if tolerations:
+        pod_spec["tolerations"] = tolerations
+
+    affinity_spec: dict[str, Any] = {}
+    if default_affinity:
+        affinity_spec = default_affinity.copy()
+
+    if resource_class.node_affinities or default_affinity.get("nodeAffinity"):
+        required_affinities = [a for a in resource_class.node_affinities if a.required_during_scheduling]
+        preferred_affinities = [a for a in resource_class.node_affinities if not a.required_during_scheduling]
```
olevski: issue: I think the lower one is supposed to be preferred during scheduling. Now they are both the same.

olevski (approved):

PR #1207, opened 2026-02-10
feat: Enable resource requests tracking

eikek:
- adds ability to track resource requests in periodic intervals
- runs as a loop in data-tasks

/deploy renku=eikek/env-vars-data-tasks amalthea-sessions=main

### DIFF
--- a/components/renku_data_services/resource_usage/api.spec.yaml
+++ b/components/renku_data_services/resource_usage/api.spec.yaml
@@ -183,7 +183,7 @@ paths:
           content:
             "application/json":
               schema:
-              $ref: "#/components/schemas/ResourceClassCostPut"
+                $ref: "#/components/schemas/ResourceClassCostPut"
         "404":
           description: "The class does not exist."
           content:


components/renku_data_services/resource_usage/api.spec.yaml
```diff
@@ -0,0 +1,320 @@
...
+components:
+  schemas:
+    ResourcePoolUsage:
+      type: object
+      additionalProperties: false
+      properties:
+        total_usage:
+          $ref: "#/components/schemas/ResourceUsageSummary"
+        user_usage:
+          $ref: "#/components/schemas/ResourceUsageSummary"
```
olevski: question: What is the difference between `total_usage` and `user_usage`?

Does user usage include only pods that have a specific user_id and therefore we can allocate them to that user?

components/renku_data_services/resource_usage/blueprints.py
```diff
@@ -0,0 +1,186 @@
...
+            resource_pool_id: int,
+            class_id: int,
+            body: apispec.ResourceClassCostPut,
+        ) -> HTTPResponse:
+            costs = validate_resource_class_costs_put(class_id, body=body)
+            result = await self.rr_repo.set_resource_class_costs(costs)
+            if not result:
+                raise errors.MissingResourceError()
+            else:
+                return empty()
```
olevski: issue: for all PUT endpoints we return the new and updated costs. Which is actually the same payload we were sent in the first place... But we already do that in all other places so lets stick to the convention.

components/renku_data_services/resource_usage/orm.py
```diff
@@ -0,0 +1,278 @@
...
+    """The user id associated to the request data."""
+
+    project_id: Mapped[ULID | None] = mapped_column("project_id", ULIDType(), nullable=True)
+    """A project id associated to the requests data."""
+
+    launcher_id: Mapped[ULID | None] = mapped_column("launcher_id", ULIDType(), nullable=True)
+    """The launcher id used to start the session."""
+
+    resource_class_id: Mapped[int | None] = mapped_column("resource_class_id", Integer(), nullable=True)
+    """The resource class id used to start the session."""
```
olevski: suggestion(non-blocking): Maybe it is already worth indexing on this as well? And maybe also the user? But we can come back to this some other time. Feel free to ignore for now.

components/renku_data_services/resource_usage/blueprints.py
```diff
@@ -0,0 +1,186 @@
...
+
+    def delete_pool_limits(self) -> BlueprintFactoryResponse:
+        """Delete defined pool limits."""
+
+        @authenticate(self.authenticator)
+        @only_admins
+        @validate_db_ids
+        async def _delete(_: Request, user: base_models.APIUser, resource_pool_id: int) -> HTTPResponse:
+            await self.rr_repo.delete_resource_pool_limits(resource_pool_id)
+            return empty()
```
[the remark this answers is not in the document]
eikek: I think, the `empty()` has the default status set to 204. It is essentially an alias for `HTTPResponse(status=204)`.

components/renku_data_services/resource_usage/blueprints.py
```diff
@@ -0,0 +1,186 @@
...
+
+    def delete_class_cost(self) -> BlueprintFactoryResponse:
+        """Delete resource class costs limits."""
+
+        @authenticate(self.authenticator)
+        @only_admins
+        @validate_db_ids
+        async def _delete(_: Request, user: base_models.APIUser, resource_pool_id: int, class_id: int) -> HTTPResponse:
+            await self.rr_repo.delete_resource_class_costs(class_id)
+            return empty()
```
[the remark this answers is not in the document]
olevski: Sorry Eike I did not know that. Somehow I thought empty is just an empty 200 response.

olevski (dismissed):
Thanks @eikek. This is good to go. Wes merged his stuff sooner so you will have to resolve some conflicts. Let me know if you need help there because I reviewed his PR.

PR #1238, opened 2026-03-11
feat: Add resource usage endpoint to get usage for a given user

eikek:
This only adds another endpoint to get resource usage of a pool for admins, that allows to specify a `user_id`.

/deploy


sgaist (changes requested):
One last question otherwise looks good

sgaist (dismissed):

leafty (changes requested):

leafty (dismissed):
LGTM

Looks very clean now 👏

components/renku_data_services/resource_usage/api.spec.yaml
leafty, line 1: [non-blocking] Why is this in a separate API spec file? Since this is about resource pools, would it not make sense to have the spec and code part of the existing resource pools?

eikek: On start, I thought to have it as a new module (not in this PR) amending the things in the crc module, like an additional feature not affecting the core features. The crc module contains already a lot of code, I thought a separate one is maybe nice. There were also separate api paths. Then later I got requests to change the paths.

leafty (approved):

leafty (approved):

PR #1244, opened 2026-03-24
feat: return used credits and remaining resource hours in pools

mohammad-alisafaee:
Modifies `/resource_pools` endpoint to return `resource_usage` for each resource pool and number of hours available (`resource_available` and `resource_available_percentage`) for each resource class in a pool.

/deploy renku=release-2.17.0


leafty (dismissed):
Not sure why, but the service is broken as is. The `/api/data/user` endpoint is not working. This may be due to mis-configured dependencies and/or issues with async code:

```
2026-05-04T07:38:42.990337 [ERROR] 166/MainThread sanic.error (error.py:201) [f7cbf263-3a85-4f20-79fe-b2f7d9f26f01][-] - Exception occurred while handling uri: 'https://renku-ci-ds-1244.dev.renku.ch/api/data/user'
  File "/app/env/lib/python3.13/site-packages/renku_data_services/authz/authz.py", line 801, in decorated_function
    await db_repo.authz.client.WriteRelationships(authz_change.apply)
  File "/app/env/lib/python3.13/site-packages/grpc/aio/_call.py", line 308, in __await__
    response = yield from self._call_response
RuntimeError: Task <Task pending name='Task-1784' coro=<connection_task() running at connection_task:15>> got Future <Task pending name='Task-1859' coro=<UnaryUnaryCall._invoke() running at /app/env/lib/python3.13/site-packages/grpc/aio/_call.py:577>> attached to a different loop

During handling of the above exception, another exception occurred:

  File "handle_request", line 99, in handle_request
    Extend = TypeVar("Extend", type)  # type: ignore
                       ^^^^^^^^^^^^^^
  File "/app/env/lib/python3.13/site-packages/renku_data_services/base_api/auth.py", line 36, in decorated_function
    response = await f(request, user, *args, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/env/lib/python3.13/site-packages/renku_data_services/base_api/auth.py", line 155, in decorated_function
    response = await f(*args, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/env/lib/python3.13/site-packages/renku_data_services/users/blueprints.py", line 62, in _get_self
    user_info = await self.repo.get_or_create_user(requested_by=user, id=user.id)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/env/lib/python3.13/site-packages/renku_data_services/users/db.py", line 144, in get_or_create_user
    return await self._add_api_user(requested_by)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/env/lib/python3.13/site-packages/renku_data_services/users/db.py", line 113, in _add_api_user
    result = await self._users_sync.update_or_insert_user(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<6 lines>...
    )
    ^
  File "/app/env/lib/python3.13/site-packages/renku_data_services/utils/core.py", line 83, in transaction_wrapper
    return await f(self, *args, **kwargs)
  File "/app/env/lib/python3.13/site-packages/renku_data_services/authz/authz.py", line 811, in decorated_function
    await asyncio.shield(db_repo.authz.client.WriteRelationships(authz_change.undo))
  File "/usr/local/lib/python3.13/asyncio/tasks.py", line 737, in _wrap_awaitable
    return await awaitable
           ^^^^^^^^^^^^^^^
  File "/app/env/lib/python3.13/site-packages/grpc/aio/_call.py", line 308, in __await__
    response = yield from self._call_response
RuntimeError: Task <Task pending name='Task-1862' coro=<ensure_future.<locals>._wrap_awaitable() running at /usr/local/lib/python3.13/asyncio/tasks.py:737> cb=[shield.<locals>._inner_done_callback() at /usr/local/lib/python3.13/asyncio/tasks.py:958]> got Future <Task pending name='Task-1861' coro=<UnaryUnaryCall._invoke() running at /app/env/lib/python3.13/site-packages/grpc/aio/_call.py:577>> attached to a different loop
```

components/renku_data_services/crc/api.spec.yaml
```diff
@@ -1101,144 +1107,93 @@ components:
...
         name: "resource class"
+        default: true
         cpu: 1.5
         memory: 2
         gpu: 0
         max_storage: 100
         default_storage: 10
         id: 1
-        default: true
         matching: true
```
olevski, line 1197: question: How come ResourceClassWithId does not have usage but ResourceClassWithIdFiltered does?

components/renku_data_services/crc/core.py
```diff
@@ -635,3 +629,41 @@ def __validate_runtime_platform(body: apispec.RuntimePlatform | None) -> models.
...
+    Return available hours and its percentage based on remaining quota, or (None, None) if no limits are set
+    """
+    limits = await resource_requests_repo.find_resource_pool_limits(resource_pool_id)
+    if not limits:
+        return None, None
+
+    # NOTE: Calculate quota based on user limit if it exists, otherwise use total limit
+    # NOTE: A value of 0 for the limit means that there is no limit, so we return None (instead of 0) to inform the
+    # caller that there is no limit.
+    # TODO: Are we sure that total limit is GE user limit?
```
olevski: No you cannot be sure that the total limit is >= the user limit. The admin is free to set these independently. So you want to find the min of the two that is not null or not zero. Because a pool with unlimimted resources has a limit = 0.

components/renku_data_services/crc/models.py
```diff
@@ -111,6 +111,8 @@ class ResourceClass(ResourcesCompareMixin):
     node_affinities: list[NodeAffinity] = field(default_factory=list)
     tolerations: list[str] = field(default_factory=list)
     quota: str | None = None
+    usage_available: float | None = None
+    usage_limit_total: float | None = None
```
olevski: issue: The units on this are very unclear. I see that Eike introduced `Credit` we can use that if these are credits. If they are hours then we can simply use a timedelta.

components/renku_data_services/crc/models.py
```diff
@@ -344,6 +346,7 @@ class ResourcePool:
     remote: RemoteConfigurationFirecrest | RemoteConfigurationRunai | None = None
     cluster: SavedClusterSettings | None = None
     platform: RuntimePlatform
+    resource_usage: float | None = None
```
olevski: Make units part of the name. Or if it is hours make it a duration.

components/renku_data_services/crc/core.py
```diff
@@ -635,3 +629,41 @@ def __validate_runtime_platform(body: apispec.RuntimePlatform | None) -> models.
...
+async def calculate_available_usage(
+    resource_requests_repo: ResourceRequestsRepo,
+    resource_pool_id: int,
+    resource_class_id: int,
+    resource_usage: float | None,
+) -> tuple[float | None, float | None]:
+    """Calculate available hours and percentage for a resource class in a pool.
+
+    Return available hours and its percentage based on remaining quota, or (None, None) if no limits are set
+    """
```
olevski: This function says it returns available hours and its percentage based on remaining quota. But from the variable names and code it seems it returns hours for both members of the tuple. Please update the docstring accordingly.

components/renku_data_services/crc/core.py
```diff
@@ -635,3 +629,41 @@ def __validate_runtime_platform(body: apispec.RuntimePlatform | None) -> models.
...
+        return None, None
+
+    resource_class_cost = await resource_requests_repo.find_resource_class_costs(resource_pool_id, resource_class_id)
+    if not resource_class_cost or resource_class_cost.cost.value == 0:
+        # NOTE: No cost is defined, so we cannot calculate remaining hours
+        return None, None
+
+    total_hours = available_quota / resource_class_cost.cost.value
+    resource_usage = resource_usage or 0
+    remaining_quota = available_quota - resource_usage
```
olevski: So `available_quota` and `remaining_quota` mean the same thing to me. Just based on the similar meaning of the words available and remaining.
But in the code `available_quota` means the total credits assigned to a pool and `remaining_quota` means how much is left to be consumed.

```suggestion
    quota = limits.user_limit.value or limits.total_limit.value
    if quota <= 0:
        return None, None

    resource_class_cost = await resource_requests_repo.find_resource_class_costs(resource_pool_id, resource_class_id)
    if not resource_class_cost or resource_class_cost.cost.value == 0:
        # NOTE: No cost is defined, so we cannot calculate remaining hours
        return None, None

    total_hours = quota / resource_class_cost.cost.value
    resource_usage = resource_usage or 0
    remaining_quota = quota - resource_usage
```

leafty (dismissed):
LGTM

components/renku_data_services/crc/orm.py
```diff
@@ -143,7 +143,9 @@ def from_unsaved_model(
         )
 
     def dump(
-        self, matching_criteria: models.ResourceClass | models.UnsavedResourceClass | None = None
+        self,
+        matching_criteria: models.ResourceClass | models.UnsavedResourceClass | None = None,
+        usage_available: float | None = None,
```
olevski: Can we add units to `usage_available`? Or if it is credits then we call it `credits_available`. Also does this refer to the total limit/quota or the stuff that is remaining.

components/renku_data_services/crc/orm.py
```diff
@@ -330,6 +333,7 @@ def dump(
         self,
         quota: models.Quota | None,
         class_match_criteria: models.ResourceClass | models.UnsavedResourceClass | None = None,
+        resource_usage: float | None = None,
```
olevski: How is `resource_usage` different from `usage_available`? What are the units - is it credits or hours?

components/renku_data_services/crc/api.spec.yaml
```diff
@@ -1695,6 +1632,25 @@ components:
...
+      example: 10
+      minimum: 0
+      default: null
+    ResourceUsage:
+      type: number
+      format: double
+      description: Amount of a resource used (in credits)
+      example: 3.141
+      minimum: 0
     Storage:
```
olevski: It is really confusing that "available" is used in the `description` of both `UsageAvailable` and `UsageLimitsTotal`. Also what do we return when there is no limit? None or zero? Are my suggestions correct?

```suggestion
    UsageRemaining:
      type: number
      format: double
      description: The runtime (in hours) that remains for a specific resource pool that has a limit. If the pool has no limit, then a value of zero is returned. <is this true?>
      example: 3.141
      minimum: 0
    UsageLimitTotal:
      type: number
      format: double
      description: The usage limit (in hours) for a specific resource pool. A value of zero indicates there is no limit. <is this true?>
      example: 10
      minimum: 0
      default: null
    ResourceUsage:
      type: number
      format: double
      description: The current usage of a resource pool converted to credits based on the runtime and cost.
      example: 3.141
      minimum: 0
    Storage:
```

olevski (changes requested):
The code looks good. My only problem is with naming and documentation of the api and units.

components/renku_data_services/crc/api.spec.yaml
mohammad-alisafaee: There is some inconsistency in the API response models. While endpoints returning resource classes (e.g., `/classes/{class_id}`) use `ResourceClassWithId`, the `/resource_pools` endpoint which was the primary target for this build, returns `ResourceClassWithIdFiltered` (via `ResourcePoolWithIdFiltered`).

I performed some initial refactoring but decided not to expand the scope to avoid the task becoming unmanageable. Consequently, I only added the quota information to `ResourceClassWithIdFiltered` and left the other models untouched. Furthermore, we currently don't have a use case requiring available and total usage data when querying resource classes specifically.

components/renku_data_services/crc/orm.py
mohammad-alisafaee: This one is not used (we call the `dump` function without passing it), so I removed it.

components/renku_data_services/crc/orm.py
mohammad-alisafaee: This was renamed to `credits_used`.

components/renku_data_services/crc/core.py
mohammad-alisafaee: I removed use of `available` and used `remaining` and `total` to make things clear. They were renamed to `usage_hours_remaining` and `usage_hours_total`.

components/renku_data_services/crc/models.py
mohammad-alisafaee: Renamed to `credits_used`.

components/renku_data_services/crc/models.py
mohammad-alisafaee: Renamed to `usage_hours_remaining` and `usage_hours_total`.

components/renku_data_services/crc/api.spec.yaml
olevski: Ok I see.

olevski (approved):

PR #1246, opened 2026-03-27
feat: Add support to start non-interactive sessions

eikek:
Using the new amalthea feature to manage non-interactive sessions, this change passes the flag from the start request.

Currently, this PR introduces this feature by allowing a launcher to configure an image that can run as a job. A launcher is now either a session or a job. Later changes will improve on that and allow for more convenient usage.

----
/deploy amalthea-sessions=main renku=release-2.17.0


components/renku_data_services/notebooks/api/classes/k8s_client.py
```diff
@@ -182,15 +183,22 @@ async def cluster_by_class_id(self, class_id: int | None, api_user: APIUser) ->
...
-    async def list_sessions(self, safe_username: str) -> list[AmaltheaSessionV1Alpha1]:
+    async def list_sessions(
+        self, safe_username: str, session_mode: SessionMode | None
+    ) -> list[AmaltheaSessionV1Alpha1]:
         """Get a list of sessions that belong to a user."""
+        session_filter = {self.__username_label: safe_username}
+        if session_mode == SessionMode.interactive:
+            session_filter.update({"renku.io/session-type": str(SessionMode.interactive)})
+        if session_mode == SessionMode.non_interactive:
+            session_filter.update({"renku.io/session-type": str(SessionMode.non_interactive)})
```
sgaist: Since session_mode has known values and is a StrEnum:
```suggestion
        session_filter = {
            self.__username_label: safe_username,
            "renku.io/session-type": session_mode,
        }
```
I would also make "renku.io/session-type" a constant.

components/renku_data_services/notebooks/core_sessions.py
```diff
@@ -888,14 +930,15 @@ async def start_session(
...
-    ui_path = f"{ingress_config.url_path}/{environment.default_url.lstrip('/')}"
+    ui_path = f"{ingress_config.url_path}/{environment.default_url.lstrip('/')}" if is_interactive else ""
 
     # Annotations
     annotations: dict[str, str] = {
         "renku.io/project_id": str(launcher.project_id),
         "renku.io/launcher_id": str(launcher_id),
         "renku.io/resource_class_id": str(resource_class.id),
         "renku.io/resource_pool_id": str(resource_pool.id),
+        "renku.io/session_type": str(launch_request.session_mode),
```
sgaist: Follow up to my comment above, `renku.io/session_type` is used in multiple places.

I think that turning these annotations into a StrEnum would be a good thing to do.

components/renku_data_services/notebooks/core_sessions.py
```diff
@@ -977,13 +1031,19 @@ async def start_session(
     launcher_env_variables = get_launcher_env_variables(launcher, launch_request)
     env.extend(launcher_env_variables)
 
-    labels = {"renku.io/safe-username": user.id}
+    labels: dict[str, str] = {
+        "renku.io/safe-username": user.id,
+        "renku.io/session-type": str(launch_request.session_mode),
```
sgaist: Does it mean that we have duplication between labels and annotations ?

components/renku_data_services/notebooks/core_sessions.py
```diff
@@ -1035,6 +1097,7 @@ async def start_session(
             template=Template(metadata=AmaltheaMetadata(annotations=annotations, labels=labels)),
         ),
     )
+    logger.info(f"Push Session Culling: {session.spec.culling}")
```
sgaist: What does this message mean ?

components/renku_data_services/notebooks/models.py
```diff
@@ -185,6 +210,39 @@ class SessionDataConnectorOverride(RCloneStorageRequestOverride):
     readonly: bool | None
 
 
+class SessionMode(StrEnum):
```
sgaist: Why not SessionType ?

Amalthea uses SessionType in its specification, that make things less coherent on the `renku-data-service` side.

Also, you don't need `from_amalthea_name` nor `to_amalthea_name`, you can use `Interactive` and `NonInteractive` as the values for this StrEnum. The casing of the values does not need to match the casing of the keys.

components/renku_data_services/notebooks/models.py
```diff
@@ -194,3 +252,9 @@ class SessionLaunchRequest:
     resource_class_id: int | None
     data_connectors_overrides: list[SessionDataConnectorOverride] | None
     env_variable_overrides: list[SessionEnvVar] | None
+    session_mode: SessionMode
+
+    @property
+    def is_non_interactive(self) -> bool:
+        """Whether this is a request to a non-interactive session."""
+        return self.session_mode == SessionMode.non_interactive
```
sgaist: You have that information directly from the enum:

```suggestion
       return self.session_mode.is_non_interactive
```

test/bases/renku_data_services/data_api/test_notebooks.py
```diff
@@ -73,7 +73,11 @@ async def amalthea_session(
     session = AmaltheaSessionV1Alpha1(
         metadata=dict(
             name=server_name,
-            labels={"renku.io/safe-username": regular_user_api_user.id, "renku.io/userId": regular_user_api_user.id},
+            labels={
+                "renku.io/safe-username": regular_user_api_user.id,
+                "renku.io/userId": regular_user_api_user.id,
+                "renku.io/session-type": "interactive",
```
sgaist, line 79: Another reason advocating for using enums: avoiding so many strings repetition for what are constants.

components/renku_data_services/notebooks/cr_amalthea_session.py
```diff
@@ -3823,6 +3823,10 @@ class Spec(BaseCRD):
         ...,
         description="Specification for the main session container that the user will access and use",
     )
+    sessionType: Optional[str] = Field(
```
sgaist: Can't this be described as `Optional[SessionType]` (or `Optional[SessionMode]` with the current implemtation)?

sgaist (changes requested):

components/renku_data_services/notebooks/cr_amalthea_session.py
olevski: Just fyi this is auto-generated code. We dont have freedom to change it much.

sgaist: Yes, that's clear, hence the use of described. I don't know if the code generator can be guided for that.

olevski: I checked and on the amalthea side we have done all the right things. So if the code generation fails to recognize them, then there isn't much we can do.

olevski: Sorry for spamming. The code generation seems to generate enums for all the other cases like this... So we should figure this out.

olevski (commented):
Eike this looks good. If you address Samuel's comments it is good to go. I have nothing to add.

components/renku_data_services/notebooks/api/classes/k8s_client.py
eikek: Thanks, oh yes the `ifs` are not necessary. I'm still unsure about using the enum value as is.

```
>> mode = SessionMode.interactive
>>> filter = {"session-type": mode}
>>> filter
{'session-type': <SessionMode.interactive: 'interactive'>}
>>> filter = {"session-type": str(mode)}
>>> filter
{'session-type': 'interactive'}
```
?

components/renku_data_services/notebooks/core_sessions.py
eikek: Yes these strings are on many places across the codebase. that is why I simply added another one. I don't want to make larger refactorings while working on this feature. I do agree that we should do it, though. I would rather create a different PR for that, wdyt?

components/renku_data_services/notebooks/core_sessions.py
eikek: Ah yes, thanks. I think I can remove it from annotations (left from earlier attempts).

components/renku_data_services/notebooks/core_sessions.py
eikek: Left over, removed.

components/renku_data_services/notebooks/models.py
eikek: I think I started with `SessionType` for the http api and `SessionMode` for the model to not get confused too much :). Then I used `SessionType` on amalthea for some reason.

For the strings: I can use the strings as required by amalthea. But then I'm risking that changing our model on data-services effects the communication to amalthea. I know it would save these two conversion functions. I usually prefer a more explicit variant for this case, because it crosses boundaries. But if there are strong opinions, I'm happy to change. Please let me know.

components/renku_data_services/notebooks/cr_amalthea_session.py
leafty: Note, the yaml has been updated, so maybe try to re-run the code generator?

test/bases/renku_data_services/data_api/test_notebooks.py
eikek: I agree, please see my other comment about doing this refactoring in a separate PR.

components/renku_data_services/notebooks/cr_amalthea_session.py
eikek: This is generated code, I don't know.

eikek: Thank you! I forgot to run the code generator again.

components/renku_data_services/notebooks/api/classes/k8s_client.py
sgaist: AFAIU [StrEnum](https://docs.python.org/3/library/enum.html#enum.StrEnum) values are also strings so unless the code below does precise type checking, it should work just fine.

components/renku_data_services/notebooks/models.py
sgaist: I can see the confusion problem but I would argue that having two different enum names representing the same concept also makes things a bit confusing. It looks like Amalthea and renku-data-services do not get along on the naming of things.

As for the string values, I have no issues with the conversion functions but in that case, I would document the class to state why the value differs between each other.

components/renku_data_services/notebooks/core_sessions.py
sgaist: Agreed for the follow up PR, that will make things cleaner.

components/renku_data_services/notebooks/api/classes/k8s_client.py
eikek: Ah ok, thanks for the pointer. The docs say, that the members can be "used in most of the same places that a string can be used". To me that sounds a bit unreliable. If I understand this, the members are implemented as a subclass of `str`, so it can be used in most places unless code checks for types more closely. I suppose when serializing to json it will work fine. I can change it to remove `str`. But what is so bad about using `str` here? To me it is more clear to create a `dict[str, str]` than `dict[str, Any]` - or is this not idiomatic in python?

sgaist: `dict[str, str]` is the right type in this case.

I haven't checked if MyPy screams about it but based on the Python documentation, it should not.

That said, it was to answer your doubts about the use of the StrEnum value, you can keep using `str`.

components/renku_data_services/notebooks/models.py
eikek: Yes it is taste or familiarity most of the time when discussing "readability" or "confusion". IMHO both projects don't need to synchronize on naming too much, they can evolve differently and data-services may not transport everything from/to amalthea exactly 1-1. I rename so that all three are the same name, but they still are different types. 

For the string values: I don't see why the 'obvious' assumption is that these values are the same as in amalthea. In my head, I would rather see the need to document why they must be the same if i remove these functions :-). It seems easy to change these later or add a new value in the data-services model, then unintentionally changing the requests to amalthea. But maybe not.

components/renku_data_services/notebooks/api.spec.yaml
```diff
@@ -210,6 +218,11 @@ components:
               description: Sentry trace ID for linking to corresponding log entries
           required: ["code", "message"]
       required: ["error"]
+    SessionType:
+      type: string
+      enum:
+        - interactive
+        - non-interactive
     SessionPostRequest:
```
leafty, line 226: Why not add `command_override` and `args_override` in this request? This would allow to easily launch a job from an image already used in a project (e.g. built from python code).

components/renku_data_services/notebooks/core_sessions.py
```diff
@@ -800,6 +839,9 @@ async def start_session(
     if session_location == SessionLocation.remote and not user.is_authenticated:
         raise errors.ValidationError(message="Anonymous users cannot start remote sessions.")
 
+    if session_location == SessionLocation.remote and launch_request.is_non_interactive:
+        raise errors.ValidationError(message="Non-Interactive sessions are not possible for remote sessions")
```
leafty:
```suggestion
        raise errors.ValidationError(message="Non-Interactive sessions are not supported for remote sessions")
```

There is no reason that this should not be possible.

components/renku_data_services/notebooks/models.py
```diff
@@ -155,7 +159,29 @@ class SessionExtraResources:
...
 
-    def concat(self, added_extras: "SessionExtraResources | None") -> "SessionExtraResources":
+    def __extra_to_sidecar_container(self, ec: ExtraContainer) -> InitContainer:
+        """Convert a ExtracContainer value to an InitContainer value.
+
+        Both types have the exact same shape, we only change the restartPolicy to 'Always' to mark it as a sidcar.
+        """
+        ic = ec.model_dump()
+        ic["restartPolicy"] = "Always"
+        return InitContainer.model_construct(**ic)
```
leafty: Why not use `.model_copy()`?

components/renku_data_services/notebooks/models.py
```diff
@@ -155,7 +159,29 @@ class SessionExtraResources:
...
+
+    def extra_container_as_sidecar(self) -> SessionExtraResources:
```
leafty:
```suggestion
    def extra_containers_as_sidecars(self) -> SessionExtraResources:
```

components/renku_data_services/notebooks/core_sessions.py
leafty: This leads to some weirdness now. Validate that the session type is the same then reject the request if the existing session is not of the same type.

leafty, line 828: Maybe the result of `renku_2_make_server_name()` should be different for jobs. Otherwise it is not possible to launch a session and a job from the same launcher.

components/renku_data_services/notebooks/api.spec.yaml
```diff
@@ -224,6 +237,8 @@ components:
           $ref: "#/components/schemas/SessionDataConnectorsOverrideList"
         env_variable_overrides:
           $ref: "#/components/schemas/EnvVariableOverrides"
+        session_type:
+          $ref: "#/components/schemas/SessionType"
```
leafty, line 241: Note: maybe we should have a `default_session_type` field on the session launcher to indicate a session launcher that works for jobs. Otherwise that session launcher will likely not work if started as an interactive session.

components/renku_data_services/solr/solr_client.py
leafty, line 1: Why the changes here? Maybe split this off into a separate PR.

components/renku_data_services/notebooks/blueprints.py
leafty, line 1: The returned session status for jobs seems wrong: I get `{ "status": { "state": "starting" } } }` from the data services API, but the Amalthea CR has: `{ "status": { "state": { "Suceeded" } } }` and the job is marked as `Completed`.

leafty (commented):
Also, other weirdness: if a job is completed and I send a new POST request to start a new job, I instead get the existing completed job back from the API. Maybe there should be a field added to the request body indicating that any existing job should be discarded.

components/renku_data_services/notebooks/api.spec.yaml
eikek: I thought to only add things currently implemented and defer other arguments for later, when they are being worked on. wdyt? Time is running out a bit and I'd like to get all these PRs done 😅

components/renku_data_services/notebooks/core_sessions.py
eikek: Thanks for the rewording. We decided to have a look at this later and error out in the first iteration. I'm myself not sure about the implications with remote sessions. But I was also thinking that it shouldn't be such a hard problem - when interactive sessions are fine, then jobs should be also good to go.

components/renku_data_services/notebooks/models.py
eikek: I simply didn't know better. How does this work when converting from `ExtraContainer` to `InitContainer`? Do you mean like this:
``` python
ic = ec.model_copy(update={"restartPolicy": "Always"}).model_dump()
return InitContainer.model_construct(**ic)
```

components/renku_data_services/notebooks/core_sessions.py
eikek: This is also something I didn't touch on purpose just yet. Do you think I should do it now? For now a launcher can be either/or.

components/renku_data_services/solr/solr_client.py
eikek: I can remove these changes, I thought they are small enough, so I included them. I did it to let my dev setup work with the new solr version.

components/renku_data_services/notebooks/api.spec.yaml
eikek: Yeah I see. There have been some discussions related to this. So in this PR I would like to refrain from changes related to having sessions for both purposes and leave it for later. If that makes sense, of course.

components/renku_data_services/notebooks/models.py
eikek: So I changed it to the above, not sure what you mean. I didn't find a nicer way to convert between types though they are structurally the same, not sure if python cares at all anyways :-)

leafty (commented):
Note: apparently you need to deploy with `renku=salimkayal-add-AUTHZ-config-to-k8s-watcher` @eikek.

components/renku_data_services/notebooks/api/classes/k8s_client.py
```diff
@@ -308,6 +314,26 @@ async def patch_session_tokens(
...
 
+    async def _get_pod_for_session(self, session: AmaltheaSessionV1Alpha1) -> Pod | None:
+        """Get the pod associated to the given session."""
+        sess_mode = SessionType.from_amalthea(session.spec.sessionType)
+        pod_name = f"{session.metadata.name}-0"
+        pod_gvk = GVK.from_kr8s_object(Pod)
+        result: K8sObject | None = None
+        if sess_mode == SessionType.non_interactive:
+            job_name = session.metadata.name
+            async for j in self.__client.list(K8sObjectFilter(gvk=pod_gvk, label_selector={"job-name": job_name})):
```
sgaist: Shouldn't that be a get rather than a list ?

components/renku_data_services/notebooks/api/classes/k8s_client.py
```diff
@@ -308,6 +314,26 @@ async def patch_session_tokens(
         await self.patch_statefulset_tokens(session_name, renku_tokens, safe_username)
         await self.patch_image_pull_secret(session_name, gitlab_token, safe_username)
 
+    async def _get_pod_for_session(self, session: AmaltheaSessionV1Alpha1) -> Pod | None:
+        """Get the pod associated to the given session."""
+        sess_mode = SessionType.from_amalthea(session.spec.sessionType)
+        pod_name = f"{session.metadata.name}-0"
+        pod_gvk = GVK.from_kr8s_object(Pod)
+        result: K8sObject | None = None
```
sgaist: nitpick: we have everything around related to pod, result feels a bit off.

components/renku_data_services/notebooks/core_sessions.py
```diff
@@ -955,18 +1003,29 @@ async def start_session(
...
+        env.extend(
+            [
+                SessionEnvItem(name="RENKU_BASE_URL_PATH", value=ingress_config.url_path),
+                SessionEnvItem(name="RENKU_BASE_URL", value=ingress_config.url),
+                SessionEnvItem(name="RENKU_SESSION_IP", value="0.0.0.0"),  # nosec B104
+                SessionEnvItem(name="RENKU_SESSION_PORT", value=f"{environment.port}"),
+                SessionEnvItem(name="RENKU_SESSION", value="1"),
+            ]
+        )
+    if not is_interactive:
```
sgaist: Since it's just one or the other I think we can simplify:
```suggestion
   else:
```

components/renku_data_services/notebooks/api.spec.yaml
leafty: :+1: We can collect these type of items for later.

components/renku_data_services/notebooks/core_sessions.py
leafty: 👍 Can be collected for later. Note that this may be quite important to allow the UX flow: "start a session; then use the rnk cli in the session to dispatch jobs using the same environment".

components/renku_data_services/notebooks/api.spec.yaml
leafty: 👍 Can be collected for later.

components/renku_data_services/notebooks/blueprints.py
leafty: I wonder if this is a code issue or if there is a de-sync from k8s watcher.

flake.nix
```diff
@@ -279,8 +279,9 @@
             SOLR_URL = "http://localhost:18983";
             SOLR_CORE = "renku-search-dev";
 
-            AMALTHEA_SESSIONS_VERSION = "refs/heads/eikek/feat-prove-active-session";
-            AMALTHEA_JS_VERSION = "refs/heads/eikek/feat-prove-active-session";
+            AMALTHEA_SESSIONS_VERSION = "refs/heads/eikek/non-interactive-session";
+            RENKU_ENV = "renku-ci-ds-1246";
+            K8S_NAMESPACE = "default";
```
sgaist, line 284: Shouldn't these have other default values ?

sgaist (changes requested):

components/renku_data_services/notebooks/api/classes/k8s_client.py
eikek: It is the _result_ of a query. What do you mean by "off"? If you give me your preferred name, I can rename it.

components/renku_data_services/notebooks/api/classes/k8s_client.py
eikek: I think a job could have multiple pods associated - how can I use a get with a given job name? The convention is one pod per session, so this stops at the first found.

components/renku_data_services/notebooks/api.spec.yaml
eikek: 🙏🏼 I think for the next iteration we should discuss again, there were some thoughts in previous meetings on how to proceed. This was one part.

Edit: I compiled your comments into this issue https://github.com/SwissDataScienceCenter/renku-data-services/issues/1277

flake.nix
eikek: This is only my personal dev environment for this pr.

components/renku_data_services/notebooks/api/classes/k8s_client.py
sgaist: You are correct, it can.

By default a Job is a `Non-parallel` job so it has only one pod. 

You can set it up to have parallel jobs or parallel jobs with a work-queue and thus multiple pods.

I don't think it's the case here, hence my question.

Just to be clear, I am not against doing a list, it was just to understand why it was done like that.

components/renku_data_services/notebooks/api/classes/k8s_client.py
sgaist: I was just thinking at `pod` since that's what you want from the query you are making just after.

components/renku_data_services/notebooks/blueprints.py
eikek: Hm, strange. I also see it with my tries. I would currently suspect being a cache issue. I'll try to find out more.

components/renku_data_services/notebooks/api/classes/k8s_client.py
eikek: After the query which returns a value of type `K8sObject` this object is converted into a value of type `Pod` (in `result.to_api_object`). So then I find it strange when something is called "pod" when just later it is converted into such a thing. But don't worry, I'll rename it to be "pod".

components/renku_data_services/notebooks/api/classes/k8s_client.py
eikek: Yes, amalthea will do a non-parallel job. So there should be just one pod per renku job. 

When I don't have the name of the pod, but only the job-name - how do I do a GET request with that information?

sgaist: Wait, there's something I misread and therefore misunderstood. I thought you were looking for the Job itself and then grab the pod out of it.

I don't think you can do what I had in mind with the client as is so no need to lose more of your time on this.

I am sorry for the noise.

components/renku_data_services/notebooks/api/classes/k8s_client.py
```diff
@@ -308,6 +314,26 @@ async def patch_session_tokens(
...
 
+    async def _get_pod_for_session(self, session: AmaltheaSessionV1Alpha1) -> Pod | None:
+        """Get the pod associated to the given session."""
+        sess_mode = SessionType.from_amalthea(session.spec.sessionType)
+        pod_name = f"{session.metadata.name}-0"
+        pod_gvk = GVK.from_kr8s_object(Pod)
+        pod: K8sObject | None = None
+        if sess_mode == SessionType.non_interactive:
+            job_name = session.metadata.name
+            async for j in self.__client.list(K8sObjectFilter(gvk=pod_gvk, label_selector={"job-name": job_name})):
```
leafty: The legacy label is deprecated, use the reserved one:

```suggestion
            async for j in self.__client.list(K8sObjectFilter(gvk=pod_gvk, label_selector={"batch.kubernetes.io/job-name": job_name})):
```

Reference: https://kubernetes.io/docs/reference/labels-annotations-taints/#batchkubernetesio-job-name

eikek: Should I change other places as well? It's not related to this feature, though.

components/renku_data_services/notebooks/core_sessions.py
```diff
@@ -789,7 +828,13 @@ async def start_session(
     )
     existing_session = await nb_config.k8s_v2_client.get_session(name=server_name, safe_username=user.id)
     if existing_session is not None:
-        return existing_session, False
+        if launch_request.session_type == SessionType.from_amalthea(existing_session.spec.sessionType):
+            return existing_session, False
+        else:
+            raise errors.ValidationError(
+                message=f"Session exists with type={existing_session.spec.sessionType}, "
+                "while request contains {launch_request.session_type}."
```
leafty: missing formatting

```suggestion
                f"while request contains {launch_request.session_type}."
```

components/renku_data_services/notebooks/blueprints.py
eikek: I found the cause. The succeeded status wasn't propagated into the apispec response structures. And the code converting the amalthea data into response data had a fallback (else branch) setting the status to "starting". So then the new case was handled in that branch as well. (https://github.com/SwissDataScienceCenter/renku-data-services/pull/1246/commits/bd48dc137cae3cd1ab3b20e59b2a6d5b74c5e456)

leafty (approved):
LGTM in the current state 👍 👏

sgaist (approved):

PR #1248, opened 2026-03-30
Feat: move resource pools check to authz

SalimKayal:
## Summary
This PR implements ResourcePool authorization in SpiceDB/Authzed, enabling centralized access control while maintaining feature parity with the DB-based implementation.

## What Changed
Authzed Schema (v9):
- Added resource_pool resource type with relations: member, prohibited, public_viewer, resource_pool_platform
- Permissions: use = (member + public_viewer + admin) - prohibited, write = admin-only
Supporting Code:
- Extended ResourceType enum with resource_pool
- Updated _AuthzConverter.resource_pool() and related methods to handle int IDs
- Enhanced permission checking signatures throughout authz module
Migration:
- Alembic migration cd424c01676e deploys Authzed schema v9
- Schema-only migration (data migration deferred for safety)

## Tests
- Authorization tests: member/prohibited/public patterns, admin bypass, isolation
- Schema Validation: SpiceDB `zed validate` confirms v9 schema correctness

---

**Deferred Items**
1. Core Logic Updates: Update CRC core and db modules to use Authzed for membership checks
2. API Refactoring: Update blueprints to leverage updated core logic
3. Cleanup Migration: Remove old DB tables after Phase 1 verification
4. Authz Relationship: Add groups and projects for authorization.

---

## PR stack

- (base/this) #1248
  - #1256 
  
---

/deploy renku=salimkayal-add-AUTHZ-config-to-k8s-watcher

### DIFF
--- a/components/renku_data_services/authz/schemas.py
+++ b/components/renku_data_services/authz/schemas.py
@@ -763,3 +763,103 @@ v8 = AuthzSchemaMigration(
     up=[WriteSchemaRequest(schema=_v8)],
     down=[WriteSchemaRequest(schema=_v7)],
 )
+
+_v9 = """\
+definition user {}
+
+definition group {
+    relation group_platform: platform
+    relation owner: user
+    relation editor: user
+    relation viewer: user
+    relation public_viewer: user:* | anonymous_user:*
+    permission read = public_viewer + read_children
+    permission read_children = viewer + write
+    permission write = editor + delete
+    permission change_membership = delete
+    permission delete = owner + group_platform->is_admin
+    permission non_public_read = owner + editor + viewer - public_viewer
+    permission exclusive_owner = owner
+    permission exclusive_editor = editor
+    permission exclusive_member = viewer + editor + owner
+    permission direct_member = owner + editor + viewer
+}
+
+definition user_namespace {
+    relation user_namespace_platform: platform
+    relation owner: user
+    relation public_viewer: user:* | anonymous_user:*
+    permission read = public_viewer + read_children
+    permission read_children = delete
+    permission write = delete
+    permission delete = owner + user_namespace_platform->is_admin
+    permission non_public_read = owner - public_viewer
+    permission exclusive_owner = owner
+    permission exclusive_member = owner
+    permission direct_member = owner
+}
+
+definition anonymous_user {}
+
+definition platform {
+    relation admin: user
+    permission is_admin = admin
+}
+
+definition project {
+    relation project_platform: platform
+    relation project_namespace: user_namespace | group
+    relation owner: user
+    relation editor: user
+    relation viewer: user
+    relation public_viewer: user:* | anonymous_user:*
+    permission read = public_viewer + read_children
+    permission read_children = viewer + write + project_namespace->read_children
+    permission write = editor + delete
+    permission change_membership = delete
+    permission delete = owner + project_platform->is_admin + project_namespace->delete
+    permission non_public_read = owner + editor + viewer + project_namespace->read_children - public_viewer
+    permission exclusive_owner = owner + project_namespace->exclusive_owner
+    permission exclusive_editor = editor
+    permission exclusive_member = owner + editor + viewer + project_namespace->exclusive_member
+    permission direct_member = owner + editor + viewer
+}
+
+definition data_connector {
+    relation data_connector_platform: platform
+    relation data_connector_namespace: user_namespace | group | project
+    relation linked_to: project
+    relation owner: user
+    relation editor: user
+    relation viewer: user
+    relation public_viewer: user:* | anonymous_user:*
+    permission read = public_viewer + viewer + write + data_connector_namespace->read_children
+    permission write = editor + delete
+    permission change_membership = delete
+    permission delete = owner + data_connector_platform->is_admin + data_connector_namespace->delete
+    permission non_public_read = owner + editor + viewer + data_connector_namespace->read_children - public_viewer
+    permission exclusive_owner = owner + data_connector_namespace->exclusive_owner
+    permission exclusive_editor = editor
+    permission exclusive_member = owner + editor + viewer + data_connector_namespace->exclusive_member
+    permission direct_member = owner + editor + viewer
+}
+
+definition resource_pool {
+    relation resource_pool_platform: platform
+    relation member: user
+    relation prohibited: user
+    relation public_user: user:* | anonymous_user:*
+    permission use = (member + public_user - prohibited) + resource_pool_platform->is_admin
+    permission write = resource_pool_platform->is_admin
+}"""
+"""Adds the resource_pool definition for Authzed authorization."""
+
+v9 = AuthzSchemaMigration(
+    up=[WriteSchemaRequest(schema=_v9)],
+    down=[
+        DeleteRelationshipsRequest(
+            relationship_filter=RelationshipFilter(resource_type=ResourceType.resource_pool.value)
+        ),
+        WriteSchemaRequest(schema=_v8),
+    ],
+)
--- a/components/renku_data_services/authz/schemas.py
+++ b/components/renku_data_services/authz/schemas.py
@@ -846,10 +846,10 @@ definition data_connector {
 
 definition resource_pool {
     relation resource_pool_platform: platform
-    relation member: user
+    relation viewer: user
     relation prohibited: user
-    relation public_user: user:* | anonymous_user:*
-    permission use = (member + public_user - prohibited) + resource_pool_platform->is_admin
+    relation public_viewer: user:* | anonymous_user:*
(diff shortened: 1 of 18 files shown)


sgaist (changes requested):
I am no specialist of Authz but the code in itself looks good.

Just some minor things.

components/renku_data_services/authz/schemas.py
```diff
@@ -763,3 +763,103 @@ def generate_v4(public_project_ids: Iterable[str]) -> AuthzSchemaMigration:
...
+    permission non_public_read = owner + editor + viewer + data_connector_namespace->read_children - public_viewer
+    permission exclusive_owner = owner + data_connector_namespace->exclusive_owner
+    permission exclusive_editor = editor
+    permission exclusive_member = owner + editor + viewer + data_connector_namespace->exclusive_member
+    permission direct_member = owner + editor + viewer
+}
+
+definition resource_pool {
+    relation resource_pool_platform: platform
+    relation member: user
```
leafty: Question: I don't see groups or projects linked to resource_pool here. Is that coming in another PR?

SalimKayal: yes, the goal is to first bring it to feature parity then add the new features.

components/renku_data_services/authz/schemas.py
```diff
@@ -763,3 +763,103 @@ def generate_v4(public_project_ids: Iterable[str]) -> AuthzSchemaMigration:
...
+    relation prohibited: user
+    relation public_user: user:* | anonymous_user:*
+    permission use = (member + public_user - prohibited) + resource_pool_platform->is_admin
```
SalimKayal: @leafty Quick question : currently admins do not see resource_pools they are not member of. this PR changes that (and makes some tests under `test/bases/renku_data_services/data_api/test_resource_pool.py` fail (the next pr will fix the tests by changing the user used to test to a non admin user)). IMHO an admin should see all resource pools but maybe that's not intended ? Would you please give me some input there ?

leafty: We currently have this when doing `GET /resource_pools` for admins:
```python
        case True, True:
            # The user is logged in and is an admin, they can see all resource pools
            pass
```

So an admin sees all resource pools at the moment.

components/renku_data_services/migrations/versions/cd424c01676e_add_resource_pool_authorization_schema.py
```diff
@@ -0,0 +1,160 @@
...
+
+        default_pool_ref = _AuthzConverter.resource_pool(default_pool_id)
+        for user_row in no_access_users:
+            user_ref = SubjectReference(object=ObjectReference(object_type="user", object_id=user_row[0]))
+            updates.append(
+                RelationshipUpdate(
+                    operation=RelationshipUpdate.OPERATION_TOUCH,
+                    relationship=Relationship(
+                        resource=default_pool_ref,
+                        relation="prohibited",
```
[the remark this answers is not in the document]
SalimKayal: fixed

test/components/renku_data_services/authz/test_authorization.py
```diff
@@ -370,3 +416,131 @@ async def test_listing_non_public_projects(app_manager_instance: DependencyManag
...
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("public_pool", [True, False])
+async def test_resource_pool_member_access(
+    app_manager_instance: DependencyManager, bootstrap_admins, public_pool: bool
+) -> None:
+    """An explicit member can use a pool regardless of visibility; non-members still follow public rules."""
+    authz = app_manager_instance.authz
+    updates = _pool_updates(POOL_ID, public=public_pool) + [_rel(POOL_ID, "member", regular_user1.id)]
```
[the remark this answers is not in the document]
SalimKayal: fixed

components/renku_data_services/authz/schemas.py
```diff
@@ -763,3 +763,103 @@ def generate_v4(public_project_ids: Iterable[str]) -> AuthzSchemaMigration:
...
+    permission exclusive_editor = editor
+    permission exclusive_member = owner + editor + viewer + data_connector_namespace->exclusive_member
+    permission direct_member = owner + editor + viewer
+}
+
+definition resource_pool {
+    relation resource_pool_platform: platform
+    relation member: user
+    relation prohibited: user
+    relation public_user: user:* | anonymous_user:*
```
olevski: question: Groups and projects have a very similar property under a slightly different name - `public_viewer` can we use that here as well? Is there a reason why you picked a different name for resource pools?

olevski: I guess the reason is the changes between the resource pool having a `member` relation and the other resources having a `viewer` relation? Can we reuse the relation name as well then. The permission can still be `viewer` but the permission is use?

components/renku_data_services/migrations/versions/cd424c01676e_add_resource_pool_authorization_schema.py
olevski, line 1: I am wondering what is the easiest way to test this to avoid issues. I think we can count the number of pools and members on prod. Then populate that many with a script and then migrate in a ci deployment.

olevski (changes requested):
Overarching comment the target is `main` can this go to a feature branch? Mainly because I think we should test the migration on a ci deployment we artificially put a lot of users into. And then see if the migration is quick enough. So for now we can merge into a feature branch and you can continue with your work and then when we merge the feature branch to main we do the test.

olevski:
For example on renkulab.io we have a total of 598 resource pool member entries spread across 34 resource pools. We can use this info to test.

components/renku_data_services/authz/schemas.py
SalimKayal: It has to do with semantics. I don't like the term viewer here because the permission has nothing to do with viewing but if you prefer having consistency across the schemas we can do that. wdyt ?

components/renku_data_services/migrations/versions/cd424c01676e_add_resource_pool_authorization_schema.py
SalimKayal: That would be a possibility. I will try that locally and report. Thanks for the amounts.

Seeing the numbers I'm not too afraid by the time it will take though.

test/bases/renku_data_services/data_api/test_migrations.py
SalimKayal, line 1: @olevski this should test for the migration. Does it seem fine ?

components/renku_data_services/authz/schemas.py
leafty: I think it would be nice to keep the same verbs we use with the other resources.

Also, when a user can "read" a resource pool, it means they can actually read  the non-public part of the resource. The best example is the resource pool usage quota and current usage. A user without the "read" access should not be able to read those fields/sub-resources from the resource pool.

components/renku_data_services/migrations/versions/cd424c01676e_add_resource_pool_authorization_schema.py
SalimKayal: is [test/bases/renku_data_services/data_api/test_migrations.py](https://github.com/SwissDataScienceCenter/renku-data-services/pull/1248/files/3c8037f425281269daa85faa717fa8aa9fd2b39b#diff-199250d2331a101b3c4bdebef8d92d1a564a76ddf4caf63810f111cd758849c9) enough ?

components/renku_data_services/authz/schemas.py
olevski: > I don't like the term viewer here because the permission has nothing to do with viewing but if you prefer having consistency across the schemas we can do that. wdyt ?

 I see your point but I dont think the disconnect is that big to warrant adding a whole new relation.

test/bases/renku_data_services/data_api/test_migrations.py
olevski: Yes that is fine. Thanks.

components/renku_data_services/authz/schemas.py
olevski: > I think it would be nice to keep the same verbs we use with the other resources.

And I agree with Flora I think it is also ok to just grit our teeth and use the same verb - `read` as all the other resources too.

components/renku_data_services/migrations/versions/cd424c01676e_add_resource_pool_authorization_schema.py
olevski: Ok thank you for adding that!

components/renku_data_services/authz/schemas.py
SalimKayal: refactored

olevski (approved):

leafty (changes requested):
Can we merge this into a feature branch? This code is not "usable" as is in `main`.

leafty (approved):

PR #1249, opened 2026-03-31
feat: use self-minted tokens for sessions

leafty:
Use self-minted tokens for sessions.

Details:
* git-proxy: use internal access and refresh tokens
* remote-session-controller: use internal access and refresh tokens
* csi-rclone: use internal refresh tokens with long expiry (24 hours)
  * the `POST /api/data/oauth2/connections/:connection_id/token_endpoint` API endpoint is updated to support only internal refresh tokens

PR Stack:
* (base) #1258
  * (this) #1249
    * (tip) #1259


components/renku_data_services/authn/chained.py
```diff
@@ -0,0 +1,37 @@
...
+        """Authenticate by going through the authenticator chain."""
+
+        user: AnyAPIUser | None = None
+        for auth in self.chain:
+            user = await auth.authenticate(access_token=access_token, request=request)
+            if user.is_authenticated:
+                return user
+
+        # NOTE: user is not None since there is at least one authenticator in the chain
+        assert user is not None
```
sgaist, line 36: Is it a test leftover ?

Otherwise it should be turned into an exception.

components/renku_data_services/connected_services/core.py
```diff
@@ -116,108 +115,76 @@ async def handle_oauth2_token_refresh(
...
+        user = base_models.AuthenticatedAPIUser(
+            is_admin=False,
+            id=parsed_renku_refresh_token["sub"],
+            access_token="",  # nosec B106
+            full_name=parsed_renku_refresh_token.get("name"),
+            first_name=parsed_renku_refresh_token.get("given_name"),
+            last_name=parsed_renku_refresh_token.get("family_name"),
+            email=parsed_renku_refresh_token["email"],
+            access_token_expires_at=None,
+            roles=[],
```
sgaist: Nitpick: shouldn't that be done outside the try block ?

Even if not *big* I usually try to keep the try except blocks to the minimal amount of code that can trigger an exception.

components/renku_data_services/notebooks/api/amalthea_patches/git_proxy.py
```diff
@@ -60,6 +56,18 @@ async def main_container(
...
+        client.V1EnvVar(name=f"{prefix}RENKU_ACCESS_TOKEN", value=internal_access_token),
+        client.V1EnvVar(name=f"{prefix}RENKU_REFRESH_TOKEN", value=internal_refresh_token),
+        # TODO: can we use a sanic method to derive this URL?
+        client.V1EnvVar(
+            name=f"{prefix}RENKU_TOKEN_URL",
+            value="https://" + config.sessions.ingress.host + "/api/data/internal/authentication/token",
+        ),
+        client.V1EnvVar(
+            name=f"{prefix}REFRESH_CHECK_PERIOD_SECONDS",
+            value="60",
```
sgaist: Should we have such values in constants ?

components/renku_data_services/notebooks/core_sessions.py
```diff
@@ -676,24 +689,25 @@ def get_remote_secret(
...
+        "RSC_AUTH_KIND": "renku_v2",
         "RSC_AUTH_TOKEN_URI": remote_provider.access_token_url,
-        "RSC_AUTH_RENKU_ACCESS_TOKEN": user.access_token,
-        "RSC_AUTH_RENKU_REFRESH_TOKEN": user.refresh_token,
-        "RSC_AUTH_RENKU_TOKEN_URI": renku_auth_token_uri,
-        "RSC_AUTH_RENKU_CLIENT_ID": config.sessions.git_proxy.renku_client_id,
-        "RSC_AUTH_RENKU_CLIENT_SECRET": config.sessions.git_proxy.renku_client_secret,
+        "RSC_AUTH_RENKU_TOKEN_URI": "https://"
+        + config.sessions.ingress.host
+        + "/api/data/internal/authentication/token",
```
sgaist: We are using f strings everywhere:
```suggestion
        "RSC_AUTH_RENKU_TOKEN_URI": f"https://{config.sessions.ingress.host}/api/data/internal/authentication/token",
```

sgaist (changes requested):

components/renku_data_services/authn/chained.py
leafty: I don't think I follow: the `assert` will raise in tests but not in production mode and this `assert` makes a type assertion: `user is None` cannot happen, so the fact that nothing runs in production is OK.

components/renku_data_services/authn/chained.py
```diff
@@ -0,0 +1,37 @@
...
+    Authenticator,
+)
+from renku_data_services.errors import errors
+
+
+@dataclass(frozen=True, kw_only=True)
+class ChainedAuthenticator(Authenticator[AnyAPIUser]):
+    """Chain authenticators until a user is authenticated or all authenticators are tried."""
+
+    token_field: str = "__not_used__"
```
olevski, line 19: This field should also never be changed here, right? And remember to import `Final` from `typing` and `field` from `dataclasses`

```suggestion
    token_field: Final[str] = field(default="__not_used__", init=False)
```

olevski: ...not sure if `init=False` makes it a bit too weird. Like it is a child of Authenticator but the API is now different. Maybe that is too much. And just `Final` is enough.

components/renku_data_services/notebooks/api/amalthea_patches/git_proxy.py
```diff
@@ -60,6 +60,17 @@ async def main_container(
                 [dict(id=provider.id, access_token_url=provider.access_token_url) for provider in git_providers]
             ),
         ),
+        client.V1EnvVar(name=f"{prefix}RENKU_AUTHENTICATION_VERSION", value="v2"),
```
olevski, line 63: That's neat. We can handle changes to the internal auth better if there are any in the future.

olevski (approved):
One small change but we can live without that also. Up to you if you want to use it or not.

components/renku_data_services/notebooks/api/amalthea_patches/git_proxy.py
leafty: Yes, though I will probably wipe the old code after a release. 😄

components/renku_data_services/authn/chained.py
leafty: The value of the field is never used, so it doesn't matter if it is changed. The value is there to indicate that it's not used directly but rather each authenticator in the chain will use its own value.

PR #1253, opened 2026-04-08
fix: cloning repos token urls

olevski:
On the meteo swiss remote cluster they block outgoing access if it is not on https. And we always used `http` to template out the renku url. This makes it so that the renku url is https if we use tls or http if we dont.

/deploy

### DIFF
--- a/components/renku_data_services/notebooks/config/dynamic.py
+++ b/components/renku_data_services/notebooks/config/dynamic.py
@@ -7,7 +7,7 @@ from collections.abc import Sequence
 from dataclasses import dataclass, field
 from enum import Enum
 from io import StringIO
-from typing import Any, ClassVar, Self, Union
+from typing import Any, ClassVar, Literal, Self, Union
 
 import yaml
 from kubernetes import client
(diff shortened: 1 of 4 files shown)


sgaist (dismissed):

components/renku_data_services/notebooks/config/dynamic.py
```diff
@@ -256,6 +256,16 @@ def from_env(cls) -> Self:
...
 
+    @property
+    def scheme(self) -> Literal["https"] | Literal["http"]:
+        if self.tls_secret or self.use_default_cluster_tls_cert:
+            return "https"
+        return "http"
+
+    @property
+    def renku_url(self) -> str:
+        return f"{self.scheme}::{self.host}"
```
leafty, line 267: I'm not sure why we derive that the Renku API URL is http or https based on the session config for ingresses. Maybe the Renku API URL should be a config field.

olevski: Ok I will require it as a full URL and only if that is not defined fallback to deriving it from the ingress. Just to avoid breaking the deployments with the data service not finding the renku url. In a future release where the helm chart includes the renku url env var we can remove the fallback.

leafty: I think we have that logic elsewhere in `core_sessions.py`, so we can try to clean up a bit the legacy "NB_CONFIG" later.

leafty (dismissed):

leafty (approved):

PR #1254, opened 2026-04-08
fix: redirect to frontend with error params on oauth callback failure

andre-code:
Currently when OAuth fails the user gets stuck on the API callback endpoint with no way for the client app to handle the error. This change makes the callback always redirect to the frontend, passing along the error type and a sanitized message as query params so the client can display appropriate feedback to the user.

/deploy


components/renku_data_services/connected_services/blueprints.py
```diff
@@ -130,9 +130,20 @@ def authorize_callback(self) -> BlueprintFactoryResponse:
 
         async def _callback(request: Request) -> HTTPResponse:
             params = CallbackParams.model_validate(dict(request.query_args))
+            callback_query = dict(request.query_args)
```
sgaist: Nitpick: I would call the variable `callback_query_args` 
This keeps consistency of what that variable represent.

```suggestion
            callback_query_args = dict(request.query_args)
            params = CallbackParams.model_validate(callback_query_args)
```

components/renku_data_services/connected_services/blueprints.py
```diff
@@ -130,9 +130,20 @@ def authorize_callback(self) -> BlueprintFactoryResponse:
...
 
             callback_url = self._get_callback_url(request)
 
+            if callback_query.get("error"):
+                next_url = await self.connected_services_repo.get_oauth2_connection_next_url_by_state(params.state)
+                if next_url:
+                    return redirect(to=self._append_query_params(next_url, callback_query))
```
sgaist: To match my suggestion above:

```suggestion
            if callback_query_args.get("error"):
                next_url = await self.connected_services_repo.get_oauth2_connection_next_url_by_state(params.state)
                if next_url:
                    return redirect(to=self._append_query_params(next_url, callback_query_args))
```

components/renku_data_services/connected_services/blueprints.py
```diff
@@ -156,6 +167,17 @@ def _get_callback_url(self, request: Request) -> str:
...
 
+    @staticmethod
+    def _append_query_params(url: str, query: dict[str, str]) -> str:
+        allowed_keys = ("error", "error_description", "state", "error_uri", "code", "iss")
+        params = [(k, v) for k, v in query.items() if k in allowed_keys and v]
+        if not params:
+            return url
+        parsed = urlparse(url)
+        existing_params = parse_qsl(parsed.query, keep_blank_values=True)
+        merged_query = urlencode([*existing_params, *params])
```
sgaist: In the same spirit as above:
```suggestion
    def _append_query_params(url: str, query_args: dict[str, str]) -> str:
        allowed_keys = ("error", "error_description", "state", "error_uri", "code", "iss")
        allowed_params = [(k, v) for k, v in query_args.items() if k in allowed_keys and v]
        if not allowed_params:
            return url
        parsed = urlparse(url)
        existing_params = parse_qsl(parsed.query, keep_blank_values=True)
        merged_query = urlencode([*existing_params, *allowed_params])
```

sgaist (changes requested):
This looks good, just some small nitpick for readability.

olevski:
@andre-code I am a bit confused about this. A few questions:
1. How does the UI know it should render a nice error page?
2. How do you know there is an error? The external service adds error query parameter to the callback url?
3. There is a TODO in line 153 in components/renku_data_services/connected_services/blueprints.py that says we should add a redirect to an error page. I was under the impression that is what you are implementing. But that code is untouched.

andre-code:
> @andre-code I am a bit confused about this. A few questions:
> 
> 1. How does the UI know it should render a nice error page?
> 2. How do you know there is an error? The external service adds error query parameter to the callback url?
> 3. There is a TODO in line 153 in components/renku_data_services/connected_services/blueprints.py that says we should add a redirect to an error page. I was under the impression that is what you are implementing. But that code is untouched.

HI @olevski , sure, let me clarify.

1. The UI detects the error through the query parameters we forward in the redirect. When the callback receives an error from the provider, we redirect back to the frontend with those parameters included, so the UI can render the appropriate page based on their presence.

2. Yes, exactly. The provider appends an `error` query parameter to the callback URL when something goes wrong (e.g. the user denied access).

3. The TODO on line 153 is actually a different scenario from what I am addressing here. That branch handles `OAuthHttpFactoryError`, which covers internal failures like missing state or no matching connection row. My change handles the provider error case, which happens earlier in the flow before `fetch_token` is even called. These are two distinct failure paths. The TODO remains valid but is out of scope for this PR since it would require changes inside `fetch_token` to surface a `next_url` from that branch, which is a broader piece of work.

sgaist (commented):
LGTM Code wise

olevski:
> The UI detects the error through the query parameters we forward in the redirect. When the callback receives an error from the provider, we redirect back to the frontend with those parameters included, so the UI can render the appropriate page based on their presence.

Ok my bad Andrea. I did not know these were part of the Oauth2 standard. That was the source of my confusion.

components/renku_data_services/connected_services/blueprints.py
```diff
@@ -129,10 +129,21 @@ def authorize_callback(self) -> BlueprintFactoryResponse:
         """OAuth2 authorization callback."""
 
         async def _callback(request: Request) -> HTTPResponse:
-            params = CallbackParams.model_validate(dict(request.query_args))
+            callback_query_args = dict(request.query_args)
+            params = CallbackParams.model_validate(callback_query_args)
```
olevski: suggestion: Can you please add `error`, `error_description` and whatever other fields you are using as optional parameters to the apispec which will make them appear as optional field in the `CallbackParams` pydantic model. Then you can use the `params` object directly and not have to check for keys in the query params dictionary.

olevski (changes requested):
Thanks for the clarification @andre-code. I just have one small request for an update of the apispec to reflect the query params we support.

olevski (dismissed):
Thanks Andrea!

olevski:
You just need to rerun that one acceptance test and you will be good to go. I am pretty sure it is just flaky.

sgaist (dismissed):

sgaist (approved):

PR #1255, opened 2026-04-09
feat: add renku-self authentication

leafty:
Details:
* Add the `RenkuSelfAuthenticator` class which authenticates internal tokens. (This is a new `authenticator`).
* Add the `RenkuSelfTokenMint` class which can create internal tokens. Tokens are signed with the `HS512` algorithm.
* Add the `POST /api/data/internal/authentication/token` endpoint to renew internal tokens. This endpoint has the same OpenAPI definition as `POST /api/data/oauth2/connections/:connection_id/token_endpoint`.
OAuth 2.0 token endpoint to support applications running in sessions
* Update the `pyjwt` dependency.

Note: the next PRs will make use of the new internal tokens for `git-proxy`, `remote-session-controller` and `csi-rclone`.

Parameters

_Merging into the `build/feat-self-mint-tokens` feature branch_

/deploy #notest renku=build/gateway-improvements


components/renku_data_services/authn/renku.py
```diff
@@ -0,0 +1,211 @@
...
+                full_name=parsed.get("name"),
+                first_name=parsed.get("given_name"),
+                last_name=parsed.get("family_name"),
+                email=email,
+                access_token_expires_at=datetime.fromtimestamp(exp) if exp is not None else None,
+                roles=[],
+            )
+            request.ctx.renku_authenticator = f"{self.__class__.__module__}.{self.__class__.__qualname__}"
+            request.ctx.renku_user = user
+            request.ctx.renku_parsed_access_token = parsed
```
leafty:
```suggestion
```

components/renku_data_services/authn/renku.py
```diff
@@ -0,0 +1,208 @@
...
+                full_name=parsed.get("name"),
+                first_name=parsed.get("given_name"),
+                last_name=parsed.get("family_name"),
+                email=email,
+                access_token_expires_at=datetime.fromtimestamp(exp) if exp is not None else None,
+                roles=[],
+            )
+            return user
+
+        # Try to get an anonymous user ID if the validation of keycloak credentials failed
```
leafty:
```suggestion
        # Try to get an anonymous user ID from headers
```

components/renku_data_services/app_config/config.py
```diff
@@ -99,3 +103,68 @@ def from_env(cls) -> TrustedProxiesConfig:
...
+        default_refresh_token_expiration = (
+            timedelta(seconds=int(default_refresh_token_expiration_str))
+            if default_refresh_token_expiration_str
+            else timedelta(hours=1)
+        )
+        long_refresh_token_expiration_str = os.environ.get("INTERNAL_AUTHN_LONG_REFRESH_TOKEN_EXPIRATION_SECONDS")
+        long_refresh_token_expiration = (
+            timedelta(seconds=int(long_refresh_token_expiration_str))
+            if long_refresh_token_expiration_str
+            else timedelta(hours=25)
```
leafty:
```suggestion
            else timedelta(hours=24)
```

components/renku_data_services/app_config/config.py
```diff
@@ -99,3 +103,68 @@ def from_env(cls) -> TrustedProxiesConfig:
...
+                secret_key=secret_key,
+                default_access_token_expiration=default_access_token_expiration,
+                default_refresh_token_expiration=default_refresh_token_expiration,
+                long_refresh_token_expiration=long_refresh_token_expiration,
+                issuer=issuer,
+                audience=audience,
+            )
+
+        secret_key_path = os.environ.get("INTERNAL_AUTHN_SECRET_KEY_PATH")
+        if secret_key_path is None:
```
sgaist: Since an environment variable can be empty, both None and empty string should be checked so I would suggest the following:
```suggestion
        secret_key_path = os.environ.get("INTERNAL_AUTHN_SECRET_KEY_PATH", "")
        if not secret_key_path:
```

components/renku_data_services/authn/api/api.spec.yaml
```diff
@@ -0,0 +1,110 @@
+openapi: 3.0.2
+info:
+  title: Renku Data Services API
+  description: |
+    This service is the main backend for Renku. It provides information about users, projects,
+    cloud storage, access to compute resources and many other things.
```
sgaist, line 6: Shouldn't that rather describe the authentication part ?

components/renku_data_services/authn/api/api.spec.yaml
```diff
@@ -0,0 +1,110 @@
...
+      description: The schema for all 4xx and 5xx responses
+      content:
+        "application/json":
+          schema:
+            $ref: "#/components/schemas/ErrorResponse"
+
+  securitySchemes:
+    oidc:
+      type: openIdConnect
+      openIdConnectUrl: /auth/realms/Renku/.well-known/openid-configuration
```
sgaist: Should it really point to Keycloak ?

components/renku_data_services/authn/api/apispec_base.py
```diff
@@ -0,0 +1,12 @@
...
+from pydantic import BaseModel
+
+
+class BaseAPISpec(BaseModel):
+    """Base API specification."""
+
+    class Config:
+        """Enables orm mode for pydantic."""
+
+        from_attributes = True
```
sgaist: `class Config` is deprecated and generates warning

```suggestion
    # Enables orm mode for pydantic.
    model_config = ConfigDict(
        from_attributes=True,
    )
```

components/renku_data_services/authn/api/blueprints.py
```diff
@@ -0,0 +1,73 @@
...
+            user = base_models.AuthenticatedAPIUser(
+                is_admin=False,
+                id=parsed_token["sub"],
+                access_token="",  # nosec B106
+                full_name=parsed_token.get("name"),
+                first_name=parsed_token.get("given_name"),
+                last_name=parsed_token.get("family_name"),
+                email=parsed_token["email"],
+                access_token_expires_at=None,
+                roles=[],
```
sgaist, line 47: Wouldn't it make sense to have roles such as admin as part of the token ?

components/renku_data_services/authn/renku.py
```diff
@@ -0,0 +1,208 @@
...
+                raise errors.UnauthorizedError(
+                    message="Your credentials are invalid or expired, please log in again."
+                ) from None
+
+            token_type = parsed.get("typ")
+            if str(token_type).lower() != "bearer":
+                raise errors.UnauthorizedError() from None
+
+            user = AuthenticatedAPIUser(
+                is_admin=False,
```
sgaist, line 71: The authenticated user is never an admin ?

test/bases/renku_data_services/data_api/test_authn_api.py
```diff
@@ -0,0 +1,105 @@
...
+    new_refresh_token = result["refresh_token"]
+    payload = {
+        "grant_type": "refresh_token",
+        "refresh_token": new_refresh_token,
+    }
+    _, response = await sanic_client.post("/api/data/internal/authentication/token", data=payload)
+    assert response.status_code == 200, response.text
+    assert isinstance(response.json, dict)
+    assert isinstance(response.json.get("access_token"), str)
+    assert response.json["access_token"] != ""
```
sgaist, line 82: Shouldn't we also check that the access token has changed ?
```suggestion
    assert response.json["access_token"] != ""
    assert response.json["access_token"] != new_access_token
```

test/bases/renku_data_services/data_api/test_authn_api.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+@pytest.mark.asyncio
+async def test_post_internal_token_reject_access_token(
+    sanic_client: SanicASGITestClient, app_manager_instance: "DependencyManager", regular_user: UserInfo
+) -> None:
+    user = base_models.AuthenticatedAPIUser(
+        id=regular_user.id,
+        email=regular_user.email or "",
+        access_token="",
```
sgaist, line 93: What about parametrizing the test to check both empty and non-empty invalid content ?

test/components/renku_data_services/authn/test_renku.py
```diff
@@ -0,0 +1,246 @@
...
+
+@pytest.fixture
+async def internal_token_mint(monkeypatch: "MonkeyPatch") -> RenkuSelfTokenMint:
+    monkeypatch.setenv("DUMMY_STORES", "true")
+    internal_authn_config = InternalAuthenticationConfig.from_env()
+    return RenkuSelfTokenMint.from_config(config=internal_authn_config)
+
+
+def test_make_payload(internal_token_mint: RenkuSelfTokenMint) -> None:
+    user = AuthenticatedAPIUser(
```
sgaist: Since Jane is use through multiple tests, making a fixture with them would reduce duplication.

sgaist (changes requested):

components/renku_data_services/authn/api/api.spec.yaml
leafty: No, all the API specs are merged into a single one and the common fields must have the same value.

components/renku_data_services/authn/api/blueprints.py
leafty: No: these tokens are to be used in sessions (instead of sending Keycloak refresh tokens and Keycloak client credentials -- which is bad practice). There is no expectation of performing admin actions from inside a session, so role mapping is not needed.

components/renku_data_services/authn/renku.py
leafty: See other comment.

test/bases/renku_data_services/data_api/test_authn_api.py
leafty: Do you have a code suggestion? I don't think I can use fixtures as test cases so I don't know if the code would like easier to read.

sgaist, line 93: In this case I was just thinking about the token itself.
Something like:
```suggestion
@pytest.mark.parametrize("access_token",["", "invalid-token"])
@pytest.mark.asyncio
async def test_post_internal_token_reject_access_token(
    sanic_client: SanicASGITestClient, app_manager_instance: "DependencyManager", regular_user: UserInfo, access_token: str
) -> None:
    user = base_models.AuthenticatedAPIUser(
        id=regular_user.id,
        email=regular_user.email or "",
        access_token=access_token,
```

sgaist: Sorry, I have messed the conversation while trying to provide a suggestion.

leafty: This doesn't test what I was testing before: that the refresh endpoint accepts only valid refresh tokens and rejects valid access tokens. This change doesn't do that part.

components/renku_data_services/authn/api/api.spec.yaml
leafty: I can remove it here, but it will be present in the merged spec.

test/bases/renku_data_services/data_api/test_authn_api.py
sgaist: My bad, I misunderstood the purpose of the test.

sgaist (approved):

components/renku_data_services/authn/api/api.spec.yaml
```diff
@@ -0,0 +1,102 @@
...
+      tags:
+        - internal_authentication
+components:
+  schemas:
+    PostTokenRequest:
+      type: object
+      additionalProperties: true
+      properties:
+        grant_type:
+          $ref: "#/components/schemas/PostTokenGrantType"
```
olevski, line 40: question: We support just some grant types, not all right? So can we be explicit about what we support in the apispec?

components/renku_data_services/authn/api/api.spec.yaml
```diff
@@ -0,0 +1,102 @@
...
+        - grant_type
+        - refresh_token
+    PostTokenResponse:
+      type: object
+      additionalProperties: true
+      properties:
+        access_token:
+          type: string
+        token_type:
+          type: string
```
olevski, line 54: question: Can we put anything here? This is just always going to be "bearer" or something like that?

components/renku_data_services/authn/api/blueprints.py
```diff
@@ -0,0 +1,73 @@
...
+        """Token endpoint for internal authentication.
+
+        Supports refreshing internal authentication tokens.
+        """
+
+        @validate(form=apispec.PostTokenRequest)
+        async def _post_token(request: Request, body: apispec.PostTokenRequest) -> JSONResponse:
+            parsed_token = await self.internal_authenticator.verify_refresh_token(refresh_token=body.refresh_token)
+            if parsed_token is None:
+                raise errors.UnauthorizedError(message="You have to be authenticated to perform this operation.")
```
olevski, line 36: question: shouldn't we check the grant_type here? This is basically supporting the `refresh_token` grant type. So we can respond with 422 in all other cases.

components/renku_data_services/authn/renku.py
```diff
@@ -0,0 +1,208 @@
...
+        with suppress(errors.UnauthorizedError, jwt.InvalidTokenError):
+            token = header_value.removeprefix("Bearer ").removeprefix("bearer ")
+            parsed = self._validate(token)
+            exp = parsed.get("exp")
+            id = parsed.get("sub")
+            email = parsed.get("email")
+            if id is None or email is None:
+                raise errors.UnauthorizedError(
+                    message="Your credentials are invalid or expired, please log in again."
+                ) from None
```
olevski: question: I think the `errors.UnauthorizedError` can occur only when you raise it yourself. So why suppress it above? Also the only place where `jwt.InvalidTokenError` can occur is in `self._validate`. So why not just supress that line?

components/renku_data_services/authn/renku.py
```diff
@@ -0,0 +1,208 @@
...
+            token_type = parsed.get("typ")
+            if str(token_type).lower() != "refresh":
+                raise errors.UnauthorizedError() from None
+            return parsed
+
+        return None
```
olevski, line 117: Same issue as above. Why wrap so many lines and suppress the erorr that you raise right here in the code.

olevski (changes requested):
Looks good Flora. Just a few questions about the apispec and some of the validation logic.

components/renku_data_services/authn/api/api.spec.yaml
leafty: 👍 Maybe I should also update the API spec for `POST /api/data/oauth2/connections/:connection_id/token_endpoint` then.

components/renku_data_services/authn/renku.py
leafty: This uses the same structure as in `keycloak.py`. I can clean up the code here and make an issue to do that for `keycloak.py`.

components/renku_data_services/authn/api/api.spec.yaml
leafty: Actually, what do you mean? The only type accepted in the spec is `refresh_token`.

components/renku_data_services/authn/api/api.spec.yaml
leafty: The API spec is shared with `POST /api/data/oauth2/connections/:connection_id/token_endpoint` where it could be something else (but it should be `"Bearer"` if the service complies with OAuth 2.0).

components/renku_data_services/authn/api/blueprints.py
leafty: This is validated by `@validate(form=apispec.PostTokenRequest)`. No need to validate it again.

components/renku_data_services/authn/renku.py
leafty: I re-read the code and I believe the original code (similar to `keycloak.py`) is the most readable: we want to execute the whole block until `return AuthenticatedAPIUser(...)` if everything goes well and return the anonymous user if there is a token validation issue.

components/renku_data_services/authn/api/api.spec.yaml
olevski: Ok I was looking at the api spec in swagger and swagger does not show this when it renders the nice page. I think it never shows more involved stuff like this for www-form-urlencoded.

components/renku_data_services/authn/api/api.spec.yaml
olevski: Yeah but we know in this case we only support and will only ever return bearer. So you can make a different schema for this and make it an enum with a single value.

components/renku_data_services/authn/api/blueprints.py
olevski: My bad I missed the spec for the grant type.

olevski (approved):

components/renku_data_services/authn/api/api.spec.yaml
leafty: On the swagger page, when I do "try it", the only option is `refresh_token` from a dropdown (which I guess is used for enum types).

PR #1259, opened 2026-04-15
feat: verify scope when refreshing internal tokens

leafty:
Verify the scope claim when refreshing internal tokens:
* Check that the corresponding session or deposit job is running

PR Stack:
* (base) #1258
    * (this, tip) #1259

### DIFF
--- a/components/renku_data_services/authn/api/core.py
+++ b/components/renku_data_services/authn/api/core.py
@@ -57,7 +57,7 @@ class ScopeVerifier:
         if not user.is_authenticated or not user.id:
             raise errors.UnauthorizedError()
 
-        session = await self.k8s_client.get_session(name=server_name, safe_username=user.id)
+        session = await self.notebook_k8s_client.get_session(name=server_name, safe_username=user.id)
         if session is None or session.status.state.value.lower() == cr_amalthea_session.State.Hibernated.value.lower():
             raise errors.ForbiddenError(detail=f"Failed to verify session scope '{server_name}'.")
         return None
--- a/components/renku_data_services/authn/api/core.py
+++ b/components/renku_data_services/authn/api/core.py
@@ -26,7 +26,7 @@ class ScopeVerifier:
     """Verifies the scope claim for internal authentication tokens."""
 
     deposit_config: "DepositConfig"
-    k8s_client: "NotebookK8sClient"
+    notebook_k8s_client: "NotebookK8sClient"
     job_client: "DepositUploadJobClient"
 
     async def verify_scope(self, user: base_models.APIUser, scope: str) -> None:
--- a/components/renku_data_services/authn/api/core.py
+++ b/components/renku_data_services/authn/api/core.py
@@ -33,6 +33,7 @@ class ScopeVerifier:
         """Verify that the scope claim is valid.
 
         * For "session:{server_name}", check that the corresponding session is running
+        * For "deposit_job:{job_name}", check that the corresponding job is running
         """
         scopes = scope.split(" ")
         logger.info(f"Verifying scopes: {scopes}")
--- a/components/renku_data_services/authn/api/core.py
+++ b/components/renku_data_services/authn/api/core.py
@@ -59,6 +59,10 @@ class ScopeVerifier:
 
         session = await self.notebook_k8s_client.get_session(name=server_name, safe_username=user.id)
         if session is None or session.status.state.value.lower() == cr_amalthea_session.State.Hibernated.value.lower():
+            logger.error(
+                f"Failed to verify session scope '{server_name}', "
+                f"session status: {'not found' if session is None else session.status.state.value.lower()}."
+            )
             raise errors.ForbiddenError(detail=f"Failed to verify session scope '{server_name}'.")
         return None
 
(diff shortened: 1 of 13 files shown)


components/renku_data_services/notebooks/data_sources.py
```diff
@@ -259,6 +259,9 @@ async def _get_oauth2_configuration_part(
 
         token_config["expiry"] = (datetime.now(UTC) + timedelta(seconds=expires_in)).isoformat()
 
+        # TODO: remove, this is for debug (forcing immediate refresh)
+        token_config["expiry"] = (datetime.now(UTC) - timedelta(days=1)).isoformat()
+
```
leafty:
```suggestion
```

components/renku_data_services/authn/api/core.py
```diff
@@ -0,0 +1,82 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True, kw_only=True)
+class ScopeVerifier:
+    """Verifies the scope claim for internal authentication tokens."""
+
+    deposit_config: "DepositConfig"
+    k8s_client: "NotebookK8sClient"
```
sgaist: There's a lot of different k8s client in the code base, I would suggest to use `notebook_client` or `notebook_k8s_client`.

components/renku_data_services/authn/api/core.py
```diff
@@ -0,0 +1,82 @@
...
+                return None
+
+    async def _verify_session_scope(self, user: base_models.APIUser, server_name: str) -> None:
+        """Verify a scope item corresponding to a session."""
+        logger.info(f"Verifying session: {server_name}")
+        if not user.is_authenticated or not user.id:
+            raise errors.UnauthorizedError()
+
+        session = await self.k8s_client.get_session(name=server_name, safe_username=user.id)
+        if session is None or session.status.state.value.lower() == cr_amalthea_session.State.Hibernated.value.lower():
```
sgaist, line 61: While I understand that no session means failure, from an admin point of view, it looks strange that a hibernated session triggers the same error message.

components/renku_data_services/data_connectors/deposits/zenodo.py
```diff
@@ -31,7 +31,8 @@ class ZenodoAPIClient:
     """Client to deal with Zenodo datasets."""
 
     def __init__(self) -> None:
-        self.__zenodo_base_url = os.environ.get("ZENODO_URL", "https://zenodo.org").rstrip("/") + "/api"
+        # self.__zenodo_base_url = os.environ.get("ZENODO_URL", "https://zenodo.org").rstrip("/") + "/api"
+        self.__zenodo_base_url = os.environ.get("ZENODO_URL", "https://sandbox.zenodo.org").rstrip("/") + "/api"
```
sgaist, line 35: If it is for testing purposes, should we have a configurable way to switch between the two ?

components/renku_data_services/data_connectors/config.py
```diff
@@ -63,5 +63,6 @@ def from_env(cls) -> DepositConfig:
             node_selector=node_selector,
             namespace=os.environ["KUBERNETES_NAMESPACE"],
             cluster_id=DEFAULT_K8S_CLUSTER,
-            zenodo_url=os.environ.get("ZENODO_URL", "https://zenodo.org").rstrip("/"),
+            # zenodo_url=os.environ.get("ZENODO_URL", "https://zenodo.org").rstrip("/"),
+            zenodo_url=os.environ.get("ZENODO_URL", "https://sandbox.zenodo.org").rstrip("/"),
         )
```
sgaist, line 68: Same as above

components/renku_data_services/authn/api/core.py
leafty: ? I don't follow.
Tokens are refreshed by sidecar services or other infra supporting the session, so a refresh should NOT happen when a session is paused (not running).

components/renku_data_services/data_connectors/deposits/zenodo.py
leafty: Yes, I am keeping this for now so that I have working deployments that upload to the sandbox. This will be removed once the whole feature branch is ready.

components/renku_data_services/authn/api/core.py
leafty: The token refresh is a valid operation if the underlying resource (session or job) is running. In other cases this operation is forbidden.

sgaist: Ok, that makes things clearer.

I would add a comment stating that. You can be sure someone else (or me again) in the future will look at that and get to the same wrong conclusion.

leafty: The public method is documented already, with exactly these rules.

```python
"""Verify that the scope claim is valid.

        * For "session:{server_name}", check that the corresponding session is running
        * For "deposit_job:{job_name}", check that the corresponding job is running
        """
```

sgaist: Sorry, I just realized that I have been mixing two different issues together...

For the method documentation: I analyzed them independently hence I missed the `session is running` from the public method documentation. So it's all good.

For the error message, what I had in mind is: should we have two different error texts:
- session is None -> `"Failed to verify session scope: '{server_name} does not exist'."`
- session is hibernated -> `"Failed to verify session scope: '{server_name} is not running / is hibernated'."`

leafty: Not sure I want to send information like that when this should not happen and could be the result of a token steal.

components/renku_data_services/authn/api/core.py
leafty:
```suggestion
    notebook_k8s_client: "NotebookK8sClient"
```

components/renku_data_services/authn/api/core.py
sgaist: I see, then all good message wise.

One last question, should the event be logged for observability ?

leafty: I added log entries in 66ec65df71fca9eb7aacf44f9717bae59362b096.

sgaist (approved):

PR #1264, opened 2026-04-16
feat(builds): add the possibility to configure insecure registries

sgaist:
## Describe your changes

This PR implements the logic that allows to use insecure registries with the buildpacks based BuildStrategy.

This allows, for example, to use a local deployment of registries such as Harbor using a self-signed certificate or no certificate at all.

The buildpacks lifecycle supports this use case through the CNB_INSECURE_REGISTRIES environment variable with a comma separated list of insecure registries.

### DIFF
--- a/components/renku_data_services/session/config.py
+++ b/components/renku_data_services/session/config.py
@@ -54,6 +54,10 @@ class BuildsConfig:
         build_run_image = os.environ.get("BUILD_RUN_IMAGE")
         build_strategy_name = os.environ.get("BUILD_STRATEGY_NAME")
         build_insecure_registries = os.environ.get("BUILD_INSECURE_REGISTRIES", "")
+        if build_insecure_registries:
+            logger.warn(
+                f"Trusting insecure registries, this is not recommended in production: {build_insecure_registries}"
+            )
         push_secret_name = os.environ.get("BUILD_PUSH_SECRET_NAME")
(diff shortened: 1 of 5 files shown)


leafty (commented):
I don't like the idea of adding a configuration option which removes security. Is it too hard to add certificates to the renku-data-services container?

sgaist:
@leafty it's not at the `renku-data-service` level that it's played but down to Tekton.

Their [documentation](https://tekton.dev/docs/pipelines/additional-configs/#configuring-self-signed-cert-for-private-registry) has three lines on the subject. I tried, but currently did not succeed to make it work.

There's https://github.com/tektoncd/website/issues/587 already opened on the subject.

components/renku_data_services/session/config.py
```diff
@@ -52,6 +53,7 @@ def from_env(cls) -> "BuildsConfig":
         build_builder_image = os.environ.get("BUILD_BUILDER_IMAGE")
         build_run_image = os.environ.get("BUILD_RUN_IMAGE")
         build_strategy_name = os.environ.get("BUILD_STRATEGY_NAME")
+        build_insecure_registries = os.environ.get("BUILD_INSECURE_REGISTRIES", "")
```
leafty, line 56: We need to print a warning-level log line if this is not empty.

```
WARNING Trusting insecure registries, this is not recommended in production: {build_insecure_registries}
```

sgaist: Agreed and done

leafty (approved):

PR #1250, opened 2026-04-01
feat: build from private repositories

sgaist:
## Describe your changes

This PR implements building images from private repositories.

### Breakdown

- Launcher creation for build environment will now support private repositories
- Logic for passing credentials to Shipwright on a per build basis
- Public repositories will not use credentials
- Updated tests to also be able to validate build features
- Refactored caching logic client side
- Session starting with a custom built image will now:
  - ensure that the repo of the project matches the one used to build the image
  - ensure that the user can pull the repository

---

<code>
/deploy #notest renku=feat/private-repository-build extra-values=dataService.imageBuilders.enabled=true,dataService.imageBuilders.strategyName=renku-buildpacks-v3,dataService.imageBuilders.outputImagePrefix=harbor.dev.renku.ch/renku-build/,dataService.imageBuilders.outputPrivateImagePrefix=harbor.dev.renku.ch/renku-build/,dataService.imageBuilders.nodeSelector.renku.io/node-purpose=user,dataService.imageBuilders.tolerations[0].effect=NoSchedule,dataService.imageBuilders.tolerations[0].key=renku.io/dedicated,dataService.imageBuilders.tolerations[0].operator=Equal,dataService.imageBuilders.tolerations[0].value=user
</code>

PR #1265, opened 2026-04-20
feat(notebook): add secret handling for private registry image pull

sgaist:
## Describe your changes

This is a followup of #1250 that implements the complementary part to support adding the private's registry corresponding image pull secret to a user's session.

This secret is created by the admin deploying the platform and is not user accessible.

### DIFF
--- a/components/renku_data_services/notebooks/core_sessions.py
+++ b/components/renku_data_services/notebooks/core_sessions.py
@@ -664,15 +664,12 @@ async def get_image_pull_secret(
         and builds_config.build_output_private_image_prefix is not None
         and image.startswith(builds_config.build_output_private_image_prefix)
     ):
-        v2_secret = ExtraSecret(
+        return ExtraSecret(
             V1Secret(
                 metadata=V1ObjectMeta(name=builds_config.pull_private_image_secret_name),
             ),
             adopt=False,
         )
-
-    if v2_secret:
-        return v2_secret
 
     if (
         nb_config.enable_internal_gitlab
--- a/components/renku_data_services/notebooks/core_sessions.py
+++ b/components/renku_data_services/notebooks/core_sessions.py
@@ -659,7 +659,11 @@ async def get_image_pull_secret(
     if v2_secret:
         return v2_secret
 
-    if builds_config.enabled and image.startswith(builds_config.build_output_private_image_prefix):
+    if (
+        builds_config.enabled
+        and builds_config.build_output_private_image_prefix is not None
+        and image.startswith(builds_config.build_output_private_image_prefix)
+    ):
         v2_secret = ExtraSecret(
             V1Secret(
                 metadata=V1ObjectMeta(name=builds_config.pull_private_image_secret_name),
--- a/components/renku_data_services/notebooks/core_sessions.py
+++ b/components/renku_data_services/notebooks/core_sessions.py
@@ -105,6 +105,7 @@ from renku_data_services.project.db import ProjectRepository, ProjectSessionSecr
 from renku_data_services.project.models import Project, SessionSecret
 from renku_data_services.repositories.db import GitRepositoriesRepository
 from renku_data_services.repositories.models import Metadata as RepoMetadata
+from renku_data_services.session.config import BuildsConfig
 from renku_data_services.session.db import SessionRepository
 from renku_data_services.session.models import SessionLauncher
 from renku_data_services.users.db import UserRepo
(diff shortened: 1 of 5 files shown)


components/renku_data_services/notebooks/core_sessions.py
```diff
@@ -657,6 +659,21 @@ async def get_image_pull_secret(
...
+    ):
+        v2_secret = ExtraSecret(
+            V1Secret(
+                metadata=V1ObjectMeta(name=builds_config.pull_private_image_secret_name),
+            ),
+            adopt=False,
+        )
+
+    if v2_secret:
+        return v2_secret
```
leafty: In the code above, `__get_connected_services_image_pull_secret` returns `ExtraSecret | None`. In the new `if` branch, you always return.

```suggestion
    if (
        builds_config.enabled
        and builds_config.build_output_private_image_prefix is not None
        and image.startswith(builds_config.build_output_private_image_prefix)
    ):
        return ExtraSecret(
            V1Secret(
                metadata=V1ObjectMeta(name=builds_config.pull_private_image_secret_name),
            ),
            adopt=False,
        )
```

components/renku_data_services/notebooks/core_sessions.py
```diff
@@ -969,7 +988,9 @@ async def start_session(
         metadata=Metadata(name=server_name, annotations=annotations),
         spec=AmaltheaSessionSpec(
             location=session_location,
-            imagePullSecrets=[ImagePullSecret(name=image_secret.name, adopt=True)] if image_secret else [],
+            imagePullSecrets=[ImagePullSecret(name=image_secret.name, adopt=image_secret.adopt)]
```
leafty, line 989: Good thing to use `image_secret.adopt`.

Maybe we should have `adopt=True` explicitly for the other cases around line L616:

```python
return ExtraSecret(
        secret = V1Secret(
            data={".dockerconfigjson": registry_secret},
            metadata=V1ObjectMeta(name=secret_name),
            type="kubernetes.io/dockerconfigjson",
        ),
       adopt=True,
    )
```

components/renku_data_services/notebooks/core_sessions.py
```diff
@@ -657,6 +659,21 @@ async def get_image_pull_secret(
     if v2_secret:
         return v2_secret
 
+    if (
+        builds_config.enabled
+        and builds_config.build_output_private_image_prefix is not None
+        and image.startswith(builds_config.build_output_private_image_prefix)
+    ):
```
leafty, line 667: Comment: this may be an issue if `build_output_private_image_prefix` is mis-configured to be the same as the public one (if `build_output_image_prefix.startswith(build_output_private_image_prefix)` then we give credentials for the public registry).

sgaist: We can add a validation upon configuration creation to ensure both are different and fail early if it's not the case.

components/renku_data_services/notebooks/core_sessions.py
sgaist: Thanks for pointing it out. I felt I was doing something not quite right.

components/renku_data_services/notebooks/core_sessions.py
sgaist: Good for me

It's the default value in the ExtraSecret class but I think being explicit here makes things better.

components/renku_data_services/notebooks/core_sessions.py
leafty: 👍  I don't think we need to add this urgently, but maybe having it in `components/renku_data_services/session/config.py` would be nice.

leafty (commented):
This looks good to me. At least I see all the needed parts now, though I would like to see if we can get a real CI deployment 😄

components/renku_data_services/notebooks/core_sessions.py
sgaist: I think it's an easy addition that can be done in this PR.

sgaist:
> This looks good to me. At least I see all the needed parts now, though I would like to see if we can get a real CI deployment 😄

"Works on my k8s on my machine" ^^

More seriously, https://github.com/SwissDataScienceCenter/renku/pull/4428 adds the bits and pieces for the deployment.

I just added the image pull secret configuration part.

sgaist:
@leafty for testing in a deployment: can you mix and match in the request ?

Because you would need https://github.com/SwissDataScienceCenter/renku-ui/pull/4133 for the UI to allow you to do the build and https://github.com/SwissDataScienceCenter/renku/pull/4428 to have things configured.

leafty (approved):

PR #1268, opened 2026-04-21
minor: switch from usage_available_percentage to usage_limit_total

ciyer:


components/renku_data_services/crc/api.spec.yaml
```diff
@@ -1707,13 +1707,12 @@ components:
       description: Amount of a resource available (in hours)
       example: 3.141
       minimum: 0
-    UsageAvailablePercentage:
+    UsageLimitTotal:
       type: number
       format: float
```
leafty: consistency
```suggestion
      format: double
```

leafty (commented):
OK otherwise

mohammad-alisafaee (approved):

PR #1269, opened 2026-04-23
fix: strip prefix for rstudio built from code

olevski:
This is the final thing needed to get r/rstudio images built from code running.

### DIFF
--- a/components/renku_data_services/session/db.py
+++ b/components/renku_data_services/session/db.py
@@ -217,7 +217,7 @@ class SessionRepository(SessionEnvironmentRepositoryProtocol):
             environment_image_source=models.EnvironmentImageSource.build,
             build_parameters_id=build_parameters_orm.id,
             build_parameters=build_parameters_orm,
-            strip_path_prefix=False,
+            strip_path_prefix=build_parameters_orm.frontend_variant == models.FrontendVariant.rstudio.value,
         )
         session.add(environment_orm)
         return environment_orm
@@ -779,7 +779,9 @@ class SessionRepository(SessionEnvironmentRepositoryProtocol):
                 launcher.environment.environment_image_source = models.EnvironmentImageSource.build
                 launcher.environment.build_parameters_id = build_parameters_orm.id
                 launcher.environment.build_parameters = build_parameters_orm
-                launcher.environment.strip_path_prefix = False
+                launcher.environment.strip_path_prefix = (
+                    build_parameters_orm.frontend_variant == models.FrontendVariant.rstudio.value
+                )
 
                 await session.flush()
             case _:
--- a/components/renku_data_services/session/db.py
+++ b/components/renku_data_services/session/db.py
@@ -479,7 +479,8 @@ class SessionRepository(SessionEnvironmentRepositoryProtocol):
                     environment_image_source=models.EnvironmentImageSource.build,
                     build_parameters_id=build_parameters_orm.id,
                     build_parameters=build_parameters_orm,
-                    strip_path_prefix=False,  # TODO: Should this maybe be adjustable?
+                    strip_path_prefix=launcher.environment.frontend_variant
+                    == models.FrontendVariant.rstudio.value,  # TODO: Should this maybe be adjustable?
                 )
                 session.add(environment_orm)
 
--- a/components/renku_data_services/session/db.py
+++ b/components/renku_data_services/session/db.py
@@ -209,16 +209,16 @@ class SessionRepository(SessionEnvironmentRepositoryProtocol):
             name=launcher.name,
             created_by_id=user.id,
             description=f"Generated environment for {launcher.name}",
-            container_image="image:unknown-at-the-moment",  # TODO: This should come from the build
-            default_url=build_env.default_url,  # TODO: This should come from the build
-            port=build_env.port,  # TODO: This should come from the build
-            working_directory=build_env.working_directory,  # TODO: This should come from the build
-            mount_directory=build_env.mount_directory,  # TODO: This should come from the build
-            uid=build_env.uid,  # TODO: This should come from the build
-            gid=build_env.gid,  # TODO: This should come from the build
+            container_image=build_env.container_image,
+            default_url=build_env.default_url,
+            port=build_env.port,
+            working_directory=build_env.working_directory,
+            mount_directory=build_env.mount_directory,
+            uid=build_env.uid,
+            gid=build_env.gid,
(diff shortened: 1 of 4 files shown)


leafty (changes requested):
The same change is missing on L220 and L782.

https://github.com/SwissDataScienceCenter/renku-data-services/blob/08225c3ceda5c0c1041f7ad63d8dc32c8cf18080/components/renku_data_services/session/db.py#L220

https://github.com/SwissDataScienceCenter/renku-data-services/blob/08225c3ceda5c0c1041f7ad63d8dc32c8cf18080/components/renku_data_services/session/db.py#L782

olevski:
> The same change is missing on L220 and L782.

Good catch @leafty ! Thank you. I applied the changes there.

SalimKayal (dismissed):

leafty (changes requested):
So, we also need to do the same type of update inside `_refresh_build()`, at line L1078: the whole `environment` may need to be modified:

```python
environment.container_image = build.result_image
environment.default_url = ??? # value depends on the frontend parameter
environment.port = 8888 # value could depend on the frontend parameter but is always 8888 at the moment
environment.working_directory = ??? # value depends on buildpack?
environment.mount_directory = environment.working_directory
environment.uid = 1000 # value depends on buildpack? always 1000 at the moment
environment.gid = 1000 # value depends on buildpack? always 1000 at the moment
environment.command = None # value  could depend on buildpack and frontend, always None at the moment
environment.args = None # value  could depend on buildpack and frontend, always None at the moment
environment.strip_path_prefix = ??? # value depends on the frontend parameter
```

olevski:
@leafty take a look. I defined one set  of configs for the builder envs and then I pulled those in from everywhere.

components/renku_data_services/session/db.py
```diff
@@ -199,25 +199,31 @@ def __insert_build_parameters_environment(
...
-            args=None,  # TODO: This should come from the build
+            default_url=build_env.default_url,  # TODO: This should come from the build
+            port=build_env.port,  # TODO: This should come from the build
+            working_directory=build_env.working_directory,  # TODO: This should come from the build
+            mount_directory=build_env.mount_directory,  # TODO: This should come from the build
+            uid=build_env.uid,  # TODO: This should come from the build
+            gid=build_env.gid,  # TODO: This should come from the build
+            environment_kind=build_env.environment_kind,
+            command=build_env.command,  # TODO: This should come from the build
+            args=build_env.args,  # TODO: This should come from the build
```
leafty: You can remove the TODO comments now 😄 .

components/renku_data_services/session/db.py
```diff
@@ -459,27 +465,31 @@ async def insert_launcher(
...
+                    environment_kind=env.environment_kind,
+                    command=env.command,
+                    args=env.args,
                     creation_date=datetime.now(UTC).replace(microsecond=0),
-                    environment_image_source=models.EnvironmentImageSource.build,
+                    environment_image_source=env.environment_image_source,
                     build_parameters_id=build_parameters_orm.id,
                     build_parameters=build_parameters_orm,
-                    strip_path_prefix=False,  # TODO: Should this maybe be adjustable?
+                    strip_path_prefix=env.strip_path_prefix,  # TODO: Should this maybe be adjustable?
```
leafty:
```suggestion
                    strip_path_prefix=env.strip_path_prefix,
```

components/renku_data_services/session/db.py
```diff
@@ -763,22 +773,32 @@ async def __update_launcher_environment(
...
 
+                frontend_variant = models.FrontendVariant(new_custom_built_environment.frontend_variant)
+                build_env = models.BUILD_ENVIRONMENT_CONFIGS.get(frontend_variant)
+                if not build_env:
+                    raise errors.ValidationError(
+                        message=f"Frontend variant {frontend_variant} is not valid or cannot be found."
+                    )
                 launcher.environment.container_image = (
-                    "image:unknown-at-the-moment"  # TODO: This should come from the build
+                    build_env.container_image  # TODO: This should come from the build
```
leafty: Same here and below.

components/renku_data_services/session/db.py
```diff
@@ -1040,7 +1060,11 @@ async def get_build_logs(
         )
 
     async def _refresh_build(self, build: schemas.BuildORM, session: AsyncSession, user_id: str) -> None:
-        """Refresh the status of a build by querying Shipwright."""
+        """Refresh the status of a build by querying Shipwright.
+
+        The update of things other than the status are applied to the buildrun first
+        and then to the DB after the buildrun succeeds. So this updates more than just the status.
```
leafty: ? what does this mean?


```suggestion
        Once the build run has completed, the corresponding session launcher's environment is updated to allow launching sessions with the newly built image.
```

components/renku_data_services/session/db.py
```diff
@@ -1079,6 +1103,19 @@ async def _refresh_build(self, build: schemas.BuildORM, session: AsyncSession, u
             if environment.args is not None:
                 environment.args = None
 
+            environment.container_image = build.result_image
+            build_env = models.BUILD_ENVIRONMENT_CONFIGS.get(frontend_var) if frontend_var else None
+            if build_env:
```
leafty, line 1093: Consolidate this (and below) with the code just above, it's doing the same thing twice.

leafty (commented):
Comment to get this clean:

components/renku_data_services/session/db.py
```diff
@@ -1068,16 +1086,23 @@ async def _refresh_build(self, build: schemas.BuildORM, session: AsyncSession, u
...
             environment = build.environment
             environment.container_image = build.result_image
-            # An older version was hardcoding the values but we can and should
-            # rely on the defaults for args and command
-            if environment.command is not None:
-                environment.command = None
-            if environment.args is not None:
-                environment.args = None
+            build_env = models.BUILD_ENVIRONMENT_CONFIGS.get(frontend_var) if frontend_var else None
+            if build_env:
```
leafty, line 1093: Q: should we log an error if this is none?

leafty (dismissed):
Looks really good now, just one comment below.

components/renku_data_services/session/db.py
olevski: Yeah that is a good point. Let me add that.

olevski: Done.

leafty (approved):

PR #1273, opened 2026-04-27
feat: use internal tokens for git-clone

leafty:
wip

/deploy renku=release-2.17.0


components/renku_data_services/notebooks/api/amalthea_patches/init_containers.py
```diff
@@ -68,7 +74,8 @@ async def git_clone_container_v2(
         },
         {
             "name": f"{prefix}USER__RENKU_TOKEN",
-            "value": str(user.access_token),
+            # "value": str(user.access_token),
```
leafty:
```suggestion
```

PR #1274, opened 2026-04-28
fix(sessions): use image pull secret from private build registry correctly

sgaist:
## Describe your changes

When retrieving the image pull secret to be used by a session, if
it's an existing secret, do not add it to the extra resources to
be created.

This is the case for the registry used to store the images built
out of private repositories.

### DIFF
--- a/components/renku_data_services/notebooks/core_sessions.py
+++ b/components/renku_data_services/notebooks/core_sessions.py
@@ -928,8 +928,8 @@ async def start_session(
         image_check_repo=image_check_repo,
         builds_config=builds_config,
     )
-    # Adopt being false means the returned secret already exists and should be used as is
-    if image_secret and image_secret.adopt:
+
+    if image_secret and image_secret.secret.data is not None:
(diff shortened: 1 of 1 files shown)


leafty (dismissed):
:+1:
Two remarks:
* Maybe the image secret should be handled as a dedicated field of `SessionExtraResources`.
* Also, afaik `adopt` is about resource finalizers, not about if we create the resource or not. Or maybe we can update documentation about `adopt` that specifies what it means

sgaist:
> 👍 Two remarks:
> 
>     * Maybe the image secret should be handled as a dedicated field of `SessionExtraResources`.

To explore

>     * Also, afaik `adopt` is about resource finalizers, not about if we create the resource or not. Or maybe we can update documentation about `adopt` that specifies what it means

That's indeed the role of adopt, hence my comment in the code. I used that property like that in this specific case because the method returning the image pull secret would return "secrets to be created and adopted" in all the cases that are not the private registry. 

But I just realized that there's a better way to do it: check whether the returned secret contains data. I will update the PR after testing.

leafty (approved):

PR #1275, opened 2026-04-28
Changes to support major solr upgrade to 10

eikek:
Solr10 requires some adoption when creating cores. By default, version 9 is assumed (we don't have to upgrade right away). Env `SOLR_MAJOR_VERSION` can be set to `10` to enable the new code.


leafty (dismissed):
LGTM

leafty (approved):

PR #1288, opened 2026-05-04
fix: set priority class name when resuming a session

### LINKED ISSUE #1287: Priority class is not re-set when resuming a session
Reproduce:
1. Start a session from the default resource pool
2. Pause the session
3. Modify the resource class so that it should use a priority class in k8s (for example, pick a resource class with a GPU)
4. Resume the session

=> Observe that the priority class is not set. If you start the same session directly with the GPU attached, it will have the correct priority class attached.

leafty:
Closes #1287.

Also, ensure the annotations are updated when resuming.

/deploy renku=release-2.17.0

### DIFF
--- a/components/renku_data_services/notebooks/core_sessions.py
+++ b/components/renku_data_services/notebooks/core_sessions.py
@@ -1169,8 +1169,7 @@ async def patch_session(
         # Affinities
         patch.spec.affinity = node_affinity_patch_from_resource_class(rc, nb_config.sessions.affinity_model)
         # Priority class (if a quota is being used)
-        if rc.quota is None:
-            patch.spec.priorityClassName = RESET
+        patch.spec.priorityClassName = rc.quota if rc.quota else RESET
         # Service account name
         if rp.cluster is not None:
             patch.spec.service_account_name = (
--- a/components/renku_data_services/notebooks/core_sessions.py
+++ b/components/renku_data_services/notebooks/core_sessions.py
@@ -83,6 +83,8 @@ from renku_data_services.notebooks.crs import (
     State,
     Storage,
     Template,
+    TemplateMetadataPatch,
+    TemplatePatch,
     TlsSecret,
 )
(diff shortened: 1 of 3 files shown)


components/renku_data_services/notebooks/core_sessions.py
```diff
@@ -1158,19 +1160,27 @@ async def patch_session(
...
         # Tolerations
-        tolerations = tolerations_from_resource_class(rc, nb_config.sessions.tolerations_model)
-        patch.spec.tolerations = tolerations
+        patch.spec.tolerations = tolerations_from_resource_class(rc, nb_config.sessions.tolerations_model)
         # Affinities
         patch.spec.affinity = node_affinity_patch_from_resource_class(rc, nb_config.sessions.affinity_model)
         # Priority class (if a quota is being used)
-        if rc.quota is None:
-            patch.spec.priorityClassName = RESET
+        patch.spec.priorityClassName = rc.quota if rc.quota else RESET
```
sgaist, line 1183: Design question: shouldn't that rather be `quota_name` ? `ResourcePool` has an optional `quota` field that is a `Quota` object but ResourceClass has an optional `quota` field that is a string.

sgaist (approved):

components/renku_data_services/notebooks/core_sessions.py
leafty: The field is typed `str | None`.

sgaist: That is clear but not my question.
Taken for the coherence angle, having two very close objects (ResourcePool and ResourceClass) using the same field name to represent two different relation (one has a, one relates to), almost three in this case (priorityClassName = quota name) makes things a bit hard to reason about.

leafty: 🤷‍♀️ I didn't design that code. Maybe open an issue.

PR #1290, opened 2026-05-04
fix: commit db session in the authz_change decorator

### LINKED ISSUE #1289: Authz change breaks asyncio?!?
This breaks everything it seems!

```
RuntimeError: Task <Task pending name='Task-220' coro=<connection_task() running at connection_task:15>> got Future <Task pending name='Task-1784' coro=<UnaryUnaryCall._invoke() running at /app/env/lib/python3.13/site-packages/grpc/aio/_call.py:577>> attached to a different loop

During handling of the above exception, another exception occurred:

  File "handle_request", line 99, in handle_request
    Extend = TypeVar("Extend", type)  # type: ignore
                       ^^^^^^^^^^^^^^
  File "/app/env/lib/python3.13/site-packages/renku_data_services/base_api/auth.py", line 36, in decorated_function
    response = await f(request, user, *args, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/env/lib/python3.13/site-packages/renku_data_services/base_api/auth.py", line 155, in decorated_function
    response = await f(*args, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/env/lib/python3.13/site-packages/renku_data_services/users/blueprints.py", line 62, in _get_self
    user_info = await self.repo.get_or_create_user(requested_by=user, id=user.id)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/env/lib/python3.13/site-packages/renku_data_services/users/db.py", line 144, in get_or_create_user
    return await self._add_api_user(requested_by)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/env/lib/python3.13/site-packages/renku_data_services/users/db.py", line 113, in _add_api_user
    result = await self._users_sync.update_or_insert_user(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<6 lines>...
    )
    ^
  File "/app/env/lib/python3.13/site-packages/renku_data_services/utils/core.py", line 83, in transaction_wrapper
    return await f(self, *args, **kwargs)
  File "/app/env/lib/python3.13/site-packages/renku_data_services/authz/authz.py", line 811, in decorated_function
    await asyncio.shield(db_repo.authz.client.WriteRelationships(authz_change.undo))
  File "/usr/local/lib/python3.13/asyncio/tasks.py", line 737, in _wrap_awaitable
    return await awaitable
           ^^^^^^^^^^^^^^^
  File "/app/env/lib/python3.13/site-packages/grpc/aio/_call.py", line 308, in __await__
    response = yield from self._call_response
RuntimeError: Task <Task pending name='Task-1787' coro=<ensure_future.<locals>._wrap_awaitable() running at /usr/local/lib/python3.13/asyncio/tasks.py:737> cb=[shield.<locals>._inner_done_callback() at /usr/local/lib/python3.13/asyncio/tasks.py:958]> got Future <Task pending name='Task-1786' coro=<UnaryUnaryCall._invoke() running at /app/env/lib/python3.13/site-packages/grpc/aio/_call.py:577>> attached to a different loop
2026-05-04T13:49:29.759222 [ERROR] 166/MainThread sanic.error (error.py:201) [40ff3a53-1f25-65d1-cc37-282b882f8d5d][-] - Exception occurred while handling uri: 'https://renku-ci-ds-1271.dev.renku.ch/api/data/user'
  File "/app/env/lib/python3.13/site-packages/renku_data_services/authz/authz.py", line 801, in decorated_function
    await db_repo.authz.client.WriteRelationships(authz_change.apply)
  File "/app/env/lib/python3.13/site-packages/grpc/aio/_call.py", line 308, in __await__
    response = yield from self._call_response
RuntimeError: Task <Task pending name='Task-815' coro=<connection_task() running at connection_task:15>> got Future <Task pending name='Task-1867' coro=<UnaryUnaryCall._invoke() running at /app/env/lib/python3.13/site-packages/grpc/aio/_call.py:577>> attached to a different loop
```

_Originally posted by @leafty in https://github.com/SwissDataScienceCenter/renku-data-services/pull/1266#discussion_r3182083016_

leafty:
Closes #1289.

/deploy renku=release-2.17.0

May need testing wrt resource pools.

### DIFF
--- a/bases/renku_data_services/data_api/main.py
+++ b/bases/renku_data_services/data_api/main.py
@@ -163,9 +163,9 @@ def create_app() -> Sanic:
     async def do_migrations(_: Sanic) -> None:
         logger.info("running migrations")
         run_migrations_for_app("common")
-        await dependency_manager.rp_repo.initialize(
-            dependency_manager.config.db.conn_url(async_client=False), dependency_manager.default_resource_pool
-        )
+        # await dependency_manager.rp_repo.initialize(
+        #     dependency_manager.config.db.conn_url(async_client=False), dependency_manager.default_resource_pool
+        # )
 
     @app.main_process_start
     async def do_solr_migrations(app: Sanic) -> None:
@@ -188,6 +188,11 @@ def create_app() -> Sanic:
         if getattr(app.ctx, "solr_reindex", False):
             logger.info("Creating solr reindex task, as required by migrations.")
             app.add_task(solr_reindex(dependency_manager.search_reprovisioning))
+        app.add_task(
+            dependency_manager.rp_repo.initialize(
+                dependency_manager.config.db.conn_url(async_client=False), dependency_manager.default_resource_pool
+            )
+        )
 
     return app
 
--- a/bases/renku_data_services/data_api/main.py
+++ b/bases/renku_data_services/data_api/main.py
@@ -163,9 +163,6 @@ def create_app() -> Sanic:
     async def do_migrations(_: Sanic) -> None:
         logger.info("running migrations")
         run_migrations_for_app("common")
-        # await dependency_manager.rp_repo.initialize(
-        #     dependency_manager.config.db.conn_url(async_client=False), dependency_manager.default_resource_pool
-        # )
 
     @app.main_process_start
     async def do_solr_migrations(app: Sanic) -> None:
--- a/components/renku_data_services/authz/authz.py
+++ b/components/renku_data_services/authz/authz.py
@@ -384,6 +384,7 @@ class Authz:
         # Drop the current client if it is attached to a different event loop
         if self._loop != running_loop:
             logger.info(f"Event loop mis-match, old = {self._loop}, new = {running_loop}")
+            self._loop = running_loop
             self._client = None
         if not self._client:
             self._client = self.authz_config.authz_async_client()
--- a/components/renku_data_services/authz/authz.py
+++ b/components/renku_data_services/authz/authz.py
@@ -799,6 +799,7 @@ class Authz:
                     result = await f(db_repo, *args, **kwargs)
                     authz_change = await _get_authz_change(db_repo, op, resource, result, *args, **kwargs)
                     await db_repo.authz.client.WriteRelationships(authz_change.apply)
+                    await session.commit()
                     return result
                 except Exception as err:
                     db_rollback_err = None
--- a/components/renku_data_services/authz/authz.py
+++ b/components/renku_data_services/authz/authz.py
@@ -375,10 +375,15 @@ class Authz:
     authz_config: AuthzConfig
     _platform: ClassVar[ObjectReference] = field(default=_AuthzConverter.platform())
     _client: AsyncClient | None = field(default=None, init=False)
+    _loop: asyncio.AbstractEventLoop | None = field(default=None, init=False)
 
     @property
     def client(self) -> AsyncClient:
         """The authzed DB asynchronous client."""
+        running_loop = asyncio.get_running_loop()
+        # Drop the current client if it is attached to a different event loop
+        if self._loop != running_loop:
+            self._client = None
         if not self._client:
             self._client = self.authz_config.authz_async_client()
         return self._client
--- a/components/renku_data_services/authz/authz.py
+++ b/components/renku_data_services/authz/authz.py
@@ -383,6 +383,7 @@ class Authz:
         running_loop = asyncio.get_running_loop()
         # Drop the current client if it is attached to a different event loop
         if self._loop != running_loop:
+            logger.info(f"Event loop mis-match, old = {self._loop}, new = {running_loop}")
             self._client = None
         if not self._client:
             self._client = self.authz_config.authz_async_client()
--- a/components/renku_data_services/crc/db.py
+++ b/components/renku_data_services/crc/db.py
@@ -938,14 +938,16 @@ class UserRepository(_Base):
             else:
                 removed_rps = [rp for rp in user.resource_pools if rp.id not in new_rp_ids]
                 user.resource_pools = list(rps_to_add)
+            # TODO: the authz changes below are done in separate db transactions,
+            # TODO: use only one (collect everything into a single authz change)
             for rp in rps_to_add:
-                await self._grant_resource_pool_members(api_user, rp.id, [keycloak_id], session=session)
+                await self._grant_resource_pool_members(api_user, rp.id, [keycloak_id])
                 if rp.default or rp.public:
-                    await self._unprohibit_resource_pool_users(api_user, rp.id, [keycloak_id], session=session)
+                    await self._unprohibit_resource_pool_users(api_user, rp.id, [keycloak_id])
             for rp in removed_rps:
-                await self._revoke_resource_pool_member(api_user, rp.id, [keycloak_id], session=session)
+                await self._revoke_resource_pool_member(api_user, rp.id, [keycloak_id])
                 if rp.default or rp.public:
-                    await self._prohibit_resource_pool_user(api_user, rp.id, [keycloak_id], session=session)
+                    await self._prohibit_resource_pool_user(api_user, rp.id, [keycloak_id])
             output: list[models.ResourcePool] = []
             for rp in rps_to_add:
                 quota = await self.quotas_repo.get_quota(rp.quota, rp.get_cluster_id()) if rp.quota else None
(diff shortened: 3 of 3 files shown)


leafty:
cc @SalimKayal : is there anything that would break or need adjusting?

olevski:
Yeah I completely missed this in the original PR. We need it back.  We want to try to commit because if there are problems in the DB side we want to roll back. And if we do not commit the transaction in the decorator we cannot handle the rollback (in all cases) properly.

leafty:
Mhm this deployment is broken right now with the asyncio error... 🤔

leafty:
So I deleted the pod and the newly created one is running correctly 😵‍💫

leafty:
My guess is that all the new uses of the `@Authz.authz_change` in `components/renku_data_services/crc/db.py` do NOT use the `with_db_transaction` transaction decorator above. This may explain the asyncio coroutine escape: the `_create_default_rp()` may be ran by the data-service pod on startup (first run when deployed on an empty database). The corresponding transaction escapes and it all cascades for further authz calls (notable `GET /api/data/user` to get self-user info).

SalimKayal:
Wouldn't the commit mess up with things that happen after when the session is set through a context manager (i.e. in crc/db.py)?

IMHO instead of commit it should be `session.flush()` if it needs to happen there.

Sorry for not seeing the issue earlier.

SalimKayal:
Per https://docs.sqlalchemy.org/en/21/orm/session_basics.html#committing

> Subsequent to that, assuming the Session is bound to an Engine, Session.commit() will then COMMIT the actual database transaction that is in place, if one was started. After the commit, the Connection object associated with that transaction is closed, causing its underlying DBAPI connection to be released back to the connection pool associated with the Engine to which the Session is bound.

leafty:
> Wouldn't the commit mess up with things that happen after when the session is set through a context manager (i.e. in crc/db.py)?
> 
> IMHO instead of commit it should be `session.flush()` if it needs to happen there.
> 
> Sorry for not seeing the issue earlier.

I think the idea of the authz change is that it should be a single transaction which I think is what happens for all existing uses of the decorator.

SalimKayal:
Ok I can refactor that way, though I don't like when the responsibility for releasing a resource is not taken by the function/class allocating it.

SalimKayal:
Also perhaps we should add a test where it failed

leafty:
> Also perhaps we should add a test where it failed

I don't know how to reliably test race conditions. Like I said, on a deployment, killing the pod will fix the issue because the `_create_default_rp()` method will not run (as it already exists), so the coroutine will not escape. I don't know how to write a test for that.

leafty:
> Ok I can refactor that way, though I don't like when the responsibility for releasing a resource is not taken by the function/class allocating it.

It is actually. We always use `@with_db_transaction` when there is an authz change. Maybe the authz decorator should handle that.

SalimKayal:
ok good to know. though just decorating will probably trigger https://github.com/SwissDataScienceCenter/renku-data-services/blob/0cdf55c3a27db9b20f274a1c6af4ef6b0025a06f/components/renku_data_services/utils/core.py#L76

I will refactor so that it works

leafty:
> ok good to know. though just decorating will probably trigger
> 
> https://github.com/SwissDataScienceCenter/renku-data-services/blob/0cdf55c3a27db9b20f274a1c6af4ef6b0025a06f/components/renku_data_services/utils/core.py#L76
> 
> I will refactor so that it works



> ok good to know. though just decorating will probably trigger
> 
> https://github.com/SwissDataScienceCenter/renku-data-services/blob/0cdf55c3a27db9b20f274a1c6af4ef6b0025a06f/components/renku_data_services/utils/core.py#L76
> 
> I will refactor so that it works

Why would it error? If you manually pass a `session` it should still work.

SalimKayal:
no it would not error, I misread. though if you manually pass a session we still have the double commit problem.

leafty (commented):
I still have the coroutine issue with this patch 😢

SalimKayal:
I'm afraid this might break with the early session.commit

https://github.com/SwissDataScienceCenter/renku-data-services/blob/0cdf55c3a27db9b20f274a1c6af4ef6b0025a06f/components/renku_data_services/crc/db.py#L877

but that's probably amongst the only things that might

components/renku_data_services/authz/authz.py
```diff
@@ -375,10 +375,17 @@ class Authz:
...
 
     @property
     def client(self) -> AsyncClient:
         """The authzed DB asynchronous client."""
+        running_loop = asyncio.get_running_loop()
+        # Drop the current client if it is attached to a different event loop
+        if self._loop != running_loop:
+            logger.info(f"Event loop mis-match, old = {self._loop}, new = {running_loop}")
+            self._loop = running_loop
+            self._client = None
```
leafty, line 388: Note: this will hang data-services for some reason (will investigate later), but it makes k8s restart the pod in a good state.

components/renku_data_services/crc/db.py
```diff
@@ -108,6 +112,7 @@ def __init__(self, session_maker: Callable[..., AsyncSession], quotas_repo: Quot
 
     async def initialize(self, async_connection_url: str, rp: models.UnsavedResourcePool) -> None:
```
leafty, line 113: Note: a proper fix is probably to either:
* use a dedicated `Authz` instance -> make another class here for initialization?
* delegate this initialization to the worker -> will use the same thread as the request handlers
* find out why the grpc stuff hangs when we drop it (?!?)

SalimKayal: As a gut feeling I prefer the second option.

components/renku_data_services/authz/authz.py
```diff
@@ -799,6 +806,7 @@ async def decorated_function(
                     result = await f(db_repo, *args, **kwargs)
                     authz_change = await _get_authz_change(db_repo, op, resource, result, *args, **kwargs)
                     await db_repo.authz.client.WriteRelationships(authz_change.apply)
+                    await session.commit()
```
SalimKayal, line 809: Do we still need to keep this here? `with_db_transaction` seems to properly use a context manager when session is not provided

SalimKayal:
Apart from the question on the commit LGTM

components/renku_data_services/authz/authz.py
leafty: Maybe we should have a look at doing that, but I would rather keep the old behavior until we validate that we can do that.

bases/renku_data_services/data_api/main.py
```diff
@@ -163,9 +163,9 @@ async def handle_head(request: Request, response: BaseHTTPResponse) -> None:
     async def do_migrations(_: Sanic) -> None:
         logger.info("running migrations")
         run_migrations_for_app("common")
-        await dependency_manager.rp_repo.initialize(
-            dependency_manager.config.db.conn_url(async_client=False), dependency_manager.default_resource_pool
-        )
+        # await dependency_manager.rp_repo.initialize(
+        #     dependency_manager.config.db.conn_url(async_client=False), dependency_manager.default_resource_pool
+        # )
```
leafty:
```suggestion
```

components/renku_data_services/authz/authz.py
leafty: Maybe we should crash?

components/renku_data_services/crc/db.py
leafty: Second option works.

components/renku_data_services/authz/authz.py
SalimKayal: as long as it doesn't break anything I'm ok with your point of view.

Even in https://github.com/SwissDataScienceCenter/renku-data-services/blob/0cdf55c3a27db9b20f274a1c6af4ef6b0025a06f/components/renku_data_services/crc/db.py#L877

I'm not sure the case would arise where two changes are triggered in the paths currently taken

components/renku_data_services/authz/authz.py
SalimKayal: It does sound saner to raise something

SalimKayal (approved):

components/renku_data_services/authz/authz.py
leafty: It happens, which is why there was one failing test. I have fixed it in e3ba59a7ce641f089153d20618edc18d5d80d6a7 and we can later update the code to handle all authz changes in a single db transaction.

components/renku_data_services/crc/db.py
```diff
@@ -924,14 +938,16 @@ async def update_user_resource_pools(
             else:
                 removed_rps = [rp for rp in user.resource_pools if rp.id not in new_rp_ids]
                 user.resource_pools = list(rps_to_add)
+            # TODO: the authz changes below are done in separate db transactions,
+            # TODO: use only one (collect everything into a single authz change)
```
leafty, line 942: TODO

PR #1267, opened 2026-04-20
feat: add group and projects to resource pool authz

SalimKayal:
## Summary

This PR extends resource pool authorization to support **group-level and project-level membership**, in addition to individual users. It introduces a new Authzed schema (v10) with `group_viewer` and `project_viewer` relations, refactors the CRC user repository into a polymorphic `MemberRepository`, and adds comprehensive tests for inheritance, prohibition, and edge cases.

## Motivation

Previously, resource pools could only be shared with individual users via direct `viewer` relations or blocked via `prohibited`. This did not scale for organizations that manage access through Renku groups or projects. By allowing groups and projects as first-class members of resource pools, access control becomes declarative and maintainable at the team level.

### Authz / SpiceDB Schema
- **New schema v10** (`components/renku_data_services/authz/schemas.py`):
  - Added `group_viewer` and `project_viewer` relations on `resource_pool`.
  - `read` permission now traverses `group_viewer->direct_member` and `project_viewer->direct_member` to resolve inherited access.
  - Preserved existing `viewer`, `public_viewer`, `prohibited`, and `resource_pool_platform->is_admin` semantics.
- **Migration** (`migrations/versions/65f8330c4d25_authz_migrate_to_v10.py`):
  - Upgrades/downgrades Authzed schema with proper relationship cleanup on rollback.

### Domain Models
- Added `MemberType` enum and `ResourcePoolMemberIdentifier` dataclass (`crc/models.py`) to represent polymorphic members (USER, GROUP, PROJECT) with ULID validation for non-user types.
- Extended `Member` model with `subject_type` field (`authz/models.py`) so membership changes carry the target resource type.
- Added `Member.with_resource_pool()` factory for building resource-pool-specific members.

### Repository Refactor
- Renamed `UserRepository` → `MemberRepository` (`crc/db.py`) to reflect its broader scope.
- Injected `GroupRepository` and `ProjectRepository` into `MemberRepository` for resolving group/project identifiers.
- Refactored internal methods:
  - `_grant_resource_pool_members`, `_revoke_resource_pool_members`, `_prohibit_resource_pool_users`, `_unprohibit_resource_pool_users` now accept `ResourcePoolMemberIdentifier` collections.
  - Added `_resolve_members()` to validate and map members to `(id, ResourceType)` tuples, rejecting missing groups/projects with `MissingResourceError`.
- Added public `grant_resource_pool_members` / `revoke_resource_pool_members` admin-only methods.

### Dependency Wiring
[...]

PR #1299, opened 2026-05-05
feat: add REST API for polymorphic resource pool membership

SalimKayal:
## Summary
This PR introduces a new set of admin-only REST endpoints to manage **users, groups, and projects** as first-class members of a resource pool. It refactors the existing user-centric membership logic to use **Authzed (SpiceDB)** as the source of truth for non-default, non-public pools, and aligns the API with the authorization model already established for groups and projects.

### New API endpoints (`components/renku_data_services/crc/api.spec.yaml`)
| Method | Path | Description |
|--------|------|-------------|
| `GET` | `/resource_pools/{id}/members` | List all members (users, groups, projects) with metadata |
| `POST` | `/resource_pools/{id}/members` | Add members to the pool |
| `PUT` | `/resource_pools/{id}/members` | Replace all members with the given set |
| `GET` | `/resource_pools/{id}/members/{type}/{id}` | Check if a specific member belongs to the pool |
| `DELETE` | `/resource_pools/{id}/members/{type}/{id}` | Remove a specific member from the pool |

**Response schemas** are discriminated by `member_type`:
- **User**: `id`, `relation` (`viewer`/`prohibited`), `email`
- **Group**: `id`, `relation` (`group_viewer`), `slug`, `name`
- **Project**: `id`, `relation` (`project_viewer`), `namespace`, `name`

### Authz layer improvements (`components/renku_data_services/authz/authz.py`)
- Added **type-safe `@overload`s** for `_AuthzConverter.to_object`, `_has_permission`, and `has_permission` so that mypy correctly narrows `_ID` types (`str` for users, `ULID` for projects/groups, `int` for resource pools).
- Added `get_resource_pool_members()` to read all direct relationships from Authzed (excluding `public_viewer` and `platform` relations).
- Updated `_add_resource_pool_members` to reject `Role.PROHIBITED` for groups and projects with a clear validation error.

### MemberRepository refactor (`components/renku_data_services/crc/db.py`)
- `get_resource_pool_users` now resolves users **via Authzed** for non-default, non-public pools instead of reading the legacy SQL join. This ensures that group/project inherited permissions are reflected in the user list.
- Added `get_resource_pool_members` -- resolves raw Authzed tuples back to domain objects (users --> emails via Keycloak, groups --> slug+name, projects --> namespace+name) and skips orphaned relations with a warning.
- Added `update_resource_pool_members` -- supports atomic **set semantics** (PUT: compute diff, revoke missing, grant new) and **append semantics** (POST).
- `_resolve_members` now carries the API relation through to the authz layer.
- `_relation_to_role` maps API strings to authorization `Role`s (`viewer`, `prohibited`, `group_viewer`, `project_viewer`).

### Blueprint (`components/renku_data_services/crc/blueprints.py`)
- New `ResourcePoolMembersBP` with `get_all`, `post`, `put`, `get`, and `delete` handlers.
- Admin-only access enforced via `@only_admins`.
- Registered in the main `app.py` under the existing `url_prefix`.

### Model changes (`components/renku_data_services/crc/models.py`)
- `ResourcePoolMemberIdentifier` now carries a `relation` field with sensible defaults per member type.

### Code generation fixes
- Bumped `datamodel-code-generator` to `^0.57.0` to pick up the fix for **discriminator on `list[oneOf]`** (koxudaxi/datamodel-code-generator#1937).
- Replaced the wildcard `*` `before`-validator in `apispec_base.py` with a targeted `model_validator` on `id` to avoid Pydantic discriminator conflicts.

### Tests (`test_resource_pools.py`)
Added comprehensive integration tests covering:
- Adding/removing groups and projects to pools
- PUT replacing members (preserving admin inherited access)
- Mixed-type membership (user + group + project)
- `prohibited` user relations blocking access
- Empty array POST/PUT
- Malformed discriminator payloads (422)
- Nonexistent member references (404)
- Deleting a user with only inherited access is a no-op (204, access preserved via group)
- Existing `/users` endpoints still sync Authzed on PUT

## Migration / compatibility notes
- **Default** and **public** pools retain their existing SQL-backed behaviour; no Authzed tuples are created for them.
- Legacy `/resource_pools/{id}/users` endpoints continue to work but now synchronise Authzed when users are added/removed.
- Admins always see all users (including themselves) when listing pool users, because Authzed resolves inherited `admin` permissions.

## PR Stack

* base: #1311 
  * [this](#1299)


sgaist (approved):

components/renku_data_services/authz/authz.py
```diff
@@ -546,6 +704,46 @@ async def users_with_permission(
...
+    ) -> list[tuple[str, str, str]]:
+        """Get all members of a resource pool from Authzed.
+
+        Returns a list of tuples: (subject_type, subject_id, relation).
+        Skips public_viewer and resource_pool_platform relations.
+        """
+        if isinstance(user, InternalServiceAdmin):
+            pass
+        elif not user.is_admin:
+            return []
```
leafty: Useless check:

```suggestion
        if not user.is_admin:
            return []
```

SalimKayal:
small comment, for readability purpose I suggest #1267 to be reviewed before I collapse this one onto it

leafty:
> small comment, for readability purpose I suggest #1267 to be reviewed before I collapse this one onto it

Can you mark PRs ready in order to be reviewed? The feature branch PR should be empty (or have a dummy change). And then the first PR into the stack can be marked ready.

Also, a simple stack navigation block in the PR description can help.

Otherwise I cannot pick which PRs need to be reviewed in which order. It is rather confusing that the "draft" PR needs to be reviewed first.

SalimKayal:
done

SalimKayal:
@leafty seems ok on the api side ?

components/renku_data_services/crc/apispec_base.py
```diff
@@ -17,13 +16,11 @@ class BaseAPISpec(BaseModel):
...
+    @field_validator("id", mode="before", check_fields=False)
     @classmethod
-    def serialize_ulid(cls, value: Any) -> Any:
-        """Handle ULIDs."""
-        if isinstance(value, ULID):
-            return str(value)
-        return value
+    def serialize_id(cls, v: ULID) -> str:
+        """Custom serializer that can handle ULIDs for id."""
+        return str(v)
```
leafty, line 23: > Replaced the wildcard * before-validator in apispec_base.py with a targeted model_validator on id to avoid Pydantic discriminator conflicts.

Not sure I understand why this is done. Using `*` (and maybe removing the other validators that do the same ULID transform) is easier for developers: they do not need to update the validators EVERY TIME a new field of type ULID is added to the API spec. This has caused issues in the past, so using the `*` validator is an easy solution.

SalimKayal: It somehow fails to work when validating `*`  when the mode is `"before"`. that's why I changed it. I also tried with a `model_validator` to recursively change ULIDs to strings but without success either.

SalimKayal: Is it a major problem ? should I try to make it work regardless ?

leafty (approved):

PR #1301, opened 2026-05-08
fix: user list permissions

olevski:
/deploy


leafty (commented):
What is this supposed to fix? We need publicly accessible user profiles for RenkuLab, e.g. open https://renkulab.io/u/tasko.olevski in a incognito tab. This change will break the platform as is.

PR #1302, opened 2026-05-08
fix: remove some information when listing users

olevski:
/deploy

### DIFF
--- a/components/renku_data_services/users/db.py
+++ b/components/renku_data_services/users/db.py
@@ -145,7 +145,7 @@ class UserRepo(DbUsernameResolver):
             if not user:
                 return None
             # If the user is not admin or they are not requesting their own info we hide some fields
-            strip_sensitive_info = not requested_by.is_admin or requested_by.id != user.id
+            strip_sensitive_info = requested_by.id != user.id and not requested_by.is_admin
             if strip_sensitive_info:
                 user = user.redact_email()
             return user
--- a/components/renku_data_services/users/db.py
+++ b/components/renku_data_services/users/db.py
@@ -140,6 +140,8 @@ class UserRepo(DbUsernameResolver):
         """
         async with self.session_maker() as session, session.begin():
             user = await self.get_user(id=id)
+            if not requested_by.is_admin and user and requested_by.id != user.id:
+                return None
             if not user and id == requested_by.id:
                 return await self._add_api_user(requested_by)
             return user
--- a/components/renku_data_services/users/db.py
+++ b/components/renku_data_services/users/db.py
@@ -140,10 +140,14 @@ class UserRepo(DbUsernameResolver):
         """
         async with self.session_maker() as session, session.begin():
             user = await self.get_user(id=id)
-            if not requested_by.is_admin and user and requested_by.id != user.id:
-                return None
             if not user and id == requested_by.id:
-                return await self._add_api_user(requested_by)
+                user = await self._add_api_user(requested_by)
+            if not user:
+                return None
+            # If the user is not admin or they are not requesting their own info we hide some fields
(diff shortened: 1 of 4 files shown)


leafty (commented):
What is this supposed to fix? We need publicly accessible user profiles for RenkuLab, e.g. open https://renkulab.io/u/tasko.olevski in a incognito tab. This change will break the platform as is.

olevski:
@leafty the problem is that the endpoint reveals the full name and email of the user. We should probably remove that when the user is not authenticated or they are not admin and they are not the same user as themselves.

olevski:
Let me see if I can do this without breaking the UI. I think yes.

leafty:
> @leafty the problem is that the endpoint reveals the full name and email of the user. We should probably remove that when the user is not authenticated or they are not admin and they are not the same user as themselves.

We cannot do this without fundamentally breaking the platform. Users need to find other users based on their name, email or username, e.g. to add to projects, to add to groups, etc.

If there is a concern about which information is exposed and where it is exposed, we need to have a proper documented discussion.

1. This does not "fix" the issue: the information is still available through other endpoints, notably search.
2. This would break some fundamental user flows like finding people to add to groups, projects, etc.

olevski:
@leafty I understand. I think I can remove the email at least from this without breaking things.

components/renku_data_services/users/db.py
```diff
@@ -140,10 +140,14 @@ async def get_or_create_user(self, requested_by: APIUser, id: str) -> UserInfo |
...
-                return None
             if not user and id == requested_by.id:
-                return await self._add_api_user(requested_by)
+                user = await self._add_api_user(requested_by)
+            if not user:
+                return None
+            # If the user is not admin or they are not requesting their own info we hide some fields
+            strip_sensitive_info = not requested_by.is_admin or requested_by.id != user.id
+            if strip_sensitive_info:
+                user = user.redact_email()
```
leafty, line 150: Note: seems OK for now. But we should discuss how to help users identify other users when adding to groups, projects, etc. as impersonation is possible.

components/renku_data_services/users/db.py
```diff
@@ -141,7 +141,13 @@ async def get_or_create_user(self, requested_by: APIUser, id: str) -> UserInfo |
         async with self.session_maker() as session, session.begin():
             user = await self.get_user(id=id)
             if not user and id == requested_by.id:
-                return await self._add_api_user(requested_by)
+                user = await self._add_api_user(requested_by)
+            if not user:
+                return None
+            # If the user is not admin or they are not requesting their own info we hide some fields
+            strip_sensitive_info = not requested_by.is_admin or requested_by.id != user.id
```
leafty: Incorrect logic.

```suggestion
            strip_sensitive_info = requested_by.id != user.id and not requested_by.is_admin
```

`GET /api/data/user` and `GET /api/data/user/:ID` will not return the user's own email.

leafty (changes requested):

leafty (changes requested):
These two tests need to be adjusted:

```
FAILED test/bases/renku_data_services/data_api/test_users.py::test_logged_in_users_can_get_other_users - AssertionError: assert {'id': 'user', 'first_name': 'User', 'last_name': 'Doe', 'email': None} == {'id': 'user', 'first_name': 'User', 'last_name': 'Doe', 'email': 'user.doe@gmail.com'}
  
FAILED test/bases/renku_data_services/data_api/test_users.py::test_anonymous_users_can_get_other_users - AssertionError: assert {'id': 'user', 'first_name': 'User', 'last_name': 'Doe', 'email': None} == {'id': 'user', 'first_name': 'User', 'last_name': 'Doe', 'email': 'user.doe@gmail.com'}
```

Should be good after that.

leafty (approved):

PR #1303, opened 2026-05-08
feat: Add launcher_type to launcher definition

eikek:
This adds the `launcher_type` property to a launcher. Launchers will now be confined to one specific type of session: interactive (what we have now), non_interactive (jobs) and apps.

This is the first iteration to allow working in parallel on top of the new launcher type. The `session_type` currently in the session create request is removed. Starting a session will use the `launcher_type` to determine what type of session to create.

The new `launcher_type` on launchers is optional and set to `interactive` by default. Users are not affected by this change.

/deploy


components/renku_data_services/session/api.spec.yaml
```diff
@@ -552,6 +552,12 @@ components:
           $ref: "#/components/schemas/IsArchivedPatch"
         strip_path_prefix:
           $ref: "#/components/schemas/StripPathPrefixPatch"
+    SessionType:
+      type: string
+      enum:
+        - interactive
+        - non-interactive
+        - app
```
leafty: Suggestion: block apps from the API for now (it can be enabled in the corresponding feature branch).


```suggestion
        # - app  # Not yet supported
```

components/renku_data_services/session/core.py
```diff
@@ -200,6 +200,13 @@ def validate_unsaved_session_launcher(
     else:
         raise errors.ValidationError(message=f"Unexpected environment type: {type(launcher.environment)}")
 
+    session_type = models.SessionType.interactive
+    match launcher.session_type:
+        case apispec.SessionType.non_interactive:
+            session_type = models.SessionType.non_interactive
+        case apispec.SessionType.app:
+            session_type = models.SessionType.app
```
leafty: Suggestion: block apps from the API for now (it can be enabled in the corresponding feature branch).


```suggestion
        case apispec.SessionType.app:
            session_type = models.SessionType.app
            raise errors.ValidationError(message="Apps are not yet supported.")
```

components/renku_data_services/session/models.py
```diff
@@ -235,6 +235,29 @@ def __post_init__(self) -> None:
...
+
+    @property
+    def is_interactive(self) -> bool:
+        """Return true when interactive."""
+        return self == SessionType.interactive
+
+    @property
+    def is_app(self) -> bool:
+        """Return true when this is 'app'."""
+        return self == SessionType.app
```
leafty: Not sure if these are useful. Writing `if sess.session_type == SessionType.app` is very short anyway.

components/renku_data_services/session/core.py
eikek: If I comment it out as above, this then goes away anyways.

components/renku_data_services/session/models.py
eikek: I found it a bit nicer, but I can also remove it.

components/renku_data_services/migrations/versions/e3cade8bf1a8_add_launcher_type_to_launcher.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+def upgrade() -> None:
+    op.execute("CREATE TYPE launchertype as ENUM ('interactive', 'non_interactive', 'app')")
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column(
+        "launchers",
+        sa.Column(
+            "launcher_type",
+            sa.Enum("interactive", "non_interactive", "app", name="launchertype", create_type=True),
```
leafty:
```suggestion
            sa.Enum("interactive", "non_interactive", "app", name="launchertype"),
```

leafty (commented):
Q: Is it a good idea to merge this to `main` now? Would it be wiser to accumulate changes into a feature branch?

eikek:
> Q: Is it a good idea to merge this to `main` now? Would it be wiser to accumulate changes into a feature branch?

I would merge it to `main`. I think once this change is approved, it is done and we can merge. It is more convenient to have the app feature and this one to evolve (and merge) independently when they don't share another branch. But I don't have a strong opinion, we can also branch off of this one.

components/renku_data_services/migrations/versions/e3cade8bf1a8_add_launcher_type_to_launcher.py
eikek: ~I think I tried that without success. But I'll try again. I then found other migrations (e.g. `providerkind` done by you I think) which creates the type using a separate statement. Then I also did it this way.~

Sorry, I misread the diff :-) you mean to remove the non-working piece :) Thanks!!

leafty (dismissed):

components/renku_data_services/notebooks/core_sessions.py
```diff
@@ -1100,13 +1100,15 @@ async def start_session(
                 stripURLPath=environment.strip_path_prefix,
                 env=env,
                 remoteSecretRef=remote_secret.ref() if remote_secret else None,
-                readinessProbe=ReadinessProbe2(type=Type3.none) if not is_interactive else ReadinessProbe2(),
+                readinessProbe=ReadinessProbe2(type=Type3.none)
+                if session_type.is_non_interactive
+                else ReadinessProbe2(),
```
olevski: nitpick: Can you import `ReadinessProbe2` to `crs.py` and re-export it that to rename with a nicer name. Like `ReadinessProbe`

olevski: same for `Type3`

components/renku_data_services/session/orm.py
```diff
@@ -160,6 +160,10 @@ class SessionLauncherORM(BaseORM):
     )
     """Environment variables to set in the session."""
 
+    launcher_type: Mapped[models.LauncherType] = mapped_column(
+        "launcher_type", default=models.LauncherType.interactive, nullable=False, index=True
+    )
```
olevski, line 169: issue: use `server_default` in addition to `default`, `default` is a default only when you instantiate the class instance in python. `server_default` is  a default on the postgres side. Which is probably what you want in the first place.

olevski (changes requested):
Thanks Eike. 2 small things.

components/renku_data_services/session/orm.py
olevski: This may need the migration to be modified too. I.e. the default will be part of the table definition I think.

eikek: The migration already has this default 😌

olevski (dismissed):

eikek:
> > Q: Is it a good idea to merge this to `main` now? Would it be wiser to accumulate changes into a feature branch?
> 
> I would merge it to `main`. I think once this change is approved, it is done and we can merge. It is more convenient to have the app feature and this one to evolve (and merge) independently when they don't share another branch. But I don't have a strong opinion, we can also branch off of this one.

@leafty @olevski what do you think about merging this one?

components/renku_data_services/session/apispec_base.py
```diff
@@ -54,3 +55,25 @@ def convert_path_to_string(cls, val: str | PurePosixPath) -> str:
...
+    all_values = [e.value for e in enum_class]
+    if isinstance(lt, enum_class):
+        return lt
+    if isinstance(lt, str):
+        try:
+            return enum_class(lt)
+        except ValueError as e:
+            raise ValueError(f"Invalid {field_name}: {lt}. Expect one of {all_values}") from e
+
+    raise ValueError(f"{field_name} must be one of: {all_values}")
```
leafty: I am not sure I understand why we have this. The blueprints.py should use pydantic validators already.

eikek: My observation was that the sevrer accepted an invalid value for an enum and used the default in that case. In the beginning I had the value `non-interactive` as the api-spec value. But this didn't work with pydantic defaults, it didn't use the value, but the field definition. Then I tried to add a conversion manually. I do another try, because since then I made all values `non_interactive` (same as field names).

leafty (approved):
LGTM, tested session launcher POSTs.

PR #1304, opened 2026-05-08
fix: avoid inheritance in openapi

mohammad-alisafaee:
The UI code generator fails to correctly interpret OpenAPI inheritance when inherited fields are marked as required.


leafty (approved):

SalimKayal:
Isn't this fixed in recent versions of datamodel-code-generator ?

leafty:
> Isn't this fixed in recent versions of datamodel-code-generator ?

See pr description, this breaks UI codegen (we'll try to update it, but this is preferred for now I think).

PR #1305, opened 2026-05-11
chore: add notice to build strategy

leafty:


sgaist (changes requested):
I think that is not enough. The renku helm chart value `dataService.imageBuilders.strategyName` default value must also be updated to use the new name.

PR #1306, opened 2026-05-11
fix: slow resource requests usage query

### LINKED ISSUE #1284: Getting resource usage times out
Tried via the swagger page on renkulab.io

Renku Version 2.16.2

Tried to get usage over 1 month period for a single resource pool.

Endpoint: `api/data/resource_pools/<resource_pool_id>/usage`

There was a gateway timeout. From the ingress controller.

olevski:
/deploy extra-values=dataService.dataTasks.enableResourceRequestTracking=true


olevski:
This does not fix the problem definitely. We should look into optimizing this a bit more. But it definitely makes the query complete in 10 seconds or so. Rather than in hours. I created a new view in prod to test for this.

eikek (commented):
Hi, I think the difference to the previous version is that same capture dates won't always advance to the next capture date with `lead()`. If we have multiple rows with the same capture_date, `lead()` selects just the next row, possibly having the same capture date.

Here is an example (hoping the queries are correct) with the current data on dev:

Previous (v1):
```
with
  cpu_obs as (
    select id, capture_date, dense_rank() over (order by capture_date asc) as rn
    from resource_pools.resource_requests_log
    where cpu_request is not null and user_id is not null
  )
select t.id, t.capture_date,
  (select min(c.capture_date) from cpu_obs c where c.rn = cs.rn + 1) as cpu_next_date,
  least((select min(c.capture_date) from cpu_obs c where c.rn = cs.rn + 1) - t.capture_date, t.capture_interval) as cpu_time
from resource_pools.resource_requests_log t
left join cpu_obs cs on cs.id = t.id
where t.cpu_request is not null and t.user_id is not null
limit 10
```

Result:
```
             id             |      capture_date      |     cpu_next_date      | cpu_time
----------------------------+------------------------+------------------------+----------
 01KJZZDYG60F9HAYAS5AZB8EB9 | 2026-03-05 21:45:59+00 | 2026-03-05 21:56:00+00 | 00:10:00
 01KK0008TP1KN5BYX7NJ2N358E | 2026-03-05 21:56:00+00 | 2026-03-05 22:06:00+00 | 00:10:00
 01KK00JK17JA0YBW31K28MT9WA | 2026-03-05 22:06:00+00 | 2026-03-05 22:16:00+00 | 00:10:00
 01KK014X67BXDVA76Z3D6YEKD3 | 2026-03-05 22:16:00+00 | 2026-03-06 14:57:23+00 | 00:10:00
 01KK1TEGG8G2P934N9NQ6KXSX6 | 2026-03-06 14:57:23+00 | 2026-03-06 15:07:24+00 | 00:10:00
 01KK1TEGG9SFFVJ0KJ7PV1VW1K | 2026-03-06 14:57:23+00 | 2026-03-06 15:07:24+00 | 00:10:00
 01KK1TEGGASS7V32BQPN118RJ3 | 2026-03-06 14:57:23+00 | 2026-03-06 15:07:24+00 | 00:10:00
 01KK1V0V3MJ150HKSG5TFVR3W6 | 2026-03-06 15:07:24+00 | 2026-03-09 13:16:14+00 | 00:10:00
 01KK1V0V3N9TT4KMSDDGJP807Y | 2026-03-06 15:07:24+00 | 2026-03-09 13:16:14+00 | 00:10:00
 01KK1V0V3N2HRP7HKB2R4PZ58Y | 2026-03-06 15:07:24+00 | 2026-03-09 13:16:14+00 | 00:10:00
```

Notice that rows 5,6,7 and 8,9,10 have same `capture_date`.

New (v2):
```
select id, capture_date,
  lead(capture_date) over (order by capture_date asc) as cpu_next_date,
  least(lead(capture_date) over (order by capture_date asc) - capture_date, capture_interval) as cpu_time
from resource_pools.resource_requests_log
where cpu_request is not null and user_id is not null
limit 10;
```

Result:
```
             id             |      capture_date      |     cpu_next_date      | cpu_time
----------------------------+------------------------+------------------------+----------
 01KJZZDYG60F9HAYAS5AZB8EB9 | 2026-03-05 21:45:59+00 | 2026-03-05 21:56:00+00 | 00:10:00
 01KK0008TP1KN5BYX7NJ2N358E | 2026-03-05 21:56:00+00 | 2026-03-05 22:06:00+00 | 00:10:00
 01KK00JK17JA0YBW31K28MT9WA | 2026-03-05 22:06:00+00 | 2026-03-05 22:16:00+00 | 00:10:00
 01KK014X67BXDVA76Z3D6YEKD3 | 2026-03-05 22:16:00+00 | 2026-03-06 14:57:23+00 | 00:10:00
 01KK1TEGG8G2P934N9NQ6KXSX6 | 2026-03-06 14:57:23+00 | 2026-03-06 14:57:23+00 | 00:00:00
 01KK1TEGG9SFFVJ0KJ7PV1VW1K | 2026-03-06 14:57:23+00 | 2026-03-06 14:57:23+00 | 00:00:00
 01KK1TEGGASS7V32BQPN118RJ3 | 2026-03-06 14:57:23+00 | 2026-03-06 15:07:24+00 | 00:10:00
 01KK1V0V3MJ150HKSG5TFVR3W6 | 2026-03-06 15:07:24+00 | 2026-03-06 15:07:24+00 | 00:00:00
 01KK1V0V3N9TT4KMSDDGJP807Y | 2026-03-06 15:07:24+00 | 2026-03-06 15:07:24+00 | 00:00:00
 01KK1V0V3N2HRP7HKB2R4PZ58Y | 2026-03-06 15:07:24+00 | 2026-03-09 13:16:14+00 | 00:10:00
```

it might not make a real difference for the real word use cases, though. or we could also just tweak the `least` to include `capture_interval` as a fallback?

olevski:
@eikek this is ready now. Tests on prod with the view manually created completed within 10-20 seconds. And this is without the indices on uid or phase or any of the other additions in this PR.

leafty (commented):
Why is the old view not deleted in the migrations? Is there a reason to not do a replace?

eikek (dismissed):
The only thing I noticed is that we now use one interval for all measurements (cpu, memory, disk, gpu). Before each one had its own time. But this is fine by the way we are collecting the information. 👍🏼

olevski:
> Why is the old view not deleted in the migrations? Is there a reason to not do a replace?

It does not take up any space so I thought I would leave it in case we need to check / compare things or go back to it.

eikek (approved):

leafty:
> > Why is the old view not deleted in the migrations? Is there a reason to not do a replace?
> 
> It does not take up any space so I thought I would leave it in case we need to check / compare things or go back to it.

I hope you'll remember in two years why there is a `view` and a `view_v2`. 😅

PR #1307, opened 2026-05-12
feat: enforce quota

mohammad-alisafaee:
* Add a `quota_enforced` field to `ResourceClass`, defaulting to true.
* Hibernate sessions if the quota is exhausted and the resource class has `quota_enforced` enabled.
* Prevent starting/resuming sessions if the resource class has no remaining quota and enforcement is enabled.

# TODO:
To be able to hibernate sessions that exhausted their quota, `ServiceAdminId.capacity_reservation` needs to have the permission to hibernate sessions.

/deploy #extra-values=dataService.dataTasks.enableResourceRequestTracking=true


components/renku_data_services/crc/core.py
```diff
@@ -644,11 +648,7 @@ async def calculate_usage_hours(
 
     # NOTE: Calculate quota based on the minimum of user and total limits that is set. User and total limits can be set
     # independent of each other.
-    total_quota = (
-        min(limits.user_limit.value, limits.total_limit.value)
-        if limits.user_limit.value and limits.total_limit.value
-        else max(limits.user_limit.value, limits.total_limit.value)
-    )
+    total_quota = get_nonzero_minimum(limits.user_limit.value, limits.total_limit.value)
```
sgaist: `get_nonzero_minimum` reads strangely with the `<= 0`  check just below especially since it can in fact return 0.

Maybe something like `get_quota_value` ?

components/renku_data_services/notebooks/core_sessions.py
```diff
@@ -789,6 +792,22 @@ def get_remote_env(
     return env
 
 
+async def _check_quota(resource_usage_service: ResourceUsageService, resource_pool_id: int, user_id: str) -> None:
+    """Raise a ValidationError if the user has exhausted their quota, do nothing otherwise."""
+    usage = await resource_usage_service.get_running_week(resource_pool_id=resource_pool_id, user_id=user_id)
+    if usage is None:
+        return
+
+    total_quota = get_nonzero_minimum(usage.pool_limits.user_limit.value, usage.pool_limits.total_limit.value)
```
sgaist: Same as above

components/renku_data_services/crc/api.spec.yaml
```diff
@@ -1191,6 +1198,10 @@ components:
           $ref: "#/components/schemas/NodeAffinityList"
         id:
           $ref: "#/components/schemas/IntegerId"
+        quota_enforced:
+          allOf:
+            - $ref: "#/components/schemas/QuotaEnforced"
+          default: true
```
olevski: A few things:
- Why is the default `true`? It should default to no quota enforcement because currently the API does not enforce quotas at all. Setting it to default to true feels like a breaking API change. Especially since we dont have this in the UI for the admin panel defaulting to having quota enforcement will surprise admins.
- I thought this type of `allOf` syntax did not work with the code generation on the ui.

components/renku_data_services/crc/models.py
```diff
@@ -93,6 +93,7 @@ class UnsavedResourceClass(ResourcesCompareMixin):
     default_storage: int = 1
     node_affinities: list[NodeAffinity] = field(default_factory=list)
     tolerations: list[str] = field(default_factory=list)
+    quota_enforced: bool = True
```
olevski: Same as the APISpec. This should default to `false`.

components/renku_data_services/crc/orm.py
```diff
@@ -97,6 +97,7 @@ class ResourceClassORM(BaseORM):
     default_storage: Mapped[int] = mapped_column(BigInteger)
     default: Mapped[bool] = mapped_column(default=False)
     gpu: Mapped[int] = mapped_column(BigInteger, default=0)
+    quota_enforced: Mapped[bool] = mapped_column(default=True)
```
olevski: Add a server default on the sql side. Set it to `false`. So that when we add the migration they are all set to `False`. And we do not want all old classes to have enforcement. We want the admins to gradually opt into enforcement.

There is a special field `server_default` that will put a default value in the table on the sql side. Use that in addition to the `default` field. The `default` field just adds a value to the instance of the class.

olevski (changes requested):
A few changes Mohammad. Otherwise this looks good.

components/renku_data_services/crc/api.spec.yaml
leafty:
```
quota_enforced:
          allOf:
            - $ref: "#/components/schemas/QuotaEnforced"
          default: true
```

This should work once https://github.com/SwissDataScienceCenter/renku-ui/pull/4169 is merged.

components/renku_data_services/crc/core.py
mohammad-alisafaee: I renamed it to `get_effective_quota` and update the docstring to reflect the functionality:
```
def get_effective_quota(a: int, b: int) -> int:
    """Return the minimum nonzero value; return zero only if both values are zero."""
```

components/renku_data_services/crc/api.spec.yaml
mohammad-alisafaee: > Why is the default `true`? It should default to no quota enforcement because currently the API does not enforce quotas at all. Setting it to default to true feels like a breaking API change. Especially since we dont have this in the UI for the admin panel defaulting to having quota enforcement will surprise admins.

This flag won't do anything if a quota isn't set on the resource class. The reason for setting it to true was exactly what you mentioned: We don't have this in the UI yet. Setting it to false means it will never work, while setting it to true allows us to control it by setting or unsetting the quota.

I changed the default value to be false.

> I thought this type of `allOf` syntax did not work with the code generation on the ui.

I believe this works correctly there. Generation there had an issue with inheritance. This exact syntax was used in my previous PR and it worked there. The PR that Flora mentioned is merged now and this shouldn't be an issue anyways.

components/renku_data_services/crc/orm.py
mohammad-alisafaee: I set `server_default` in the Alembic script:

https://github.com/SwissDataScienceCenter/renku-data-services/blob/cdc438ac096dd2deb6bb57f7efaaaa9f86d15931/components/renku_data_services/migrations/versions/a8f0e7b3c2d1_add_quota_enforced_to_resource_class.py#L25-L25

Setting it here isn't strictly necessary because we don't generate the schema from SQLAlchemy models. Anyways, I added it to be consistent with other places in the code.

olevski (dismissed):

olevski (approved):

PR #1308, opened 2026-05-12
feat: add /apps create and get endpoints

wesjdj:


components/renku_data_services/renku_apps/k8s_client.py
```diff
@@ -0,0 +1,182 @@
...
+    "autoscaling.knative.dev/min-scale": "0",
+    "autoscaling.knative.dev/max-scale": "3",
+    "autoscaling.knative.dev/scale-to-zero-pod-retention-period": "2m",
+}
+
+
+def _generate_app_name(project: Project) -> str:
+    """Generate a DNS-1035 label name for an app from its project path."""
+    namespace = project.namespace.path.serialize().replace("/", "-")
+    return f"{project.slug}-{namespace}".lower()[:63]
```
olevski: @wesjdj I thought we discussed about adding some hash at the end of this to avoid collisions. You can look at how we do this for the user sessions. We has a bunch of things that make the session unique and then take the first 10 or so characters from the hash.

components/renku_data_services/renku_apps/k8s_client.py
```diff
@@ -0,0 +1,182 @@
...
+
+    def __init__(self, client: K8sClusterClientsPool, cluster_repo: ClusterRepository) -> None:
+        self.__client = client
+        self.__cluster_repo = cluster_repo
+
+    async def create_app_deployment(
+        self, session_launcher: SessionLauncher, resource_class: ResourceClass | None, project: Project
+    ) -> AppRuntimeState:
+        """Create a deployment for the given app and return its observed runtime state."""
+        cluster_id: ClusterId = DEFAULT_K8S_CLUSTER
```
olevski: suggestion: it makes sense to limit apps to the default cluster. So I would pull out the cluster_id and add it as a parameter on the class. I.e. make it `self.__cluster_id` in the `__init__`.

components/renku_data_services/renku_apps/k8s_client.py
```diff
@@ -0,0 +1,182 @@
...
+
+    return KnativeService.model_validate(
+        {
+            "apiVersion": "serving.knative.dev/v1",
+            "kind": "Service",
+            "metadata": {
+                "name": app_name,
+                "annotations": {
+                    "renku.io/launcher_id": str(session_launcher.id),
+                    "renku.io/project_id": str(session_launcher.project_id),
```
olevski: suggestion(non-blocking): We can revisit this but you will most likely have to filter apps for a specific project. So I would put either both or at least the project_id in the labels. Quick filtering is possible on labels but not on annotations in k8s.

components/renku_data_services/renku_apps/k8s_client.py
```diff
@@ -0,0 +1,182 @@
...
+                },
+            },
+            "spec": {
+                "template": {
+                    "metadata": {"annotations": _APP_AUTOSCALING_ANNOTATIONS},
+                    "spec": {"containers": [container]},
+                },
+            },
```
olevski, line 159: Add the same metadata from above in here so that it propagates to the children. I think it will propagate only to the Revision but still better than nothing. We will know which revision belongs to which project and launcher when troubleshooting or checking things.

components/renku_data_services/renku_apps/repository.py
```diff
@@ -0,0 +1,67 @@
...
+
+        resource_class: ResourceClass | None = None
+        if launcher.resource_class_id is not None:
+            resource_class = await self.rp_repo.get_resource_class(user, launcher.resource_class_id)
+
+        project = await self.project_repo.get_project(user, launcher.project_id)
+        if await self.k8s_client.get_app_deployment_for_project(project) is not None:
+            raise errors.ConflictError(message=f"An app already exists for project '{launcher.project_id}'.")
+        runtime_state = await self.k8s_client.create_app_deployment(launcher, resource_class, project)
+        return build_app(launcher, runtime_state)
```
olevski, line 56: Check that the launcher type is right - i.e. of type app. We can delay this for a follwup pr. But lets not forget it.

bases/renku_data_services/k8s_cache/main.py
```diff
@@ -47,7 +48,7 @@ async def main() -> None:
     ):
         clusters[client.get_cluster().id] = client
 
-    kinds = [AMALTHEA_SESSION_GVK]
+    kinds = [AMALTHEA_SESSION_GVK, KNATIVE_SERVICE_GVK]
     if dm.config.v1_services.enabled:
         kinds.append(JUPYTER_SESSION_GVK)
     if dm.config.image_builders.enabled:
```
olevski, line 54: issue: We should put this behind a config.knative.enabled flag. Beacuse if we do this and there is no knative in the cluster the watcher and clients keep spamming warnings or I cannot remember maybe they even freak out and crash. But I think it is just warnings. Still they pollute the logs.

olevski (changes requested):
This looks great stuff Wes. I just have high level / picky things.

components/renku_data_services/renku_apps/k8s_client.py
wesjdj: @olevski We'd mentioned that for the URLs, but yes, worth doing for the app names, too!

components/renku_data_services/renku_apps/k8s_client.py
wesjdj: Done, thanks!

components/renku_data_services/renku_apps/k8s_client.py
wesjdj: Good catch, thank you!

components/renku_data_services/renku_apps/repository.py
wesjdj: Yep - will sort this in a follow up PR

bases/renku_data_services/k8s_cache/main.py
wesjdj: Done, thanks! I also gated the data API side.

olevski (approved):

Issue #1176, opened 2026-01-16
Admin user deletion fails with authorization error despite misleading success logs


SalimKayal:
### Summary
When attempting to delete any admin user through the Renku API, the operation returns a permission error (code 1404) with misleading success logs. The user remains in both databases despite logs claiming successful deletion. This occurs even when deleting different admin users, and the authorization system (SpiceDB) consistently blocks the operation.

### Steps to Reproduce

1. Log in as an admin user
2. Attempt to delete another user (Admin2) via the Swagger API endpoint
3. **Note**: Deletion of Admin1 user (`ea209425-eaa2-43d8-9271-1cda870ffd3c`) also fails with the same error
4. Verify the user remains in both databases

### Expected Behavior

- Admin users should be able to delete other users
- Authorization checks should pass for valid operations
- Logs should accurately reflect transaction outcomes

### Actual Behavior

**API Response:**
```json
{
  "error": {
    "code": 1404,
    "message": "The user with ID df9b9993-ed85-4f81-8a5a-64692e0a7f62 cannot perform operation Scope.DELETE on user with ID 00c40bad-a4bf-40a0-8ff5-5fb7de3f6282 or the resource does not exist."
  }
}
```

**Misleading Logs:**
```
2026-01-15T16:56:02.931631 [INFO] - remove_user: Trying to remove user with ID 00c40bad-a4bf-40a0-8ff5-5fb7de3f6282
2026-01-15T16:56:02.936110 [INFO] - User with ID 00c40bad-a4bf-40a0-8ff5-5fb7de3f6282 was removed from the database.
2026-01-15T16:56:02.936197 [INFO] - User namespace with ID 00c40bad-a4bf-40a0-8ff5-5fb7de3f6282 was removed from the authorization database.
```

**Database State - User NOT Deleted:**

Renku database (`users.users`):
```
             keycloak_id              | first_name | last_name |         email         | id
--------------------------------------+------------+-----------+-----------------------+----
 ea209425-eaa2-43d8-9271-1cda870ffd3c | Admin1     | ADMIN1    | admin1.admin1@myorg.com | 11
 00c40bad-a4bf-40a0-8ff5-5fb7de3f6282 | Admin2     | ADMIN2    | admin2.admin2@myorg.com | 12
 df9b9993-ed85-4f81-8a5a-64692e0a7f62 | Admin2     | ADMIN2    | admin2.admin2@myorg.com | 13
```

[...]

PR #1311, opened 2026-05-18
feat: resource-pool authz for groups and projects

SalimKayal:
## Summary

This PR extends resource pool authorization to support **group-level and project-level membership**, and introduces a new admin-only REST API to manage resource pool members declaratively. It upgrades the Authzed schema to v10 with `group_viewer` and `project_viewer` relations, refactors the CRC repository layer into a polymorphic `MemberRepository`, and adds comprehensive integration tests for inheritance, prohibition, and edge cases.

## Motivation

Previously, resource pools could only be shared with individual users via direct `viewer` relations or blocked via `prohibited`. This did not scale for organizations that manage access through Renku groups or projects. By allowing groups and projects as first-class members of resource pools, access control becomes declarative and maintainable at the team level.

### Authz / SpiceDB Schema
- **New schema v10** (`components/renku_data_services/authz/schemas.py`):
  - Added `group_viewer` and `project_viewer` relations on `resource_pool`.
  - `read` permission now traverses `group_viewer->direct_member` and `project_viewer->direct_member` to resolve inherited access.
  - Preserved existing `viewer`, `public_viewer`, `prohibited`, and `resource_pool_platform->is_admin` semantics.
- **Migration** (`migrations/versions/65f8330c4d25_authz_migrate_to_v10.py`):
  - Upgrades/downgrades Authzed schema with proper relationship cleanup on rollback.

### Domain Models
- Added `MemberType` enum and `ResourcePoolMemberIdentifier` dataclass (`crc/models.py`) to represent polymorphic members (USER, GROUP, PROJECT) with ULID validation for non-user types.
- Extended `Member` model with `subject_type` field (`authz/models.py`) so membership changes carry the target resource type.
- Added `Member.with_resource_pool()` factory for building resource-pool-specific members.

### Repository Refactor
- Renamed `UserRepository` → `MemberRepository` (`crc/db.py`) to reflect its broader scope.
- Injected `GroupRepository` and `ProjectRepository` into `MemberRepository` for resolving group/project identifiers.
- Refactored internal methods:
  - `_grant_resource_pool_members`, `_revoke_resource_pool_members`, `_prohibit_resource_pool_users`, `_unprohibit_resource_pool_users` now accept `ResourcePoolMemberIdentifier` collections.
  - Added `_resolve_members()` to validate and map members to `(id, ResourceType)` tuples, rejecting missing groups/projects with `MissingResourceError`.
- Added public `grant_resource_pool_members` / `revoke_resource_pool_members` admin-only methods.
- `get_resource_pool_users` now resolves users **via Authzed** for non-default, non-public pools instead of reading the legacy SQL join. This ensures that group/project inherited permissions are reflected in the user list.
- Added `get_resource_pool_members` — resolves raw Authzed tuples back to domain objects (users → emails via Keycloak, groups → slug+name, projects → namespace+name) and skips orphaned relations with a warning.
- Added `update_resource_pool_members` — supports atomic **set semantics** (PUT: compute diff, revoke missing, grant new) and **append semantics** (POST).
- `_relation_to_role` maps API strings to authorization `Role`s (`viewer`, `prohibited`, `group_viewer`, `project_viewer`).

### New REST API (`components/renku_data_services/crc/api.spec.yaml`)
| Method | Path | Description |
|--------|------|-------------|
| `GET` | `/resource_pools/{id}/members` | List all members (users, groups, projects) with metadata |
| `POST` | `/resource_pools/{id}/members` | Add members to the pool |
| `PUT` | `/resource_pools/{id}/members` | Replace all members with the given set |
| `GET` | `/resource_pools/{id}/members/{type}/{id}` | Check if a specific member belongs to the pool |
| `DELETE` | `/resource_pools/{id}/members/{type}/{id}` | Remove a specific member from the pool |

**Response schemas** are discriminated by `member_type`:
- **User**: `id`, `relation` (`viewer`/`prohibited`), `email`
- **Group**: `id`, `relation` (`group_viewer`), `slug`, `name`
- **Project**: `id`, `relation` (`project_viewer`), `namespace`, `name`

### Authz layer improvements (`components/renku_data_services/authz/authz.py`)
- Added **type-safe `@overload`s** for `_AuthzConverter.to_object`, `_has_permission`, and `has_permission` so that mypy correctly narrows `_ID` types (`str` for users, `ULID` for projects/groups, `int` for resource pools).
- Added `get_resource_pool_members()` to read all direct relationships from Authzed (excluding `public_viewer` and `platform` relations).
- Updated `_add_resource_pool_members` to reject `Role.PROHIBITED` for groups and projects with a clear validation error.

### Blueprint & Dependency Wiring
- New `ResourcePoolMembersBP` (`crc/blueprints.py`) with `get_all`, `post`, `put`, `get`, and `delete` handlers.
- Admin-only access enforced via `@only_admins`.
- Updated `DependencyManager`, blueprints, and app registration to use `member_repo` instead of `user_repo` throughout `data_api`.

### Supporting Queries
- Added `GroupRepository.get_group_by_id()` (used internally for member resolution).
- Added `ProjectRepository.get_project_by_id()` (admin-only, for resolving project references back to human-readable paths).

### Code generation fixes
- Bumped `datamodel-code-generator` to `^0.57.0` to pick up the fix for **discriminator on `list[oneOf]`** (koxudaxi/datamodel-code-generator#1937).
- Replaced the wildcard `*` `before`-validator in `apispec_base.py` with a targeted `model_validator` on `id` to avoid Pydantic discriminator conflicts.

### Tests
- **Schema tests** (`test_schemas.py`): Full v10 fixture with `assertTrue`/`assertFalse` and validation paths for public, private, group-inherited, and project-inherited pools.
- **Authorization tests** (`test_authorization.py`):
  - `test_resource_pool_member_relationships_subject_types`: verifies correct subject types in authz updates.
  - `test_resource_pool_polymorphic_inheritance`: confirms group/project members gain access via `direct_member`.
  - `test_resource_pool_polymorphic_prohibition`: confirms `prohibited` still blocks regardless of membership type.
  - `test_resource_pool_admin_bypass_polymorphic`: confirms admin bypass.
  - `test_resource_pool_no_indirect_inheritance`: confirms only `direct_member` is traversed (not namespace owners).
- **Integration tests** (`test_resource_pools.py`):
  - Adding/removing groups and projects to pools
  - PUT replacing members (preserving admin inherited access)
  - Mixed-type membership (user + group + project)
  - `prohibited` user relations blocking access
  - Empty array POST/PUT
  - Malformed discriminator payloads (422)
  - Nonexistent member references (404)
  - Deleting a user with only inherited access is a no-op (204, access preserved via group)
  - Existing `/users` endpoints still sync Authzed on PUT
- **Repository tests** (`test_sqlalchemy_pool_repo.py`): coverage for `get_resource_pool_members` resolution and error paths.

## Backward Compatibility

- Existing user-only pools continue to work unchanged.
- The old `viewer` relation on `resource_pool` is preserved; only `group_viewer` and `project_viewer` are new.
- Default and public pools retain their existing SQL-backed behaviour; no Authzed tuples are created for them.
- Legacy `/resource_pools/{id}/users` endpoints continue to work but now synchronise Authzed when users are added/removed.
- Admins always see all users (including themselves) when listing pool users, because Authzed resolves inherited `admin` permissions.
- The DB migration is reversible (downgrade restores v9 and deletes v10 resource_pool relationships).
- **Warning**: Downgrading removes the `group_viewer` and `project_viewer` relationships, revoking access to the resource pool.

## Checklist

- [x] Authzed schema v10 defined and migrated
- [x] `MemberRepository` supports user/group/project identifiers
- [x] Admin-only grant/revoke API surface ready
- [x] SpiceDB schema tests pass for inheritance and prohibition
- [x] Authorization integration tests pass
- [x] REST API integration tests pass
- [x] datamodel-code-generator bumped and discriminator validated
- [x] Backward compatibility verified for default/public pools and legacy `/users` endpoints

---
  
/deploy


leafty (commented):
I tried to create a private resource pool as an admin and this is what I am getting:

Image

```
GET https://renku-ci-ds-1311.dev.renku.ch/api/data/resource_pools/2/users
```

```json
{
    "error": {
        "code": 1404,
        "message": "The user with ID dc1cec24-b5b3-49d3-a59e-c5212dde8bba cannot perform operation Scope.READ on resource_pool with ID 2 or the resource does not exist."
    }
}
```

But the `user` endpoint returns `is_admin: true` (and I have the admin role in Keycloak).

components/renku_data_services/migrations/versions/65f8330c4d25_authz_migrate_to_v10.py
```diff
@@ -14,7 +14,7 @@
 
 # revision identifiers, used by Alembic.
 revision = "65f8330c4d25"
-down_revision = "cd424c01676e"
+down_revision = "e3cade8bf1a8"
```
leafty: Note: when you do this, the deployment gets broken. To fix it, you have to manually downgrade the alembic revision and then redeploy.

SalimKayal: thanks for the info

SalimKayal:
> I tried to create a private resource pool as an admin and this is what I am getting:

yep let me fix this and I'll update you

SalimKayal:
I wonder if it is linked with #1176

leafty:
It is working now, so I'm not sure what went wrong.

components/renku_data_services/crc/blueprints.py
```diff
@@ -245,6 +253,169 @@ async def _delete(
...
+    authenticator: base_models.Authenticator
+
+    def get_all(self) -> BlueprintFactoryResponse:
+        """Get all members of a specific resource pool."""
+
+        @authenticate(self.authenticator)
+        @only_admins
+        @validate_db_ids
+        async def _get_all(_: Request, user: base_models.APIUser, resource_pool_id: int) -> HTTPResponse:
+            members = await self.repo.get_resource_pool_members(user, resource_pool_id)
```
leafty, line 270: This endpoint doesn't validate that the resource pool exists. In this case it should return a 404 error.

Also, the returned members do not correspond to what the users endpoint is returning: maybe the users endpoint returns indirect members?

SalimKayal: Noted, will fix.

it does indeed. now I resolve all `users` that are authorized through the /users/ enbpoint.

SalimKayal:
> It is working now, so I'm not sure what went wrong.

I think it's an authzed sync issue, per #1176

it feels like there's a lag between getting admin in keycloak and getting admin in authzed

SalimKayal:
other bug, the default resource pool does not seem to appear to non admin users

the response to /resource_pools is `[]` as salim.kayal-rk3@idiap.ch (non admin, non part of groups or projects.

SalimKayal:
finally when the schemathesis tests fails I don't believe it's a flake.

leafty:
> finally when the schemathesis tests fails I don't believe it's a flake.

? They did not fail here.

SalimKayal:
> > finally when the schemathesis tests fails I don't believe it's a flake.
> 
> ? They did not fail here.

not on this run but you can see it happen here 
https://github.com/SwissDataScienceCenter/renku-data-services/actions/runs/26214142870/job/77133165598

SalimKayal:
@leafty everything seems fixed

components/renku_data_services/crc/blueprints.py
SalimKayal: I raise `MissingResourceError` from `crc/db.py` now. it 404 as intended.

when this is merged, I will update the admin page so that it works directly with `members` instead of `users`. this way it will be less confusing for admins.

components/renku_data_services/authz/authz.py
```diff
@@ -275,7 +317,9 @@ def decorator(
         async def decorated_function(
             self: "Authz", user: base_models.APIUser, *args: _P.args, **kwargs: _P.kwargs
         ) -> _T:
-            if isinstance(user, base_models.InternalServiceAdmin):
+            if isinstance(user, base_models.InternalServiceAdmin) or (
+                isinstance(user, base_models.APIUser) and user.is_admin
+            ):
```
leafty: I don't know if we should skip here too for ` isinstance(user, base_models.APIUser) and user.is_admin` 🤔 .

components/renku_data_services/authz/authz.py
```diff
@@ -540,6 +708,44 @@ async def users_with_permission(
...
+        )
+        responses: AsyncIterable[ReadRelationshipsResponse] = self.client.ReadRelationships(
+            ReadRelationshipsRequest(
+                consistency=consistency,
+                relationship_filter=rel_filter,
+            )
+        )
+
+        members: list[tuple[str, str, str]] = []
+        skip_relations = {_Relation.public_viewer.value, _Relation.resource_pool_platform.value}
```
leafty, line 741: Potentially, we should also skip the admin platform relationship type.

leafty (commented):
Non blocking comments here.

I think listing admin users as members of all pools at the moment is going to confuse admins as it is not possible to remove admin from resource pools (and they are not "direct" members, except if added specifically).

See screenshot: the "remove" buttons do not do anything since you cannot remove admins from resource pools.
Image

leafty (commented):
I have different results for the old `users` API vs the new `members` API:

```
GET https://renku-ci-ds-1311.dev.renku.ch/api/data/resource_pools/3/users
```
```json
[
  {
    "id": "97871a35-fb4b-4afb-a97e-a1cb4b413266",
    "no_default_access": false
  },
  {
    "id": "b94e7b6b-262b-4158-9bae-e15cf5a7824a",
    "no_default_access": false
  },
  {
    "id": "95e6f10e-9e7c-4126-8a80-49df9f9c1f6b",
    "no_default_access": false
  }
]
```

```
GET' https://renku-ci-ds-1311.dev.renku.ch/api/data/resource_pools/3/members
```
```json
[
  {
    "member_type": "group",
    "id": "01KSJAHGYHBJRCQ4HFA5KJ4EYM",
    "relation": "group_viewer",
    "slug": "test-group",
    "name": "Test group"
  },
  {
    "member_type": "project",
    "id": "01KSJAHB3GJJJRZ8WW3A14Z28S",
    "relation": "project_viewer",
    "namespace": "flora.thiebaut/test-project",
    "name": "Test project"
  },
  {
    "member_type": "user",
    "id": "97871a35-fb4b-4afb-a97e-a1cb4b413266",
    "relation": "viewer",
    "email": "salim.kayal@idiap.ch"
  },
  {
    "member_type": "user",
    "id": "b94e7b6b-262b-4158-9bae-e15cf5a7824a",
    "relation": "viewer",
    "email": "salim.kayal-rk1@idiap.ch"
  }
]
```

The users are not the same: admins are always added in the old users endpoint.

SalimKayal:
My thought process was that the admin page should show `members` not `users` and that `users` should represent "anyone that has access". I will make that backward compatible.

There is a problem for backward compatibility though, `/users` will show users with access but deleting them won't work when they are added through a group or a project in the admin panel.

The proper way would be to adapt the admin panel to show members and their properties. It remain unclear to me as to how to enhance admin experience.

components/renku_data_services/authz/authz.py
SalimKayal: I can do that

components/renku_data_services/authz/authz.py
SalimKayal: This one was there so that the case where a freshly added admin in authzed is not yet properly synced, per your previous remark

https://github.com/SwissDataScienceCenter/renku-data-services/pull/1311#pullrequestreview-4343225483

I may revert this if you think it's better

leafty:
> > It is working now, so I'm not sure what went wrong.
> 
> I think it's an authzed sync issue, per #1176
> 
> it feels like there's a lag between getting admin in keycloak and getting admin in authzed

I will make a new issue: our current setup syncs Keycloak roles to Authz every three hours which is a potential security issue.

components/renku_data_services/authz/authz.py
leafty: @olevski What do you think? I would suppose we want to short-circuit the authz checks as little as possible.

The authz sync is a separate issue: we sync roles from Keycloak only every three hours, so we should change that to make the role changes sync to Authz immediately if possible.

leafty (commented):
Looks ok, maybe we should just not keep the is_admin short-circuit.

leafty:
https://github.com/SwissDataScienceCenter/renku-data-services/issues/1330

components/renku_data_services/authz/authz.py
olevski: `user.is_admin` is coming from the access token. So if we want quick responsiveness then using the token would have a delay equal to how long the access token from Keycloak is valid. I was thinking I would rather stick with 1 single source of truth. And I think that should be Authzed. Especially since if we add more complicated logic about what admins are or are not supposed to do then that would land in Authzed.

But for now we can live with this. We can have a set of issues we try to tackle in the extended cooldown to address user syncing, user management and syncing with Authzed.

components/renku_data_services/authz/authz.py
```diff
@@ -540,6 +708,44 @@ async def users_with_permission(
...
+        *,
+        zed_token: ZedToken | None = None,
+    ) -> list[tuple[str, str, str]]:
+        """Get all members of a resource pool from Authzed.
+
+        Returns a list of tuples: (subject_type, subject_id, relation).
+        Skips public_viewer and resource_pool_platform relations.
+        """
+        if not user.is_admin:
+            return []
```
olevski: issue: raise an error here, this will make someone think they can see the users in the pool and there is no one in it.

components/renku_data_services/authz/schemas.py
```diff
@@ -863,3 +863,112 @@ def generate_v4(public_project_ids: Iterable[str]) -> AuthzSchemaMigration:
...
+
+v10 = AuthzSchemaMigration(
+    up=[WriteSchemaRequest(schema=_v10)],
+    down=[
+        DeleteRelationshipsRequest(
+            relationship_filter=RelationshipFilter(resource_type=ResourceType.resource_pool.value)
+        ),
+        WriteSchemaRequest(schema=_v9),
+    ],
+)
```
olevski, line 983: question: So the relationships you added from v9 to v10 are `group_viewer` and `project_viewer`. So you should delete those only when you downgrade. This will delete all relationships for the Resource pools. The filter can filter by relationship name. So you need 2 one for `group_viewer` and another for `project_viewer`.

components/renku_data_services/authz/models.py
```diff
@@ -75,6 +75,19 @@ def with_group(self, group_id: ULID) -> "Member":
...
+    def with_resource_pool(
+        self, resource_id: ULID | int, resource_type: ResourceType, subject_type: ResourceType
+    ) -> "Member":
+        """Turn to member with group."""
+        return Member(
+            role=self.role,
+            user_id=self.user_id,
+            resource_id=resource_id,
+            resource_type=resource_type,
+            subject_type=subject_type,
```
olevski, line 90: Update the docstring accordingly. It is the same as `with_group`.

components/renku_data_services/authz/authz.py
```diff
@@ -413,6 +520,67 @@ async def _has_permission(
...
+        resource_type: Literal[ResourceType.resource_pool],
+        resource_id: int,
+        scope: Scope,
+    ) -> bool: ...
+
+    @overload
+    async def has_permission(
+        self, user: base_models.APIUser, resource_type: Literal[ResourceType.platform], resource_id: Any, scope: Scope
+    ) -> bool: ...
+
```
olevski, line 574: nitpick(non-blocking): I think for platform the resource_id is `str`. I tested this locally and setting this to `str` did not upset mypy. So you can change it to str across the board if you want.

components/renku_data_services/crc/api.spec.yaml
```diff
@@ -1593,6 +1800,187 @@ components:
       items:
         $ref: "#/components/schemas/PoolUserWithId"
       uniqueItems: true
+    PoolMemberUser:
+      type: object
+      additionalProperties: false
+      required:
+        - member_type
+        - id
+        - relation
```
olevski: I know this is annoying but it would be nicer to rename this to `role`. If you look at the api for project members it is `id` and `role`. We use `relation` internally because that is how authzed calls it but I think we have succeeded in not letting it bleed through the api.

components/renku_data_services/crc/api.spec.yaml
```diff
@@ -1593,6 +1800,187 @@ components:
...
+      properties:
+        member_type:
+          type: string
+          enum: [group]
+        id:
+          type: string
+          description: Group id (ULID)
+          example: "01ARZ3NDEKTSV4RRFFQ69G5FAV"
```
olevski: There is a ULID type you can copy from another spec and use.

components/renku_data_services/crc/api.spec.yaml
```diff
@@ -1593,6 +1800,187 @@ components:
...
+        member_type: group
+        id: "01ARZ3NDEKTSV4RRFFQ69G5FAV"
+        relation: group_viewer
+    PoolMemberProject:
+      type: object
+      additionalProperties: false
+      required:
+        - member_type
+        - id
+        - relation
```
olevski: Switch to `role`

components/renku_data_services/crc/api.spec.yaml
```diff
@@ -1593,6 +1800,187 @@ components:
...
+      oneOf:
+        - $ref: "#/components/schemas/PoolMemberUser"
+        - $ref: "#/components/schemas/PoolMemberGroup"
+        - $ref: "#/components/schemas/PoolMemberProject"
+      discriminator:
+        propertyName: member_type
+        mapping:
+          user: "#/components/schemas/PoolMemberUser"
+          group: "#/components/schemas/PoolMemberGroup"
+          project: "#/components/schemas/PoolMemberProject"
```
olevski, line 1873: Wow this is great. Did not know openapispec supported this.

components/renku_data_services/crc/api.spec.yaml
```diff
@@ -1593,6 +1800,187 @@ components:
...
+      items:
+        $ref: "#/components/schemas/PoolMember"
+      uniqueItems: true
+    PoolMemberUserResponse:
+      type: object
+      additionalProperties: false
+      required:
+        - member_type
+        - id
+        - relation
```
olevski: use `role` instead of `relation`.

components/renku_data_services/crc/apispec_base.py
```diff
@@ -17,13 +16,11 @@ class BaseAPISpec(BaseModel):
...
+    @field_validator("id", mode="before", check_fields=False)
     @classmethod
-    def serialize_ulid(cls, value: Any) -> Any:
-        """Handle ULIDs."""
-        if isinstance(value, ULID):
-            return str(value)
-        return value
+    def serialize_id(cls, v: ULID) -> str:
+        """Custom serializer that can handle ULIDs for id."""
+        return str(v)
```
olevski, line 41: I think the problem is that ULIDs can be found in fiels with other names other than `id`. That is why we have this.

components/renku_data_services/project/db.py
```diff
@@ -124,6 +124,25 @@ async def get_project(
...
+        async with self.session_maker() as session:
+            stmt = select(schemas.ProjectORM).where(schemas.ProjectORM.id == project_id)
+            result = await session.execute(stmt)
+            project_orm = result.scalars().first()
+
+            if project_orm is None:
+                raise errors.MissingResourceError(message=f"Project with id '{project_id}' does not exist.")
+
+            return project_orm.dump()
+
```
olevski: Why is the above `get_project` not good enough? It looks up by project id.

olevski (changes requested):

components/renku_data_services/authz/schemas.py
olevski: Actually I realize that maybe what is happening is that we merged the schema changes for v10 but did not ever do a migration. So the change in the schema code does not really matter. We never mad a migration. So if this will be the first migration then this is fine.

leafty: V9 is already applied so maybe this needs more filtering for the down migration.

components/renku_data_services/authz/authz.py
SalimKayal: done

components/renku_data_services/authz/models.py
SalimKayal: done

components/renku_data_services/authz/schemas.py
SalimKayal: fixed.

components/renku_data_services/crc/api.spec.yaml
SalimKayal: well, datamodel-code-generator did not in the version previously in the lockfile

components/renku_data_services/authz/authz.py
SalimKayal: I changed that back. throughout the code I use authzed as source of truth now. this also allows to remove the `get_projects_by_id`

components/renku_data_services/authz/authz.py
SalimKayal: done

components/renku_data_services/crc/api.spec.yaml
SalimKayal: done

components/renku_data_services/crc/api.spec.yaml
SalimKayal: done

components/renku_data_services/crc/api.spec.yaml
SalimKayal: done

components/renku_data_services/crc/api.spec.yaml
SalimKayal: done

components/renku_data_services/project/db.py
SalimKayal: it was useful when I wanted to check for user.is_admin (for the keycloak admin->authzed admin lag). removed.

components/renku_data_services/crc/apispec_base.py
SalimKayal: with `@field_validator("*", mode="before", check_fields=False)` I was getting

```
E   pydantic.errors.PydanticUserError: Cannot use a mode='before' validator in the discriminator field 'member_type' of Model 'PoolMemberUser'
E
E   For further information visit https://errors.pydantic.dev/2.11/u/discriminator-validator
```
according to the docs the wildcard * approach is fundamentally incompatible with discriminated unions.

I implemented a model_validator but this won't work on the ORM side of things.

we could be more generic by implementing a Pydantic ULID compatible type but it would change the apispec a bit

```python
def _ulid_to_str(v: Any) -> str:
    if isinstance(v, ULID):
        return str(v)
    if isinstance(v, str):
        return v
    raise ValueError("Input should be a valid ULID or ULID string")
UlidStr = Annotated[str, BeforeValidator(_ulid_to_str)]
```

what do you think ?

olevski: I think what you have with serialize_ulid is good enough. I did not realize the `*` approach does not work with discriminated unions. I will resolve this.

components/renku_data_services/authz/authz.py
SalimKayal: since `_Relation.resource_pool_platform` contains the `admin` relationship I don't think it's needed. wdyt ?

components/renku_data_services/crc/apispec_base.py
SalimKayal: > Wow this is great. Did not know openapispec supported this.

so, maybe not so great. Should I find another way ?

components/renku_data_services/authz/authz.py
olevski: Yes I agree the admins come from from resource_pool -> resource_pool_platform -> is_admin. So skipping `resource_pool_platorm` is sufficient.

olevski (approved):
Thank you @SalimKayal!!! This is awesome.

SalimKayal:
This pr breaks the admin page as shown yesterday. @olevski @leafty should we fix this before merging ?

SalimKayal:
also note : `@_only_admins` uses keycloak and not authzed, should I change this too ?

olevski:
> also note : @_only_admins uses keycloak and not authzed, should I change this too ?

Lets just leave that for now.

olevski:
> This pr breaks the admin page as shown yesterday. @olevski @leafty should we fix this before merging ?

@SalimKayal how does it break it? I am checking the CI deployment - it looks fine.

leafty:
> This pr breaks the admin page as shown yesterday. @olevski @leafty should we fix this before merging ?

No, update the admin page after merging this.

PR #1266, opened 2026-04-20
Feat: ResourcePool Authorization via Authzed/SpiceDB

SalimKayal:
## Summary:

Migrates ResourcePool access control from SQL-based checks to Authzed/SpiceDB, combining two PRs:

- **#1248** — Adds Authzed schema v9, supporting code, and an Alembic migration that both deploys the schema and backfills relations from existing DB data.
- **#1256** — Switches runtime access control to Authzed and auto-syncs DB changes to SpiceDB.

## Key Changes

- **Schema v9**: New `resource_pool` type with `viewer`, `prohibited`, `public_viewer` (wildcards), and `resource_pool_platform` relations; `read` and `write` permissions with admin bypass.
- **Migration `cd424c01676e`**: Deploys schema and backfills viewers, prohibited users, platform links, and public wildcards using `OPERATION_TOUCH` in batches of 500 (idempotent).
- **Runtime**: Replaced legacy access-control helpers with `_filter_by_authz`, which resolves permissions via Authzed and filters SQL queries by allowed IDs.
- **Sync**: `@Authz.authz_change` decorators on `ResourcePoolRepository` and `UserRepository` keep SpiceDB in sync for pool create/update/delete and member/prohibited changes.
- **DI**: `Config` and `DependencyManager` now inject `Authz` into repositories.

## Notes

- Public access via `user:*` / `anonymous_user:*` wildcards.
- `no_default_access` enforced through `prohibited` on the default (public) pool.
- All writes idempotent; migration safely re-runnable.

## Deferred

- cleanup of old DB tables
- extending authz to groups/projects.

---

/deploy renku=salimkayal-add-AUTHZ-config-to-k8s-watcher

PR #1314, opened 2026-05-19
feat: auto-grant/revoke resource pool viewer access on CRUD based on OAuth2 provider linkage

SalimKayal:
## Summary

This PR implements automatic authorization for individual users on resource pools by linking them to OAuth2 providers. When an admin creates or updates a resource pool with a `remote.provider_id`, all users who currently have an active (`connected`) OAuth2 connection to that provider are automatically granted `viewer` access. When the provider is changed or removed, access is revoked from the previously connected users and granted to the new ones.

## Motivation & Context

Resource pools already support polymorphic members (users, groups, projects) via Authzed/SpiceDB, and connected-services already supports HPC integrations (FirecREST, Run:AI) through OAuth2. However, granting access to a resource pool tied to an OAuth provider previously required manual admin intervention.

The goal is to make the OAuth2 connection the **source of truth** for resource pool access: if a user is connected to a provider, they should automatically see the resource pools linked to that provider.

### 1. No Database Migrations
We rely entirely on existing schema:
- `ResourcePoolORM.remote_provider_id` already links a resource pool to an OAuth2 client.
- `OAuth2ConnectionORM` already tracks `user_id`, `client_id`, and `status` (`connected` | `pending`).
- The Authz/SpiceDB schema (v10) already defines `relation viewer: user` on `resource_pool`.

This keeps the change minimal and non-invasive.

### 2. Direct User Authorization (No Hidden Groups)
Rather than creating synthetic/hidden groups to bridge OAuth connections and resource pools, we authorize users **directly** via the `viewer` relation in SpiceDB. This is simpler, more transparent, and avoids the complexity of managing hidden group lifecycle.

A known consequence is that manual admin `viewer` grants and auto-grants are indistinguishable in SpiceDB. Disconnecting a provider unconditionally revokes `viewer` for all its connected users, which is accepted because the OAuth connection is the source of truth.

### 3. Validation Before Mutation
Before inserting or updating a resource pool with a `provider_id`, we validate that the referenced OAuth2 client actually exists:
```python
client = await session.scalar(select(OAuth2ClientORM).where(OAuth2ClientORM.id == provider_id))
if client is None:
    raise errors.MissingResourceError(...)
```
This prevents creating dangling links to non-existent providers.

### 4. Best-Effort Membership Sync with Soft Failure
The membership sync (grant/revoke) happens **outside** the main DB transaction. This is intentional because `MemberRepository` operations open their own session and write to SpiceDB via the `@Authz.authz_change` decorator. If a bulk grant fails (e.g., a connected user is missing from Keycloak), the resource pool DB row has already been committed.

We handle this by granting/revoking users **individually** in a loop with `try/except`, logging warnings on failure:
```python
for uid in user_ids:
    member = ResourcePoolMemberIdentifier(member_type=MemberType.USER, member_id=uid)
    try:
        await self.member_repo.grant_resource_pool_members(api_user, result.id, [member])
    except errors.BaseError as e:
        logger.warning(f"Failed to auto-grant member {uid} ...")
```
This ensures one bad user record does not block access for everyone else.

### 5. Update Sync: Capture Old Provider Before Transaction
For updates, we capture `old_provider_id` **before** the transaction begins, then compare it to the `new_provider_id` after the ORM update commits. If they differ:
1. Revoke all connected users of the old provider.
2. Grant all connected users of the new provider.

This cleanly handles:
- Provider swap (A → B)
- Provider removal (A → `None` / `RESET`)
- Provider addition (`None` → B)

## Changes

- **`components/renku_data_services/crc/db.py`**
  - `ResourcePoolRepository.__init__` now accepts an optional `member_repo: MemberRepository`.
  - Added `_get_connected_user_ids(provider_id)` to query `OAuth2ConnectionORM` for users with `status == connected`.
  - `insert_resource_pool`: validates provider existence, then auto-grants connected users after successful insert.
  - `update_resource_pool`: captures old provider, validates new provider, then syncs memberships (revoke old → grant new) after the transaction.
  - `delete_resource_pool`: no changes needed — Authz cascade already cleans up all relations.

- **`bases/renku_data_services/data_api/dependencies.py`** & **`test/utils.py`**
  - Wired `member_repo` into `ResourcePoolRepository` construction.

- **`test/components/renku_data_services/db/test_sqlalchemy_pool_repo.py`**
  - Added comprehensive integration tests:
    - `test_resource_pool_insert_with_remote_grants_connected_users`
    - `test_resource_pool_update_remote_to_new_provider_swaps_access`
    - `test_resource_pool_update_remove_remote_revokes_access`
    - `test_resource_pool_delete_cleans_authz`
    - `test_resource_pool_insert_with_nonexistent_provider_id_fails`
    - `test_get_connected_user_ids_returns_only_connected_users`
    - `test_resource_pool_update_with_nonexistent_provider_id_fails`
    - `test_resource_pool_insert_auto_grant_skips_missing_keycloak_users`
    - `test_resource_pool_update_auto_grant_skips_missing_keycloak_users`
    - `test_resource_pool_update_remote_swap_skips_missing_keycloak_users`

### Stricter validation of `remote.provider_id`

`insert_resource_pool` and `update_resource_pool` now validate that the referenced `provider_id` actually exists in `oauth2_clients` **before** persisting the resource pool. If it doesn't exist, they raise `MissingResourceError`.

**Previously:** You could create or update a resource pool with any arbitrary `provider_id` (valid or not), and it would be stored as-is.  
**Now:** The request will fail with a 404-style error.

This is a contract change for the `POST /api/data/resource_pools` and `PATCH /api/data/resource_pools/{id}` endpoints (and their repository equivalents). Any client or script that was previously sending non-existent provider IDs will now get an error instead of silent acceptance.

Everything else is backward-compatible:
- `ResourcePoolRepository.__init__` adds `member_repo: MemberRepository | None = None` — optional, defaults to `None`.
- The auto-grant/revoke logic only runs when `member_repo` is provided.
- No database migrations, no Authz schema changes, no API response format changes.

---

## PR Stack

* base: #1315
  * [this](#1314)
  * #1317
  * #1327

### DIFF
--- a/components/renku_data_services/crc/db.py
+++ b/components/renku_data_services/crc/db.py
@@ -476,6 +476,8 @@ class ResourcePoolRepository(_Base):
                 session=session,
             )
 
+            # TODO figure out how to ensure we can commit AUTHZ changes in the same DB session
+
             if members is not None and self.member_repo is not None:
                 try:
                     await self.member_repo.grant_resource_pool_members(api_user, result.id, members, skip_missing=True)
--- a/components/renku_data_services/crc/db.py
+++ b/components/renku_data_services/crc/db.py
@@ -429,6 +429,17 @@ class ResourcePoolRepository(_Base):
         """Get the default resource class in the default resource pool."""
         return await self.__query_repository.get_default_resource_class()
 
+    async def filter_resource_pools(
+        self,
+        api_user: base_models.APIUser,
+        cpu: float = 0,
+        memory: int = 0,
+        max_storage: int = 0,
+        gpu: int = 0,
+    ) -> list[models.ResourcePool]:
+        """Get resource pools from database with indication of which resource class matches the specified criteria."""
+        return await self.__query_repository.filter_resource_pools(api_user, cpu, memory, max_storage, gpu)
+
     async def _get_connected_user_ids(self, provider_id: str) -> list[str]:
         """Query all users with a connected OAuth2 connection to the given provider."""
         async with self.session_maker() as session:
--- a/components/renku_data_services/crc/db.py
+++ b/components/renku_data_services/crc/db.py
@@ -474,13 +474,13 @@ class ResourcePoolRepository(_Base):
 
         if provider_id and self.member_repo is not None:
             user_ids = await self._get_connected_user_ids(provider_id)
-            if user_ids:
-                members = [ResourcePoolMemberIdentifier(member_type=MemberType.USER, member_id=uid) for uid in user_ids]
+            for uid in user_ids:
+                member = ResourcePoolMemberIdentifier(member_type=MemberType.USER, member_id=uid)
                 try:
-                    await self.member_repo.grant_resource_pool_members(api_user, result.id, members)
+                    await self.member_repo.grant_resource_pool_members(api_user, result.id, [member])
                 except errors.BaseError as e:
                     logger.warning(
-                        f"Failed to auto-grant members to resource pool {result.id} for provider {provider_id}: {e}"
+                        f"Failed to auto-grant member {uid} to resource pool {result.id} for provider {provider_id}: {e}"
                     )
 
         return result
@@ -575,14 +575,6 @@ class ResourcePoolRepository(_Base):
         self, api_user: base_models.APIUser, resource_pool_id: int, update: models.ResourcePoolPatch
     ) -> models.ResourcePool:
         """Update an existing resource pool in the database."""
-        old_provider_id: str | None = None
-        async with self.session_maker() as session:
-            rp_old = await session.scalar(
-                select(schemas.ResourcePoolORM).where(schemas.ResourcePoolORM.id == resource_pool_id)
-            )
-            if rp_old is not None:
-                old_provider_id = rp_old.remote_provider_id
-
         async with self.session_maker() as session, session.begin():
             stmt = (
                 select(schemas.ResourcePoolORM)
@@ -593,6 +585,7 @@ class ResourcePoolRepository(_Base):
             rp = res.one_or_none()
             if rp is None:
                 raise errors.MissingResourceError(message=f"Resource pool with id {resource_pool_id} cannot be found")
+            old_provider_id = rp.remote_provider_id
             quota = await self.quotas_repo.get_quota(rp.quota, rp.get_cluster_id()) if rp.quota else None
 
             validate_resource_pool_update(existing=rp.dump(quota=quota), update=update)
@@ -678,9 +671,18 @@ class ResourcePoolRepository(_Base):
                 rp.remote_provider_id = None
                 rp.remote_json = None
             elif update.remote is not None:
-                rp.remote_provider_id = (
+                new_provider_id = (
                     update.remote.provider_id if update.remote.provider_id is not None else rp.remote_provider_id
                 )
+                if new_provider_id is not None and new_provider_id != rp.remote_provider_id:
+                    from renku_data_services.connected_services.orm import OAuth2ClientORM
+
+                    client = await session.scalar(select(OAuth2ClientORM).where(OAuth2ClientORM.id == new_provider_id))
+                    if client is None:
+                        raise errors.MissingResourceError(
+                            message=f"OAuth2 Client with id '{new_provider_id}' does not exist."
+                        )
+                rp.remote_provider_id = new_provider_id
                 remote_json = rp.remote_json if rp.remote_json is not None else dict()
                 remote_json.update(update.remote.to_dict())
                 del remote_json["provider_id"]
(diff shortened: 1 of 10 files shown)


components/renku_data_services/crc/db.py
```diff
@@ -436,18 +440,52 @@ async def filter_resource_pools(
...
         """Insert resource pool into database."""
+        provider_id = None
+        if new_resource_pool.remote and new_resource_pool.remote.provider_id:
+            provider_id = new_resource_pool.remote.provider_id
+            async with self.session_maker() as session:
+                from renku_data_services.connected_services.orm import OAuth2ClientORM
+
+                client = await session.scalar(select(OAuth2ClientORM).where(OAuth2ClientORM.id == provider_id))
+                if client is None:
+                    raise errors.MissingResourceError(message=f"OAuth2 Client with id '{provider_id}' does not exist.")
```
leafty: Why do we add this check here? The foreign key constraint already rejects the update. Maybe the code in `_insert_resource_pool()` can be updated to provide a better error message (currently an HTTP 500 from the database integrity error).

Adding the check in `_insert_resource_pool()` will avoid running another DB transaction.

components/renku_data_services/crc/db.py
```diff
@@ -436,18 +440,52 @@ async def filter_resource_pools(
...
+                member = ResourcePoolMemberIdentifier(member_type=MemberType.USER, member_id=uid)
+                try:
+                    await self.member_repo.grant_resource_pool_members(api_user, result.id, [member])
+                except errors.BaseError as e:
+                    logger.warning(
+                        f"Failed to auto-grant member {uid} to resource pool {result.id} "
+                        f"for provider {provider_id}: {e}"
+                    )
+
+        return result
```
leafty: Can we do this as a single authz operation?

components/renku_data_services/crc/db.py
SalimKayal: I kept the check in `insert_resource_pool` to keep the separation of concerns. I refactored to only use one session there. though `grant_resource_pool_members` will use one for itself because of the `session.commit()` in `authz.py`

components/renku_data_services/crc/db.py
SalimKayal: done

SalimKayal: almost, by doing this as a single authz operation they either succeed together or fail together. I will implement a way of having missing members fail gracefully without making the  whole transaction fail.

SalimKayal: done

components/renku_data_services/crc/db.py
```diff
@@ -436,18 +440,47 @@ async def filter_resource_pools(
         """Get resource pools from database with indication of which resource class matches the specified criteria."""
         return await self.__query_repository.filter_resource_pools(api_user, cpu, memory, max_storage, gpu)
 
+    async def _get_connected_user_ids(self, provider_id: str) -> list[str]:
```
leafty: Remark: would this use less memory with streaming results? It's OK as it this.

components/renku_data_services/crc/db.py
```diff
@@ -436,18 +440,47 @@ async def filter_resource_pools(
...
     ) -> models.ResourcePool:
         """Insert resource pool into database."""
         async with self.session_maker() as session, session.begin():
-            return await self._insert_resource_pool(
+            provider_id = None
+            if new_resource_pool.remote and new_resource_pool.remote.provider_id:
+                provider_id = new_resource_pool.remote.provider_id
+                client = await session.scalar(select(OAuth2ClientORM).where(OAuth2ClientORM.id == provider_id))
+                if client is None:
+                    raise errors.MissingResourceError(message=f"OAuth2 Client with id '{provider_id}' does not exist.")
```
leafty: May be a nitpick: I think I would use a `errors.ValidationError()` in this case: the field `remote.provider_id` has an incorrect value.

components/renku_data_services/crc/db.py
leafty: If I understand correctly: Inserting the resource pool into the database is done in its own operation, so that it can be successful on its own and auto-membership changes failing would still keep the new resource pool.

I think I would like the auto-membership changes to be done in one DB+Authz transaction if possible.

If I read the code correctly:
1. A db transaction is open and closed to get the user_Id list
2. A db transaction+authz transaction is open and closed to add the new members.

If not done in this PR, maybe we should have a TODO comment and issue to make sure we have the code from L475 to L487 done in one transaction.

components/renku_data_services/crc/db.py
```diff
@@ -646,12 +687,45 @@ async def update_resource_pool(
...
+                    await self.member_repo.grant_resource_pool_members(
+                        api_user, resource_pool_id, members, skip_missing=True
+                    )
+                except errors.BaseError as e:
+                    logger.warning(
+                        f"Failed to auto-grant members to resource pool {resource_pool_id} "
+                        f"for provider {new_provider_id}: {e}"
+                    )
+
+        return transaction_result
```
leafty, line 744: Same here, we have up to 3 db transactions.

components/renku_data_services/crc/db.py
```diff
@@ -1271,28 +1345,30 @@ async def _resolve_members(
...
+        """Resolve ResourcePoolMemberIdentifiers to (internal_id, subject_type, role, member_type) tuples."""
+        resolved: list[tuple[str, ResourceType, str, MemberType]] = []
         for member in members:
             match member.member_type:
                 case MemberType.USER:
                     kc_user = await self.kc_user_repo.get_user(id=member.member_id)
                     if kc_user is None:
+                        if skip_missing:
+                            logger.warning(f"Skipping auto-grant for missing user {member.member_id}")
+                            continue
```
leafty, line 1417: Why not `if not skip_missing: raise`?

Or:
```python
if skip_missing:
    logger.warning(...)
else:
    raise ...
```
To avoid having a `continue`.

components/renku_data_services/crc/db.py
```diff
@@ -1365,10 +1443,13 @@ async def grant_resource_pool_members(
         resource_pool_id: int,
         members: Collection[ResourcePoolMemberIdentifier],
         append: bool = True,
+        skip_missing: bool = False,
     ) -> None:
         """Grant access to a resource pool for a collection of members (Users, Groups, Projects)."""
         async with self.session_maker() as session, session.begin():
-            await self._grant_resource_pool_members(api_user, resource_pool_id, members, session=session)
+            await self._grant_resource_pool_members(
```
leafty, line 1508: Note: reading the method definition from `_grant_resource_pool_members()` got me very confused: https://github.com/SwissDataScienceCenter/renku-data-services/pull/1266/changes#r3349332532

Reading:
```python
    @with_db_transaction
    @Authz.authz_change(op=AuthzOperation.create, resource=ResourceType.resource_pool)
    async def _grant_resource_pool_members(...)
```
This feels like a resource pool is being created which is not the case. See comment from above: maybe this should have been a `AuthzOperation.update_membership` operation to make clear that the method will update the resource pool relations in authz and not that a new resource pool is created in authz.

leafty (commented):
Note: sorry if I didn't see that this is the first PR in the stack earlier.

components/renku_data_services/crc/db.py
leafty: Nevermind.

leafty (commented):
Note: @SalimKayal should be OK if the addressable comments are fixed.

components/renku_data_services/crc/db.py
SalimKayal: fixed in 78a41bc

components/renku_data_services/crc/db.py
SalimKayal: Removed the ones I could remove

components/renku_data_services/crc/db.py
SalimKayal: done

components/renku_data_services/crc/db.py
SalimKayal: The problem is still the `session.commit()` here

https://github.com/SwissDataScienceCenter/renku-data-services/blob/4425e29a68159cf0eb7bc3878557011d7db0ed8b/components/renku_data_services/authz/authz.py#L1017

Let me add the TODO

leafty (approved):

PR #1316, opened 2026-05-20
feat: add option to remove iframe from sessions

olevski:
For remote K8s clusters getting the iframes of sessions to work is fairly complicated dur to CORS/CSP. This is because the sesssion is running on one domain (for example dev.renku.ch) whereas the UI is running on another - like renkulab.io. For clusters we do not have control and are limited it is hard to properly configure these options. Also in some cases the configuration is required at the Jupyterlab or RStudio level which makes it hard to generalize. This is just a quick relief for currently active remote clusters in production being used right now where a lot of users are having trouble and really bad UX because the default button to open a session leads to a 50x page.

We can think and work on a long term solution in the future so that we still have iframes or the same UI for remote sessions.

/deploy


leafty (commented):
General comments:
1. Why don't we tunnel the session http port?
2. What if we only change the UI? Remote cluster -> open in new tab, "Open in iframe" as a secondary action.

olevski:
> Why don't we tunnel the session http port?

I wanted to avoid this if possible. Not having tunneling is definitely simpler.

> What if we only change the UI? Remote cluster -> open in new tab, "Open in iframe" as a secondary action.

Yes this is an option but then all remote clusters do not have this problem - just some. And the iframe brings a lot of useful information. So with this one small change on the backend and the changes we would do on the frontend anyhow we can apply this only to clusters that have problems with iframes.

PR #1317, opened 2026-05-20
feat: sync resource pool viewer access on OAuth2 connect and disconnect

SalimKayal:
## Summary

This PR implements the second half of automatic resource pool authorization: when a user successfully completes an OAuth2 authorization flow (connects to a provider), they are automatically granted `viewer` access to all resource pools whose `remote.provider_id` matches that OAuth client. Conversely, when a user deletes their OAuth2 connection, that access is revoked.

This complements #1314  to make the OAuth2 connection the **source of truth** for resource pool access.

## Motivation & Context

Resource pools can be linked to OAuth2 providers via `remote.provider_id`. In Chunk 1, we ensured that when an admin creates or updates such a resource pool, all currently-connected users are automatically granted access. However, if a user connects to the provider *after* the resource pool already exists, they would not receive access until the resource pool is modified.

This PR closes that gap by hooking into the OAuth2 lifecycle events:
- **Connect callback** (`GET /api/data/oauth2/callback`) — triggered after the token exchange succeeds.
- **Disconnect** (`DELETE /api/data/oauth2/connections/{id}`) — triggered when the user deletes their connection.

### 1. Event-Driven Sync at OAuth Lifecycle Boundaries
We hook into two existing flows rather than adding new endpoints:

| Event | Location | Action |
|-------|----------|--------|
| User connects to provider P | `connected_services/blueprints.py::authorize_callback` | Call `_on_oauth2_connected(user_id, provider_id)` |
| User disconnects from provider P | `connected_services/db.py::delete_oauth2_connection` | Call `_on_oauth2_disconnected(user_id, provider_id)` |

This keeps the change localized and avoids polling or background jobs.

### 2. Best-Effort Sync — OAuth Flow Must Not Fail
Both the connect callback and disconnect operation are wrapped in `try/except` with warning logs:

```python
try:
    await self.connected_services_repo._on_oauth2_connected(user_id, provider_id)
except Exception as e:
    logger.warning(f"Failed to sync resource pool memberships after OAuth2 connection ...")
```

**Rationale:** The OAuth2 connection itself is the primary operation. If SpiceDB is temporarily unavailable or a user record is inconsistent, we still want the user to successfully connect to the provider. The RP access can be retried or fixed later by reconnecting.

### 3. Disconnect Double-Call Pattern (Transactional + Best-Effort)
In `delete_oauth2_connection`, `_on_oauth2_disconnected` is called **twice**:

1. **Before `session.delete(conn)`** — inside the transaction, so that if revocation fails due to a transient error, the transaction rolls back and the connection is preserved (allowing retry).
2. **After `session.flush()`** — in a separate `try/except`, as a best-effort cleanup in case the first call succeeded but we want to ensure the connection is deleted even if revocation fails.

This ensures we do not leave orphaned connections when revocation fails, while still attempting to keep DB and Authz state consistent.

### 4. Cross-Component Dependency Injection
`ConnectedServicesRepository` now accepts an optional `member_repo: MemberRepository | None`. This is a cross-component import (`connected_services` → `crc`), which is already practiced elsewhere (e.g., `data_api/dependencies.py` constructs both repositories). Using `TYPE_CHECKING` avoids circular import issues at runtime.

### 5. Exposing `user_id` and `client_id` on `OAuth2Connection`
The `OAuth2Connection` model and `OAuth2ConnectionORM.dump()` now include `user_id` and `client_id` fields. Previously, only `provider_id` (which is the same as `client_id`) was exposed. This allows the `authorize_callback` blueprint to access the connecting user's ID and the provider ID directly from the `fetch_token` result without re-querying the database.

## Changes

- **`components/renku_data_services/connected_services/db.py`**
  - `ConnectedServicesRepository.__init__` now accepts an optional `member_repo: MemberRepository`.
  - Added `_on_oauth2_connected(user_id, client_id)`: queries all RPs linked to the provider and grants the user as `viewer` on each.
  - Added `_on_oauth2_disconnected(user_id, client_id)`: queries all RPs linked to the provider and revokes the user's `viewer` access from each.
  - Both helpers use `InternalServiceAdmin()` for authz and loop through matching RPs with per-RP `try/except`.
  - `delete_oauth2_connection`: calls `_on_oauth2_disconnected` before deletion and again after flush as best-effort.

- **`components/renku_data_services/connected_services/blueprints.py`**
  - `authorize_callback`: after successful `fetch_token`, extracts `user_id` and `provider_id` from `client.connection` and calls `_on_oauth2_connected` with failure isolation.

- **`components/renku_data_services/connected_services/models.py`**
  - `OAuth2Connection` dataclass added `user_id: str | None = None` and `client_id: str | None = None`.

- **`components/renku_data_services/connected_services/orm.py`**
  - `OAuth2ConnectionORM.dump()` now includes `user_id` and `client_id` in the returned model.

- **`bases/renku_data_services/data_api/dependencies.py`** & **`test/utils.py`**
  - Wired `member_repo` into `ConnectedServicesRepository` construction.

- **`test/components/renku_data_services/connected_services/test_db.py`**
  - Added comprehensive integration tests:
    - `test_oauth_connect_adds_user_to_rp`
    - `test_oauth_disconnect_removes_user_from_rp`
    - `test_oauth_connect_with_no_matching_rp_does_nothing`
    - `test_two_users_connect_same_integration_both_get_access`
    - `test_user_disconnect_only_affects_their_rp_access`
    - `test_delete_connection_revokes_rp_access`

### OAuth2 Callback Now Triggers Resource Pool Access Grants

**Previously:** After a user completed the OAuth2 callback (`/api/data/oauth2/callback`), they were connected to the provider but received no automatic resource pool access. Admins had to manually grant `viewer` on each relevant resource pool.

**Now:** The callback automatically grants `viewer` access to all resource pools linked to that provider. The user will see these resource pools immediately on their next `GET /api/data/resource_pools` request.

### OAuth2 Disconnect Now Triggers Resource Pool Access Revocation

**Previously:** Deleting an OAuth2 connection only removed the connection row. Any resource pool `viewer` grants previously received via that connection remained in SpiceDB.

**Now:** Disconnecting unconditionally revokes `viewer` access from all resource pools linked to that provider. This is accepted because the OAuth connection is the source of truth (see Chunk 1 rationale).

### `OAuth2Connection` Model Now Exposes `user_id` and `client_id`

**Previously:** `OAuth2Connection` only exposed `id`, `provider_id`, `status`, and `next_url`.

**Now:** It also exposes `user_id` and `client_id`. This is a additive change — existing consumers are unaffected. It enables the callback blueprint to pass the correct IDs to `_on_oauth2_connected` without additional DB queries.

### Resilience: OAuth Flow Succeeds Even If RP Sync Fails

If SpiceDB or the membership repository is unavailable during the OAuth callback, the user will still successfully connect to the provider. A warning is logged and the RP access can be re-synced by reconnecting. This is a deliberate soft-failure strategy to avoid making the OAuth flow dependent on the authz subsystem.

Everything else is backward-compatible:
- `ConnectedServicesRepository.__init__` adds `member_repo: MemberRepository | None = None` — optional, defaults to `None`.
- The auto-grant/revoke logic only runs when `member_repo` is provided.
- No database migrations, no Authz schema changes, no API response format changes.

---

## PR Stack

* base: #1315
  * #1314
  * [this](#1317)
  * #1327

### DIFF
--- a/components/renku_data_services/connected_services/db.py
+++ b/components/renku_data_services/connected_services/db.py
@@ -46,7 +46,7 @@ class ConnectedServicesRepository:
         self.supported_image_registry_providers = {models.ProviderKind.gitlab, models.ProviderKind.github}
         self.member_repo = member_repo
 
-    async def _on_oauth2_connected(self, user_id: str, client_id: str) -> None:
+    async def on_oauth2_connected(self, user_id: str, client_id: str) -> None:
         """Grant user viewer access to all RPs linked to the provider."""
         if self.member_repo is None:
             return
@@ -69,7 +69,7 @@ class ConnectedServicesRepository:
                         f"Failed to auto-grant member {user_id} to resource pool {rp.id} for provider {client_id}: {e}"
                     )
 
-    async def _on_oauth2_disconnected(self, user_id: str, client_id: str) -> None:
+    async def on_oauth2_disconnected(self, user_id: str, client_id: str) -> None:
         """Revoke user viewer access from all RPs linked to the provider."""
         if self.member_repo is None:
             return
@@ -243,14 +243,14 @@ class ConnectedServicesRepository:
             if conn is None:
                 return False
 
-            await self._on_oauth2_disconnected(conn.user_id, conn.client_id)
+            await self.on_oauth2_disconnected(conn.user_id, conn.client_id)
             await session.delete(conn)
             await session.flush()
 
-            # Ensure session is closed before calling _on_oauth2_disconnected
+            # Ensure session is closed before calling on_oauth2_disconnected
             # so that the connection is deleted even if revoke fails.
             try:
-                await self._on_oauth2_disconnected(conn.user_id, conn.client_id)
+                await self.on_oauth2_disconnected(conn.user_id, conn.client_id)
             except Exception as e:
                 logger.warning(
                     f"Failed to sync resource pool memberships after OAuth2 disconnection for user "
--- a/components/renku_data_services/connected_services/db.py
+++ b/components/renku_data_services/connected_services/db.py
@@ -251,6 +251,7 @@ class ConnectedServicesRepository:
             if conn is None:
                 return False
 
+            await self._on_oauth2_disconnected(conn.user_id, conn.client_id)
             await session.delete(conn)
             return True
 
--- a/components/renku_data_services/connected_services/db.py
+++ b/components/renku_data_services/connected_services/db.py
@@ -68,14 +68,6 @@ class ConnectedServicesRepository:
                     logger.warning(
                         f"Failed to auto-grant member {user_id} to resource pool {rp.id} for provider {client_id}: {e}"
                     )
-            member = ResourcePoolMemberIdentifier(member_type=MemberType.USER, member_id=user_id)
-            for rp in rps:
-                try:
-                    await self.member_repo.grant_resource_pool_members(admin, rp.id, [member])
-                except errors.BaseError as e:
-                    logger.warning(
-                        f"Failed to auto-grant member {user_id} to resource pool {rp.id} for provider {client_id}: {e}"
-                    )
 
     async def _on_oauth2_disconnected(self, user_id: str, client_id: str) -> None:
         """Revoke user viewer access from all RPs linked to the provider."""
@@ -253,6 +245,18 @@ class ConnectedServicesRepository:
 
             await self._on_oauth2_disconnected(conn.user_id, conn.client_id)
             await session.delete(conn)
+            await session.flush()
+
+            # Ensure session is closed before calling _on_oauth2_disconnected
+            # so that the connection is deleted even if revoke fails.
+            try:
+                await self._on_oauth2_disconnected(conn.user_id, conn.client_id)
+            except Exception as e:
+                logger.warning(
+                    f"Failed to sync resource pool memberships after OAuth2 disconnection for user "
+                    f"{conn.user_id} and provider {conn.client_id}: {e}"
+                )
+
             return True
 
     async def get_oauth2_connections(
--- a/components/renku_data_services/connected_services/db.py
+++ b/components/renku_data_services/connected_services/db.py
@@ -1,6 +1,9 @@
 """Adapters for connected services database classes."""
 
+from __future__ import annotations
+
 from collections.abc import Callable
+from typing import TYPE_CHECKING
 from urllib.parse import urlparse
 
 from sqlalchemy import and_, select
@@ -21,6 +24,9 @@ from renku_data_services.notebooks.api.classes.image import Image, ImageRepoDock
 from renku_data_services.users.db import APIUser
 from renku_data_services.utils.cryptography import encrypt_string
 
+if TYPE_CHECKING:
+    from renku_data_services.crc.db import MemberRepository
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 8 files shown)


components/renku_data_services/connected_services/db.py
```diff
@@ -32,11 +38,59 @@ def __init__(
...
         encryption_key: bytes,
+        member_repo: MemberRepository | None = None,
     ):
         self.session_maker = session_maker
         self.encryption_key = encryption_key
         self.oauth_client_factory = oauth_client_factory
         self.supported_image_registry_providers = {models.ProviderKind.gitlab, models.ProviderKind.github}
+        self.member_repo = member_repo
+
+    async def _on_oauth2_connected(self, user_id: str, client_id: str) -> None:
```
leafty: Q: why is this a hidden method when it's called from another class?

components/renku_data_services/connected_services/db.py
```diff
@@ -32,11 +38,59 @@ def __init__(
         session_maker: Callable[..., AsyncSession],
         oauth_client_factory: OAuthHttpClientFactory,
         encryption_key: bytes,
+        member_repo: MemberRepository | None = None,
```
leafty: Why do we allow `None`? It seems that this is always setup with the `MemberRepository`.

components/renku_data_services/connected_services/db.py
```diff
@@ -32,11 +38,59 @@ def __init__(
...
         self.supported_image_registry_providers = {models.ProviderKind.gitlab, models.ProviderKind.github}
+        self.member_repo = member_repo
+
+    async def _on_oauth2_connected(self, user_id: str, client_id: str) -> None:
+        """Grant user viewer access to all RPs linked to the provider."""
+        if self.member_repo is None:
+            return
+        from renku_data_services.base_models.core import InternalServiceAdmin
+        from renku_data_services.crc import orm as crc_schemas
+        from renku_data_services.crc.models import MemberType, ResourcePoolMemberIdentifier
```
leafty: Not sure why we do not have these imports at the top of the file.

components/renku_data_services/connected_services/db.py
```diff
@@ -32,11 +38,59 @@ def __init__(
...
+            )
+            rps_list = list(rps)
+            member = ResourcePoolMemberIdentifier(member_type=MemberType.USER, member_id=user_id)
+            for rp in rps_list:
+                try:
+                    await self.member_repo.grant_resource_pool_members(admin, rp.id, [member])
+                except errors.BaseError as e:
+                    logger.warning(
+                        f"Failed to auto-grant member {user_id} to resource pool {rp.id} for provider {client_id}: {e}"
+                    )
```
leafty: Do we have a single operation for this instead? In the API there seems to be an operation to add a user to multiple resource pools at once -> https://dev.renku.ch/swagger/#/users/post_users__user_id__resource_pools.

It would be nice if the operation here is done as a single authz operation (it should be possible to do it),

components/renku_data_services/connected_services/db.py
```diff
@@ -189,7 +243,20 @@ async def delete_oauth2_connection(self, user: base_models.APIUser, connection_i
             if conn is None:
                 return False
 
+            await self._on_oauth2_disconnected(conn.user_id, conn.client_id)
```
leafty: As `session_maker()` is not re-entrant, this call will create a new SQLAlchemy session. Pass the current session and transaction to `_on_oauth2_disconnected()` so that it uses the same database connection.

components/renku_data_services/connected_services/db.py
SalimKayal: fixed

components/renku_data_services/connected_services/db.py
SalimKayal: I reordered the `DependencyManager` setup `from_env` in bases/renku_data_services/data_api/dependencies.py

we don't need the `| None`

components/renku_data_services/connected_services/db.py
SalimKayal: fixed

components/renku_data_services/connected_services/db.py
SalimKayal: a bit of a deeper refactor for this one but it's done

components/renku_data_services/connected_services/db.py
SalimKayal: fixed

components/renku_data_services/connected_services/db.py
```diff
@@ -32,11 +41,51 @@ def __init__(
...
     ):
         self.session_maker = session_maker
         self.encryption_key = encryption_key
         self.oauth_client_factory = oauth_client_factory
         self.supported_image_registry_providers = {models.ProviderKind.gitlab, models.ProviderKind.github}
+        self.member_repo = member_repo
+
+    async def _get_rp_ids_for_provider(self, client_id: str, session: AsyncSession) -> list[int]:
+        """Get all resource pool IDs linked to a given OAuth2 provider."""
+        async with self.session_maker() as session:
```
leafty: A new session is still made here 😅 

```suggestion
```

components/renku_data_services/connected_services/db.py
SalimKayal: INFO: it can't be done in a single operation due to the `session.commit()` in the `authz.py` file

components/renku_data_services/connected_services/db.py
SalimKayal: well caught, thanks

leafty (approved):
👏

PR #1318, opened 2026-05-21
fix: assume no label is interactive in list_sessions()

leafty:
Tested: start sessions and remove the "session_type" label, check that the session is still usable from the UI.

/deploy

### DIFF
--- a/components/renku_data_services/notebooks/api/classes/k8s_client.py
+++ b/components/renku_data_services/notebooks/api/classes/k8s_client.py
@@ -190,7 +190,9 @@ class NotebookK8sClient(SecretClient):
         session_filter = {self.__username_label: safe_username}
         match session_type:
             case SessionType.non_interactive:
-               session_filter.update({"renku.io/session-type": SessionType.non_interactive.value}) 
+                session_filter.update({"renku.io/session-type": SessionType.non_interactive.value})
+            case _:
+                pass
         sessions = [
             self.__session_type.model_validate(s.manifest)
             async for s in self.__client.list(
@@ -201,10 +203,13 @@ class NotebookK8sClient(SecretClient):
                 )
             )
         ]
-        def _filter_by_label(s: AmaltheaSessionV1Alpha1) -> bool:
-            session_type_label : str= s.metadata.labels.get("renku.io/session-type")
-            return not session_type_label or session_type_label.lower() == session_type.value.lower()
-        sessions = filter(_filter_by_label, sessions)
+        if session_type:
+
+            def _filter_by_label(s: AmaltheaSessionV1Alpha1) -> bool:
+                session_type_label: str | None = s.metadata.labels.get("renku.io/session-type")
+                return not session_type_label or session_type_label.lower() == session_type.value.lower()
+
+            sessions = list(filter(_filter_by_label, sessions))
         return sorted(sessions, key=lambda sess: sess.metadata.name)
 
     async def get_session(self, name: str, safe_username: str) -> AmaltheaSessionV1Alpha1 | None:
--- a/components/renku_data_services/notebooks/api/classes/k8s_client.py
+++ b/components/renku_data_services/notebooks/api/classes/k8s_client.py
@@ -188,8 +188,9 @@ class NotebookK8sClient(SecretClient):
     ) -> list[AmaltheaSessionV1Alpha1]:
         """Get a list of sessions that belong to a user."""
         session_filter = {self.__username_label: safe_username}
-        if session_type is not None:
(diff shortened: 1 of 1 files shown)


components/renku_data_services/notebooks/api/classes/k8s_client.py
```diff
@@ -188,8 +188,11 @@ async def list_sessions(
     ) -> list[AmaltheaSessionV1Alpha1]:
         """Get a list of sessions that belong to a user."""
         session_filter = {self.__username_label: safe_username}
-        if session_type is not None:
-            session_filter.update({"renku.io/session-type": str(session_type)})
+        match session_type:
```
eikek, line 191: I'm wondering if we could use `renku.io/session-type != non_interactive` to select interactive sessions. Reading [this](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors) it seems to select all objects that are not `non_interactive` and all without that label. But I'm not sure how to apply that with the python api we have.

eikek (approved):

components/renku_data_services/notebooks/api/classes/k8s_client.py
```diff
@@ -200,6 +203,13 @@ async def list_sessions(
                 )
             )
         ]
+        if session_type:
+
+            def _filter_by_label(s: AmaltheaSessionV1Alpha1) -> bool:
+                session_type_label: str | None = s.metadata.labels.get("renku.io/session-type")
+                return not session_type_label or session_type_label.lower() == session_type.value.lower()
+
+            sessions = list(filter(_filter_by_label, sessions))
```
sgaist, line 212: From what I understood:
- if `session_type` == None => return all sessions
- if `session_type` == `non_interactive` => double filtering happening
- if `session_type` == `interactive` => filter on all sessions

From the looks of it, `if session_type:` on line 206  could be `if session_type != SessionType.non_interactive.value:` as it's the only time the filtering needs to be done.

Otherwise, we could also have the "reverse match" to be consistent with the block above:
```suggestion
        match session_type:
            case SessionType.non_interactive:
                pass
            case _:
                def _filter_by_label(s: AmaltheaSessionV1Alpha1) -> bool:
                    session_type_label: str | None = s.metadata.labels.get("renku.io/session-type")
                    return not session_type_label or session_type_label.lower() == session_type.value.lower()

                sessions = list(filter(_filter_by_label, sessions))
```

sgaist (dismissed):

components/renku_data_services/notebooks/api/classes/k8s_client.py
leafty: > I'm wondering if we could use renku.io/session-type != non_interactive to select interactive sessions.

~~I don't think this is available from the k8s API.~~

components/renku_data_services/notebooks/api/classes/k8s_client.py
leafty: This suggestion doesn't work: `case _` would also apply when session_type is None.

leafty: Also note that double filtering may be fixed later, the issue being fixed is happening in production.

olevski: I think this is fine as is. If we wanted to adopt the suggestion you could add a specific `None` case and just do nothing in it.

olevski (approved):

components/renku_data_services/notebooks/api/classes/k8s_client.py
leafty: Once again, we wrote "simplified" k8s API objects. We could do this later, maybe.

olevski: Oh k8s can actually filter labels based on `!=`?

leafty: Yes, you can do much more (see link from @eikek). The issue is that the selector needs to be serialized as a string (e.g. `label != value, label notin (v1, v2), !other_label`).

Kr8s supports this but also exposes a "dict" API which serializes to just `{k1}={v1},{k2}={v2},...`.

Maybe a more complex API based on operators should also exist?!? 😅 
```python
def dict_to_selector(selector_dict: dict) -> str:
    """Convert a dictionary to a Kubernetes selector.

    Args:
        selector_dict: The dictionary to convert to a Kubernetes selector.

    Returns:
        A Kubernetes selector string.
    """
    return ",".join(f"{k}={v}" for k, v in selector_dict.items())
```

PR #1320, opened 2026-05-21
fix: update session labels upon patch

leafty:
Tested in the deployment: the session type label is added with `interactive` if not present.

/deploy


components/renku_data_services/notebooks/api/classes/k8s_client.py
leafty, line 1: Note: updating this file to cleanup the code and also support more type values (e.g. `app`).

olevski (approved):
This looks good to me Flora. Let me know if you want me to re-approve or review again when there are more changes.

leafty:
> This looks good to me Flora. Let me know if you want me to re-approve or review again when there are more changes.

It seems to work with jobs and sessions in the deployment.

PR #1324, opened 2026-05-21
fix: simplify validation for valid container image

olevski:
This had a really long complicated regex. This regex did not include images on a repository that has a custom port. Like `repo.com:5000/prj/img`. Instead of changing the already involved regex I removed it fully. Now we do a lot of checks for image existence on the frontend and backend so users would know quickly the image is not accessible.

/deploy


leafty (dismissed):

leafty (commented):
Note: need to also update the other API spec where the same schema is used.

Also, if any regex validation or regex parsing is done, we should use the reference ones -> https://github.com/distribution/reference/blob/main/regexp.go.

olevski:
> Note: need to also update the other API spec where the same schema is used.

By this you mean other places in the apispecs in data services where we use the regex? I did not realize there was one more in projects to handle v1->v2 project migrations. Which I removed also. So that is it.

leafty (dismissed):

olevski:
Sorry @leafty I accidentally pushed the wrong commit here (that should have gone to another branch). I removed this commit but your approval got dismissed.

leafty (approved):

PR #1310, opened 2026-05-18
fix: finding data connectors by slug

olevski:
/deploy

fixes #1309

PR #1326, opened 2026-05-21
fix: filtering for data connectors by slugs

### LINKED ISSUE #1325: Cannot list DCs in project from the user or group page
> This breaks listing data connectors from a user or group namespace! 

 _Originally posted by @leafty in [#1310](https://github.com/SwissDataScienceCenter/renku-data-services/pull/1310/changes#r3281637974)_

Run `cypress-tests/cypress/e2e/v2/dataConnectors.cy.ts` from the renku repository to reproduce the error.

Alternatively:
1. Create a project
2. Create a DC in the project (owned by the project)
3. Go the the project's owner page (user or group) and observe that the DC does not appear in the "Data" box. Check also the browser's network tab for the API call (e.g. `GET https://dev.renku.ch/api/data/data_connectors?namespace=flora.thiebaut&page=1&per_page=12`).

cc @olevski

olevski:
Follow up from #1310 which limited the filtering too much. Because of this we were not showing some nested data connectors in some cases - i.e. connectors in the project when filtering only the user namespace.

What we were doing:
- filter connectors for userA
- showing data connectors only in the user namespace

What was expected:
- filter connectors for userA
- show connectors in the user namespace
- show connectors in every project in the user namespce <--- this was missing

/deploy


components/renku_data_services/data_connectors/db.py
```diff
@@ -1161,8 +1156,16 @@ async def delete_data_connector_secrets(self, user: base_models.APIUser, data_co
 _T = TypeVar("_T", int, schemas.DataConnectorORM)
 
 
-def _filter_by_namespace_slug(stmt: Select[tuple[_T]], namespace: ProjectPath | NamespacePath) -> Select[tuple[_T]]:
-    """Filters a select query on data connectors to a given namespace."""
+def _filter_by_namespace_slug(
```
leafty, line 1159: Since the two filters are orthogonal, it makes sense to me that there should be two methods:

```python
def _filter_by_namespace(stmt, namespace: ProjectPath | NamespacePath):
    """Filters a DC query on the namespace which is a user, a group or a project."""
    pass

def _filter_by_slug(stmt, slug: DataConnectorSlug):
    """Filters a DC query on the slug."""
    pass
```

leafty (approved):
Looks OK, but I think it would be simpler to have two methods 👇 .

PR #1328, opened 2026-05-22
feat: Allow multiple jobs per job launcher

eikek:
/deploy extra-values=dataService.imageBuilders.enabled=true,dataService.imageBuilders.strategyName=renku-buildpacks-v3,dataService.imageBuilders.outputImagePrefix=harbor.dev.renku.ch/renku-build/,dataService.imageBuilders.nodeSelector.renku.io/node-purpose=user,dataService.imageBuilders.tolerations[0].effect=NoSchedule,dataService.imageBuilders.tolerations[0].key=renku.io/dedicated,dataService.imageBuilders.tolerations[0].operator=Equal,dataService.imageBuilders.tolerations[0].value=user amalthea-sessions=fix/job-status-when-pod-missing


components/renku_data_services/notebooks/api.spec.yaml
```diff
@@ -233,10 +233,23 @@ components:
         resource_class_id:
           nullable: true
           type: integer
+        submission_id:
+          nullable: true
+          type: string
+          pattern: "^[0-9A-Za-z]{4,}$"
```
olevski: issue: This will go straight into k8s as part of the name for the pod/session. In k8s pod names cannot have uppercase letters - so we can actually make that clear here. If we accept both and we make the uppercase letters to lowercase we may end up with collisions later on.

If you want you can also allow `-`. But note that the starting character for a name cannot be a `-` but numbers or letters are allowed as the start.

olevski: I think what you can do is `"^[a-z0-9][a-z0-9-]{3,19}$"`

Also there is a max limit to the length. Since we are merging this with other things I would cut this at ~20 characters. I know we truncate this in the code so the length is not a problem so you can increase it if you want. But I would still not leave it unbounded.

components/renku_data_services/notebooks/util/kubernetes_.py
```diff
@@ -68,6 +75,8 @@ def _make_server_name_prefix(safe_username: str) -> str:
         prefix = "n"
 
     prefix = f"{prefix}{safe_username}"
+    if submission_id:
+        prefix = f"{prefix}_{submission_id}"
```
olevski: The prefix goes directly into the name of the pod/sessions. `_` is not allowed in kubernetes names. But `-` are allowed
```suggestion
    if submission_id:
        prefix = f"j-{prefix}-{submission_id}"
```

Also we should prefix it by `j-` or something to make all the jobs show up together. Just a quality of life improvement for admins. Right now the  sessions do not have this and they show up all over the place when listing pods. I.e. they show up interspersed with the renku session pods. And I used a single letter so we dont take up more of the meaningful stuff. And `j` beacuse of `job`.

components/renku_data_services/notebooks/core_sessions.py
```diff
@@ -987,6 +999,8 @@ async def start_session(
         "renku.io/resource_class_id": str(resource_class.id),
         "renku.io/resource_pool_id": str(resource_pool.id),
     }
+    if launch_request.submission_id:
+        annotations.update({"renku.io/submission_id": str(launch_request.submission_id)})
```
olevski, line 1006: question: I wonder if this should be in labels because you can filter when listing by labels but not by annotations.

components/renku_data_services/session/apispec_base.py
```diff
@@ -54,3 +54,14 @@ def convert_path_to_string(cls, val: str | PurePosixPath) -> str:
         if isinstance(val, PurePosixPath):
             return val.as_posix()
         return val
+
+    @field_validator("launcher_type", mode="before", check_fields=False)
+    @classmethod
+    def serialize_launcher_type(cls, launcher_type: models.LauncherType | str) -> str:
+        """Custom serializer that can handle ULIDs."""
```
olevski:
```suggestion
        """Custom serializer that can handle launcher types."""
```

olevski (changes requested):
A few minor changes Eike. Everything else look good.

components/renku_data_services/notebooks/api.spec.yaml
eikek: Thank you, yes that makes sense.

components/renku_data_services/notebooks/core_sessions.py
eikek: Yeah, good point. I was wondering the same and couldn't decide. Currently I cannot think about filtering for a specific submission id. That's why it's now in annotations. @leafty and I were talking about it quickly, and only because we didn't know better, it is in annotations. I can put it into labels?

sgaist: Beside the filtering, which I agree is a good point, annotations are usually used to alter the behaviour of a pod so if it is not used for that purpose, it pushes in favor of moving it to labels.

components/renku_data_services/notebooks/models.py
```diff
@@ -28,6 +28,39 @@
...
+
+    @classmethod
+    def validate(cls, v: Any) -> str | None:
+        """Validate a string for a submission id."""
+        if not isinstance(v, str):
+            return f"Invalid submission id: {v}"
+        if len(v.strip()) < 4:
+            return f"Invalid submission id (requires at least 4 characters): {v}"
+        return None
+
```
sgaist, line 69: `validate` reads boolean like, I think it should rather raise an error if something is wrong.
```suggestion
class SubmissionId:
    """Type for a submission id used to start jobs."""

    def __init__(self, value: Any) -> None:
        self.validate(value)
        self.value: str = value
        
    def __str__(self) -> str:
        return self.value

    def __repr__(self) -> str:
        return self.value

    @classmethod
    def from_str(cls, v: str) -> SubmissionId | None:
        """Create a submission id from a str if it validates."""
        try:
            cls.validate(v)
        except ValueError:
            return None
        return SubmissionId(v.strip())

    @classmethod
    def validate(cls, v: Any) -> None:
        """Validate a string for a submission id."""
        if not isinstance(v, str):
            raise ValueError(f"Invalid submission id: {v}")
        if len(v.strip()) < 4:
            raise ValueError(f"Invalid submission id (requires at least 4 characters): {v}")
```

sgaist (dismissed):

components/renku_data_services/notebooks/models.py
eikek: Are you ok with renaming then? I try to avoid raising errors for doing control flow.

besides, i forgot to update the validation logic 🤦🏼

components/renku_data_services/notebooks/core_sessions.py
leafty: What? NO! Labels are expensive! They are used by etcd to perform selections so they go into an indexing engine. Annotations are just metadata which are not indexed, any metadata can be added to annotations if they are of some usefuleness.

Reference: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/

> Labels allow for efficient queries and watches and are ideal for use in UIs and CLIs. Non-identifying information should be recorded using [annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/).

@eikek Please keep in annotations.

leafty: > Beside the filtering, which I agree is a good point, annotations are usually used to alter the behaviour of a pod so if it is not used for that purpose, it pushes in favor of moving it to labels.

The k8s docs DO NOT say anything about "alter the behaviour of a pod":

> You can use either labels or annotations to attach metadata to Kubernetes objects. Labels can be used to select objects and to find collections of objects that satisfy certain conditions. In contrast, annotations are not used to identify and select objects. The metadata in an annotation can be small or large, structured or unstructured, and can include characters not permitted by labels. It is possible to use labels as well as annotations in the metadata of the same object.

https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/

Annotations is a very good place to store metadata about a k8s resource. It is _de-facto_ the first place where any piece of metadata about a k8s resource should be placed unless there is a use-case for labels (identifying data) or other way of storing that metadata.

components/renku_data_services/notebooks/util/kubernetes_.py
```diff
@@ -76,7 +76,7 @@ def _make_server_name_prefix(safe_username: str, submission_id: SubmissionId | N
 
     prefix = f"{prefix}{safe_username}"
     if submission_id:
-        prefix = f"{prefix}_{submission_id}"
+        prefix = f"j-{prefix}-{submission_id}"
```
leafty, line 79: Do we _really_ need the `j-` prefix? This eats out 2 characters out of the 12 available.

components/renku_data_services/notebooks/models.py
sgaist: Sure thing !

components/renku_data_services/notebooks/core_sessions.py
leafty: As a matter of fact, using annotations to change the behavior of a k8s resource is an anti-pattern and is why the Gateway API is replacing the Ingress resource in k8s.

https://gateway-api.sigs.k8s.io/guides/getting-started/migrating-from-ingress/

> The annotations approach to extensibility has two negative consequences for users of the Ingress API:
> 
> - Limited portability. Because so many features are available through annotations, switching between Ingress controllers becomes difficult or even impossible, as it is necessary to convert the annotations of one implementation to another (the other implementation might not even support some features of the first one). This limits the portability of the Ingress API.
> - Awkwardness of the API. Because annotations are key-value strings (as opposed to a structured scheme like the spec of the Ingress resource) and applied at the top of a resource (rather than in the relevant parts of the spec), the Ingress API can become awkward to use, especially when a large number of annotations are added to an Ingress resource.

sgaist: I was talking about a common use case of annotations through the [downward API](https://kubernetes.io/docs/concepts/workloads/pods/downward-api/#downwardapi-fieldRef) not the purpose of the annotations themselves.

In the case at hand, unless I misunderstood, it would be useful information to be able to filter these pods. If that is not the case, then yes, it should stay into annotations.

leafty (dismissed):

components/renku_data_services/notebooks/util/kubernetes_.py
eikek: oh right, i overlooked that the caller only takes the first 12 characters of the result of that function. As I understood, the `j-` is for ui purposes for admins, right? So perhaps we can remove it to save two characters. I don't have a strong opinion, though, slightly in favor of not using `j-`.

olevski: See the other comment. It is so that all jobs show up together when you list them. If we dont add this, the jobs show up interspersed with all the renku services and renku sessions. And it is hard to see what is what. It is also hard to see at a glance what is a job and what is not when looking at pods. So it just for admins. And I think is worth it. I have gotten similar requests even for session from admins.

leafty: Would this be better served by using k8s labels?

olevski (dismissed):

components/renku_data_services/notebooks/core_sessions.py
leafty, line 1006: Comment: it could be useful to also set an env var inside the job (added to the existing `RENKU_` env vars), e.g. `RENKU_JOB_SUBMISSION_ID=...`

eikek: Ah yes, good idea. I will add it to the env vars.

components/renku_data_services/notebooks/util/kubernetes_.py
olevski: K8s labels do not affect how the pods are listed when you use kubectl or k9s and are just listing all pods. And the main reason for the addition is to make the lives of admins easier so that they do not have to filter by labels and can just glance at the listing of pods to see what is a job and what is not.

olevski (dismissed):

olevski (dismissed):

olevski (dismissed):

olevski (approved):

PR #1329, opened 2026-05-26
feat: export to envidat

mohammad-alisafaee:
/deploy


components/renku_data_services/data_connectors/deposits/envidat.py
```diff
@@ -0,0 +1,45 @@
...
+
+    package_id: str
+    status: str  # "draft", "pending", or "published"
+    renku_id: str
+    doi: str
+
+    @property
+    def is_published(self) -> bool:
+        """Whether the deposit has been published on Envidat."""
+        return self.status.lower() == "published"
```
sgaist: Shouldn't that be a StrEnum ?

e.g.
```suggestion
EnvidatStatus(StrEnum):
    draft = auto()
    pending = auto()
    published = auto()

class EnvidatDepositStatus(BaseModel):
    """Status response from the Envidat deposit status endpoint."""

    package_id: str
    status: EnvidatStatus
    renku_id: str
    doi: str

    @property
    def is_published(self) -> bool:
        """Whether the deposit has been published on Envidat."""
        return self.status == EnvidatStatus.published
```

components/renku_data_services/data_connectors/deposits/envidat.py
```diff
@@ -0,0 +1,45 @@
...
+        url = f"{self.__base_url}/api/action/renku/{renku_id}/"
+        res = await self.__client.get(url)
+        if res.status_code == 404:
+            raise errors.MissingResourceError(
+                message=f"Envidat deposit with Renku ID '{renku_id}' was not found.",
+            )
+        if res.status_code >= 300 or res.status_code < 200:
+            raise errors.ThirdPartyAPIError(
+                message=f"Received unexpected status code {res.status_code} when checking Envidat deposit status.",
+                detail=f"Message from Envidat: {res.text[:500]}",
```
sgaist: Out of curiosity, why the 500 limit ?

components/renku_data_services/data_connectors/blueprints.py
```diff
@@ -687,23 +720,31 @@ async def _patch_deposit(
...
+                        token = await self.__get_zenodo_access_token(user)
+                        zenodo_dep = await self.zenodo_client.get_deposit(token, saved_dep.deposit.original_id)
+                        if not zenodo_dep:
+                            raise errors.MissingResourceError(
+                                message=f"The deposit with id {saved_dep.deposit.original_id} "
+                                "cannot be found from the provider."
+                            )
+                        if not zenodo_dep.submitted:
+                            raise errors.ValidationError(
+                                message="The deposit needs to be completed and published first before being completed."
```
sgaist: The error on line 728 mentions Envidat but here Zenodo is missing. Should we have it as well ?

components/renku_data_services/data_connectors/deposits/envidat.py
```diff
@@ -0,0 +1,45 @@
...
+        """Whether the deposit has been published on Envidat."""
+        return self.status.lower() == "published"
+
+
+class EnvidatClient:
+    """Client for interacting with the Envidat deposit status API."""
+
+    def __init__(self) -> None:
+        self.__base_url = os.environ.get("ENVIDAT_URL", "https://www.envidat.ch").rstrip("/")
+        self.__client = httpx.AsyncClient()
```
olevski: There likely wont be any redirects but it does not hurt to be safe if some appear later on.

```suggestion
        self.__client = httpx.AsyncClient(follow_redirects=True)
```

components/renku_data_services/data_connectors/config.py
```diff
@@ -24,6 +24,12 @@ class DepositConfig:
     namespace: str
     renku_url: str
     zenodo_url: str
+    envidat_url: str
+    envidat_rclone_image: str
+    envidat_s3_endpoint: str
+    envidat_s3_bucket: str
+    envidat_s3_access_key_id: str
+    envidat_s3_secret_access_key: str
```
sgaist: Since more deposit are getting added with more complex configuration requirements, shouldn't we use a dedicated classes for these ?
That would help avoid constructors becoming giant lists of parameters.

```
@dataclass
class EnvidatConfig:
    url: str
    rclone_image: str
    s3_endpoint: str
    s3_bucket: str
    s3_access_key_id: str
    s3_secret_access_key: str

    @classmethod
    def from_env(cls):
        return cls(
            url=os.environ.get("ENVIDAT_URL", "https://www.envidat.ch").rstrip("/"),
            rclone_image=os.environ.get("ENVIDAT_RCLONE_IMAGE", "rclone/rclone:1"),
            s3_endpoint=os.environ.get("ENVIDAT_S3_ENDPOINT", ""),
            s3_bucket=os.environ.get("ENVIDAT_S3_BUCKET", ""),
            s3_access_key_id=os.environ.get("ENVIDAT_S3_ACCESS_KEY_ID", ""),
            s3_secret_access_key=os.environ.get("ENVIDAT_S3_SECRET_ACCESS_KEY", ""),
        )
```

Then 

```
@dataclass
class DepositConfig:
    """The configuration for running data deposit uploads."""

    image: str
    namespace: str
    renku_url: str
    zenodo_url: str
    envidat: EnvidatConfig
    node_selector: dict[str, str] | None = None
    tolerations: list[V1Toleration] | None = None
    cluster_id: Final[ClusterId] = DEFAULT_K8S_CLUSTER
```

I know this moves the problem in a different class but I think it's worth pondering.
Each deposit configuration will manage what concerns itself while DepositConfig merely provides them.
I would look like overkill in the case of Zenodo but would likely be cleaner in the long term.

components/renku_data_services/data_connectors/config.py
```diff
@@ -64,4 +70,10 @@ def from_env(cls, renku_url: str) -> DepositConfig:
             namespace=os.environ["KUBERNETES_NAMESPACE"],
             cluster_id=DEFAULT_K8S_CLUSTER,
             zenodo_url=os.environ.get("ZENODO_URL", "https://zenodo.org").rstrip("/"),
+            envidat_url=os.environ.get("ENVIDAT_URL", "https://www.envidat.ch").rstrip("/"),
+            envidat_rclone_image=os.environ.get("ENVIDAT_RCLONE_IMAGE", "rclone/rclone:1"),
+            envidat_s3_endpoint=os.environ.get("ENVIDAT_S3_ENDPOINT", ""),
+            envidat_s3_bucket=os.environ.get("ENVIDAT_S3_BUCKET", ""),
+            envidat_s3_access_key_id=os.environ.get("ENVIDAT_S3_ACCESS_KEY_ID", ""),
+            envidat_s3_secret_access_key=os.environ.get("ENVIDAT_S3_SECRET_ACCESS_KEY", ""),
```
sgaist: With the suggestion above this would become:

```
    envidat=EnvidatConfig.from_env()
```

components/renku_data_services/data_connectors/core.py
```diff
@@ -890,15 +982,26 @@ async def dc_iter() -> AsyncIterator[models.DataConnectorWithSecrets]:
...
+        case _:
+            job = _create_zenodo_upload_job_manifest(
+                deposit_config=deposit_config,
+                deposit_job=deposit_job,
+                api_key_secret_name=base_name,
+                work_dir=work_dir,
+                suspended=True,
+                labels=labels,
+                pvc_name=pvc_name,
+            )
```
olevski, line 1009: Suggestion: I think is safer to not fallback to zenodo but check it and then for the fallback (i.e. when it is not zenodo or envidat we raise an error)

components/renku_data_services/data_connectors/core.py
```diff
@@ -593,16 +593,28 @@ def validate_deposit_patch(body: apispec.DepositPatch) -> models.DepositPatch:
     )
 
 
-def serialize_deposit(deposit: models.DepositJob) -> dict[str, Any]:
+def serialize_deposit(deposit: models.DepositJob, deposit_config: DepositConfig) -> dict[str, Any]:
     """Create an apispec Deposit from the internal model."""
+    match deposit.deposit.source:
+        case models.DepositSource.zenodo:
+            external_url = f"https://zenodo.org/uploads/{deposit.deposit.original_id}"
```
sgaist: Shouldn't the FQDN come from `deposit_config.zenodo_url` ?

components/renku_data_services/data_connectors/core.py
```diff
@@ -909,10 +1012,21 @@ async def dc_iter() -> AsyncIterator[models.DataConnectorWithSecrets]:
...
+        case models.DepositSource.envidat:
+            # Use RCLONE_CONFIG_{REMOTE}_{KEY} naming so that rclone auto-configures the Envidat remote
+            job_secret_data = {
+                "RCLONE_CONFIG_ENVIDAT_ACCESS_KEY_ID": deposit_config.envidat_s3_access_key_id,
+                "RCLONE_CONFIG_ENVIDAT_SECRET_ACCESS_KEY": deposit_config.envidat_s3_secret_access_key,
+            }
+        case _:
+            if deposit_api_key is None:
+                raise errors.ProgrammingError(message="A Zenodo deposit requires an API key.")
+            job_secret_data = {"ZENODO_API_KEY": deposit_api_key}
```
olevski, line 1032: Same as above. Add a case for zenodo and then raise an error if the source is not known. Rather than assuming that unknown source requries the same treatment as zenodo.

components/renku_data_services/data_connectors/blueprints.py
```diff
@@ -573,22 +575,40 @@ async def __get_zenodo_access_token(self, user: base_models.APIUser) -> str:
...
+        ):
+            # TODO: Is this check required!?
+            # NOTE: For now we disable this for tests to pass
+            pass
+            # raise errors.ConfigurationError(
+            #     message="Envidat S3 is not configured on this Renku instance.",
+            #     detail="Please ask your administrator to set ENVIDAT_S3_ENDPOINT, ENVIDAT_S3_BUCKET, "
+            #     "ENVIDAT_S3_ACCESS_KEY_ID, and ENVIDAT_S3_SECRET_ACCESS_KEY.",
+            # )
+
```
olevski: I think you can pass a flag that indicates that exporting to envidat is enabled. And if it is enabled you should have all those parameters set. Also you can make it so that they are all part of one class specific to envidat and then they are part of one field from the parent config. That way you dont have to check all of them. If envidat export is enabled the field with the envidat config should not be None.

components/renku_data_services/data_connectors/blueprints.py
```diff
@@ -687,23 +720,31 @@ async def _patch_deposit(
...
+                        zenodo_dep = await self.zenodo_client.get_deposit(token, saved_dep.deposit.original_id)
+                        if not zenodo_dep:
+                            raise errors.MissingResourceError(
+                                message=f"The deposit with id {saved_dep.deposit.original_id} "
+                                "cannot be found from the provider."
+                            )
+                        if not zenodo_dep.submitted:
+                            raise errors.ValidationError(
+                                message="The deposit needs to be completed and published first before being completed."
+                            )
```
olevski, line 753:
```suggestion
                match saved_dep.deposit.source:
                    case models.DepositSource.envidat:
                        envidat_status = await self.envidat_client.get_deposit_status(saved_dep.deposit.original_id)
                        if not envidat_status.is_published:
                            raise errors.ValidationError(
                                message="The deposit needs to be published on Envidat before being marked complete."
                            )
                    case models.DepositSource.zenodo:
                        token = await self.__get_zenodo_access_token(user)
                        zenodo_dep = await self.zenodo_client.get_deposit(token, saved_dep.deposit.original_id)
                        if not zenodo_dep:
                            raise errors.MissingResourceError(
                                message=f"The deposit with id {saved_dep.deposit.original_id} "
                                "cannot be found from the provider."
                            )
                        if not zenodo_dep.submitted:
                            raise errors.ValidationError(
                                message="The deposit needs to be completed and published first before being completed."
                            )
                    case x:
                        raise errors.ValidationError(message=f"Received unknown deposit source {x} when pathing deposit with ID {deposit_id}.")
```

components/renku_data_services/data_connectors/blueprints.py
```diff
@@ -600,15 +620,29 @@ async def _post_deposit(
...
+                    # NOTE: Each new Envidat deposit gets its own dir, so a failed deposit cannot be resumed and need to
+                    # be cleaned up manually on Envidat.
+                    original_id = str(ULID())
+                    deposit_api_key: str | None = None
+
+                case apispec.DepositProvider.zenodo:
+                    token = await self.__get_zenodo_access_token(user)
+                    zenodo_dep = await self.zenodo_client.create_deposit(token, body.name)
+                    original_id = str(zenodo_dep.id)
+                    deposit_api_key = token
```
olevski, line 652:
```suggestion
            match body.provider:
                case apispec.DepositProvider.envidat:
                    self._check_envidat_config()
                    # TODO: Should we use the deposit ULID as the directory name!?
                    # NOTE: Each new Envidat deposit gets its own dir, so a failed deposit cannot be resumed and need to
                    # be cleaned up manually on Envidat.
                    original_id = str(ULID())
                    deposit_api_key: str | None = None

                case apispec.DepositProvider.zenodo:
                    token = await self.__get_zenodo_access_token(user)
                    zenodo_dep = await self.zenodo_client.create_deposit(token, body.name)
                    original_id = str(zenodo_dep.id)
                    deposit_api_key = token

                case x:
                    raise errors.ValidationError(message=f"Received unknown deposit provider {x} when creating deposit.")
```

olevski (changes requested):

components/renku_data_services/data_connectors/config.py
olevski: I made the same comment because I was doing the review in vscode and did not see this comment. But yeah I agree with @sgaist here.

components/renku_data_services/data_connectors/blueprints.py
olevski: Can you clarify this comment please @sgaist. I am really not sure what you mean here. There is code to do the same check for envidat.

components/renku_data_services/data_connectors/deposits/envidat.py
olevski: I removed the limit. Maybe it was in case envidat responds with a full html or some really long response.

olevski:
I tested the CI deployment and data does get uploaded to the envidat bucket. There are a few outstanding problems however:
1. The page in envidat that shows what files have been uploaded and that allows to upload more files does not show the uploaded files we already placed in the bucket. This is something that envidat have to fix on their end.
2. Once the upload is complete we keep showing the url for it and a user can keep visiting the link. And every visit on the link allows for the creation of a new dataset. We should prevent this. I think we can hit the API from envidat and check the status of the dataset and prevent the link from showing up all the time.

EDIT: number 2 gets complicated and it requires pulling in even more of envidat's api into our own code. I think we can live with it. And if not we can do it as a followup / improvement.

components/renku_data_services/data_connectors/blueprints.py
sgaist: Sure:
In the `models.DepositSource.envidat` case, the error message mentions envidat explicitly.
I was wondering whether the zenodo specific errors should also explicitly mention zenodo.
That would help to diagnose the issues more easily.

olevski: Done I added Zenodo in the error messages. They are not generic now.

sgaist (dismissed):

olevski (approved):

PR #1332, opened 2026-05-29
fix: restore event-based sync from Keycloak

leafty:
wip

/deploy #notest


leafty:
Note: can try to improve event polling from Keycloak with version 26 deployed.

PR #1271, opened 2026-04-24
fix: error type when rejecting session start

leafty:
Closes #1270.

<code>
/deploy renku=feat/private-repository-build renku-ui=feat/build-private-repositories extra-values=dataService.imageBuilders.enabled=true,dataService.imageBuilders.strategyName=renku-buildpacks-v3,dataService.imageBuilders.outputImagePrefix=harbor.dev.renku.ch/renku-ci-private-images-build-public-project/,dataService.imageBuilders.outputPrivateImagePrefix=harbor.dev.renku.ch/renku-ci-private-images-build-private-project/,dataService.imageBuilders.pushSecretName=harbor-public-session,dataService.imageBuilders.pushPrivateSecretName=harbor-private-session,dataService.imageBuilders.pullPrivateSecretName=harbor-private-session,dataService.imageBuilders.nodeSelector.renku.io/node-purpose=user,dataService.imageBuilders.tolerations[0].effect=NoSchedule,dataService.imageBuilders.tolerations[0].key=renku.io/dedicated,dataService.imageBuilders.tolerations[0].operator=Equal,dataService.imageBuilders.tolerations[0].value=user
</code>

PR #1333, opened 2026-05-29
Refactor: improve image accessibility checks

sgaist:
## Describe your changes

This PR updates two things:

- The image availability checks with regard to private repository builds
- The image check endpoint

The former has moved into the ImageCheckRepository class so it is also available for other components to use.

The class is used by the `start_session` method as well as the Docker image check end point.

It order to make full use of it, the image check endpoint now has two exclusive parameters:

- An image url
- A launcher id

The former has been kept to avoid breaking change during implementation.
The latter shall be the new and prefered way to check for images.

## Tickets and links

Supersede #1271
Supersede #1283

/deploy #notest renku=feat/private-repository-build extra-values=dataService.imageBuilders.enabled=true,dataService.imageBuilders.strategyName=renku-buildpacks-v3,dataService.imageBuilders.outputImagePrefix=harbor.dev.renku.ch/renku-ci-private-images-build-public-project/,dataService.imageBuilders.outputPrivateImagePrefix=harbor.dev.renku.ch/renku-ci-private-images-build-private-project/,dataService.imageBuilders.nodeSelector.renku.io/node-purpose=user,dataService.imageBuilders.tolerations[0].effect=NoSchedule,dataService.imageBuilders.tolerations[0].key=renku.io/dedicated,dataService.imageBuilders.tolerations[0].operator=Equal,dataService.imageBuilders.tolerations[0].value=user


components/renku_data_services/notebooks/image_check.py
```diff
@@ -91,15 +97,78 @@ class ImageCheckRepository:
...
+                )
+
+    @overload
+    async def check_image(
+        self, user: APIUser, gitlab_user: APIUser | None, image_src: SessionLauncher
+    ) -> CheckResult: ...
+
+    @overload
+    async def check_image(self, user: APIUser, gitlab_user: APIUser | None, image_src: Image) -> CheckResult: ...
+
```
leafty: I don't think these overloads do anything. The `image_src` type is already `SessionLauncher | Image`.

components/renku_data_services/notebooks/image_check.py
```diff
@@ -91,15 +97,78 @@ class ImageCheckRepository:
...
+    async def check_image(
+        self, user: APIUser, gitlab_user: APIUser | None, image_src: SessionLauncher | Image
+    ) -> CheckResult:
+        """Check access to the image from the given launcher or image and provide image and access details."""
+
+        if isinstance(image_src, SessionLauncher):
+            await self.__check_built_image_accessibility(user, gitlab_user, image_src)
```
leafty: The `check_image()` method should not `raise` when the user has no access to the image. In this case, the method should return a `CheckResult` instance, e.g.

```python
return CheckResult(
    accessible=False,
    platforms=<platforms>,
    response_code=status_code,  # Should be a 401 or 403
    image_provider=image_provider, # Can be None
    token=None,
    error=unauth_error, # The error
)
```

See how it is implemented for the existing cases.

components/renku_data_services/notebooks/image_check.py
```diff
@@ -91,15 +97,78 @@ class ImageCheckRepository:
...
+    ) -> None:
+        # image can be "image:unknown-at-the-moment" which is returned until the first
+        # successful build however users can force a session start although it will fail
+        environment = launcher.environment
+        image = launcher.environment.container_image
+
+        if (
+            self.builds_config.build_output_private_image_prefix
+            and image.startswith(self.builds_config.build_output_private_image_prefix)
+            or image == "image:unknown-at-the-moment"
```
leafty: ? Why do we match `image:unknown-at-the-moment`?

components/renku_data_services/notebooks/image_check.py
```diff
@@ -91,15 +97,78 @@ class ImageCheckRepository:
...
+            if environment.build_parameters is None:
+                raise errors.ValidationError(message="Image was built but build parameters are missing")
+
+            project = await self.project_repo.get_project(user=user, project_id=launcher.project_id)
+            git_providers = await self.git_provider_helper.get_providers(user=user)
+            repositories = repositories_from_project(project, git_providers)
+
+            build_repository = environment.build_parameters.repository
+            if build_repository not in [repository.url for repository in repositories]:
+                raise errors.ValidationError(message="Image was not built from a repository in this project")
```
leafty: Why do we check this?

components/renku_data_services/notebooks/image_check.py
```diff
@@ -91,15 +97,78 @@ class ImageCheckRepository:
...
+            self.builds_config.build_output_private_image_prefix
+            and image.startswith(self.builds_config.build_output_private_image_prefix)
+            or image == "image:unknown-at-the-moment"
+        ):
+            if environment.build_parameters is None:
+                raise errors.ValidationError(message="Image was built but build parameters are missing")
+
+            project = await self.project_repo.get_project(user=user, project_id=launcher.project_id)
+            git_providers = await self.git_provider_helper.get_providers(user=user)
+            repositories = repositories_from_project(project, git_providers)
```
leafty: We won't need this `repositories`.

components/renku_data_services/notebooks/image_check.py
```diff
@@ -91,15 +97,78 @@ class ImageCheckRepository:
...
+
+            build_repository = environment.build_parameters.repository
+            if build_repository not in [repository.url for repository in repositories]:
+                raise errors.ValidationError(message="Image was not built from a repository in this project")
+
+            repo_data = await self.git_repositories_repo.get_repository(
```
leafty: Again incorrect logic:
1. List the builds from the builds history, and find the latest successful build
2. The latest successful build is an instance of `BuildORM` and should have the following columns populated:
    * `result_image` -> if `latest_successful_build.result_image != image` raise a Programing error (database has an inconsistent state)
    * `result_repository_url` -> this is the repository used for the build (this is reported by Shipwright). Use _that_ URL to check the permissions, NOT anything found from the project.

components/renku_data_services/notebooks/image_check.py
```diff
@@ -91,15 +97,78 @@ class ImageCheckRepository:
...
+                repository_url=build_repository,
+                user=user,
+                etag=None,
+                internal_gitlab_user=gitlab_user if gitlab_user else APIUser(),
+            )
+            if repo_data.is_error:
+                raise errors.ValidationError(message=str(repo_data.error))
```
leafty: nitpick: ugh can we avoid raising `metadata_unknown`? It's super unhelpful for the user.

components/renku_data_services/notebooks/image_check.py
```diff
@@ -91,15 +97,78 @@ class ImageCheckRepository:
...
+            if isinstance(repo_data.metadata, RepoMetadata) and not repo_data.metadata.pull_permission:
+                raise errors.ForbiddenError(
+                    message="You do not have pull access to the code repository used for this session."
+                )
```
leafty: Note: This actually never gets raised because when the user has no access the error is `metadata_unknown`.

components/renku_data_services/notebooks/image_check.py
```diff
@@ -91,15 +97,78 @@ class ImageCheckRepository:
...
+        self.builds_config = builds_config
+        self.git_provider_helper = git_provider_helper
+        self.project_repo = project_repo
+        self.git_repositories_repo = git_repositories_repo
         self.connected_services_repo = connected_services_repo
         self.oauth_client_factory = oauth_client_factory
 
-    async def check_image(self, user: APIUser, gitlab_user: APIUser | None, image: Image) -> CheckResult:
-        """Check access to the given image and provide image and access details."""
+    async def __check_built_image_accessibility(
```
leafty: This method needs to not `raise` or `check_image` needs to catch.

components/renku_data_services/notebooks/core_sessions.py
```diff
@@ -681,12 +668,16 @@ def __format_image_pull_secret(secret_name: str, access_token: str, registry_dom
 
 
 async def __get_connected_services_image_pull_secret(
```
leafty, line 670: Rename the method because `__get_connected_services_image_pull_secret` is not what is happening now.

leafty (changes requested):

components/renku_data_services/notebooks/image_check.py
leafty: To perform (1), you most likely need to query the db or call `get_environment_builds()` from `components/renku_data_services/session/db.py`. `get_environment_builds()` returns builds with the latest first, so the first item with `status==Succeeded` is the last successful build.

components/renku_data_services/notebooks/core_sessions.py
sgaist: I agree there are additional checks done under the hood in `image_check_repo.check_image`  but that function still does the same job, doesn't it ?

What name would you suggest ?

components/renku_data_services/notebooks/image_check.py
sgaist: Because if we don't, while the first build is not done, it would bypass these checks.

components/renku_data_services/notebooks/image_check.py
sgaist: I thought it would be strange to try to start a session with an image that will contain the code from a repository that is not listed as part of the project.
But this is not something that a user will usually see when starting a session so I can drop it.

components/renku_data_services/notebooks/image_check.py
sgaist: Would you happen to know if you also get `metadata_unknown` with GitLab ? The Guest role for example cannot pull while it can see the code on GitLab itself.

components/renku_data_services/notebooks/image_check.py
sgaist: I mainly used them for the type checking while developing. I don't mind removing them.

components/renku_data_services/notebooks/image_check.py
sgaist: Ok, I'll refactor it.

components/renku_data_services/notebooks/image_check.py
sgaist: A generic message such as "Something went wrong during image check" ?

components/renku_data_services/notebooks/image_check.py
sgaist: Noted, I will refactor this code a bit based on your feedback.

components/renku_data_services/notebooks/image_check.py
sgaist: For registered users, this won't work until the first successful build which can take a while.
It is also not something that can be used when the user is anonymous (it triggers are rightful 401).

So I think we must clarify some rules for that check. I would suggest:
- If an anonymous user try to use an image from a private repository build, it should fail
- Until a successful build, the image should be considered inaccessible with an adequate message
- Once a build is successful we can do the checks you suggest

What do you think ?

leafty: Pleaaaaase keep to the rules listed, they work as intended!

* if the user is anonymous, do not add any `ImagePullSecret`: this is the existing behavior. => If the image is not publicly accessible, the image cannot be pulled. There should already be a short-circuit for that case in `core_sessions.py`.
* if the user is authenticated:
  1. List the builds from the builds history, and find the latest successful build. If the user does not have access to the build history, go to step (iii).
  2. The latest successful build is an instance of `Build`/`BuildORM` and should have the following columns populated:
      * `result.image`/`result_image` -> if `latest_successful_build.result_image != image` raise a Programing error (database has an inconsistent state)
      * `result.repository_url`/`result_repository_url` -> this is the repository used for the build (this is reported by Shipwright). Use _that_ URL to check the permissions, NOT anything found from the project.
  3. If the step above did not yield a match, do not add any `ImagePullSecret`: this is also existing behavior.

Why does this work? And why don't we stop the user?

1. We need to keep the current behavior of the `/api/data/sessions` API. The API does NOT stop users from launching sessions that cannot work. E.g. if the image is `foobar:doesnotexist` then it will attempt launch.
2. If the image is not built yet, this is the same story, the API will not stop the user.

You may think that the API _should_ stop the user from launching the session, but this is incorrect! We are NOT changing the design and choices that already exist about sessions. Stopping the user is done by the UI which checks image access, code repository access and data connector access. If you are not happy about this, e.g. maybe the API should have an "easy" endpoint to check for launch issues (e.g. `POST /api/data/sessions?dry_run=1`) then let's open a GitHub issue to fix this.

components/renku_data_services/notebooks/image_check.py
leafty: OK, we can leave this out for now. I see that what should be done is to properly convert the `type RepositoryError = GitUrlError | RepositoryMetadataError` errors into user-readable error message and not just use the enum values like `metadata_unknown`.

components/renku_data_services/notebooks/image_check.py
leafty: I see. Both can happen, so we can keep this code or update it for more user-friendly messages.

components/renku_data_services/notebooks/image_check.py
sgaist: Done

components/renku_data_services/notebooks/image_check.py
sgaist: Refactored to follow the rules you described.
I put them in the commit message to ensure we have a reference for why the implementation is done as it is.
I'll create a ticket as suggested.

components/renku_data_services/notebooks/image_check.py
sgaist: Refactored as requested

components/renku_data_services/notebooks/image_check.py
sgaist: Removed

components/renku_data_services/notebooks/image_check.py
sgaist: Removed with the refactoring

components/renku_data_services/notebooks/api.spec.yaml
```diff
@@ -150,13 +150,21 @@ paths:
...
       summary: Check if a session image exists
       parameters:
-      - description: The Docker image URL (tag included) that should be fetched.
-        in: query
-        name: image_url
-        required: true
-        schema:
-          type: string
-          minLength: 1
+        - description: >
```
leafty: GitHub cannot parse:

```suggestion
        - description: |
```

components/renku_data_services/notebooks/api.spec.yaml
```diff
@@ -150,13 +150,21 @@ paths:
...
+            The Docker image URL (tag included) that should be fetched.
+            (Must NOT be used at the same time as session_id)
+          in: query
+          name: image_url
+          schema:
+            type: string
+            minLength: 1
+        - description: >
```
leafty:
```suggestion
        - description: |
```

components/renku_data_services/notebooks/core_sessions.py
```diff
@@ -681,12 +668,20 @@ def __format_image_pull_secret(secret_name: str, access_token: str, registry_dom
...
-    secret_name: str, image_check_repo: ImageCheckRepository, image: str, user: APIUser
+    secret_name: str,
+    image_check_repo: ImageCheckRepository,
+    launcher: SessionLauncher,
+    user: APIUser,
 ) -> ExtraSecret | None:
```
leafty, line 675: I don't understand why this method `__get_connected_services_image_pull_secret()` is modified here. The `__get_connected_services_image_pull_secret()` needs to be called when the image is hosted in an external registry and we may have a OAuth 2.0 token for it.

I do not see the need to change the logic.

components/renku_data_services/notebooks/core_sessions.py
```diff
@@ -711,16 +706,11 @@ async def get_image_pull_secret(
...
-    if v2_secret:
-        return v2_secret
-
     if (
         builds_config.enabled
         and builds_config.build_output_private_image_prefix is not None
-        and image.startswith(builds_config.build_output_private_image_prefix)
+        and launcher.environment.container_image.startswith(builds_config.build_output_private_image_prefix)
+        and not user.is_anonymous
     ):
```
leafty, line 740: We need to perform the permission check here!
If the image is built from a private repository, this if condition will match and we will provide the image pull secret without checking the permissions!

Suggestion for the code:

```python
    # 1. Check for images hosted on external registries
    v2_secret = await __get_connected_services_image_pull_secret(
        f"{server_name}-image-secret", image_check_repo, image, user
    )
    if v2_secret:
        return v2_secret

    # 2. Check for images built from a private repository
    if (
        builds_config.enabled
        and builds_config.build_output_private_image_prefix is not None
        and launcher.environment.container_image.startswith(builds_config.build_output_private_image_prefix
    ):
        v2_secret = await __check_private_image_build(...) # <- this is a new method
        if v2_secret:
            return v2_secret
```

With:

```python
async def __check_private_image_build(
    secret_name: str,
    image_check_repo: ImageCheckRepository,
    launcher: SessionLauncher,
    user: APIUser,
) -> ExtraSecret | None:
     """Return a secret for pulling an image built from a private repository if the user can pull from the source repository."""
     ...
```

components/renku_data_services/notebooks/image_check.py
```diff
@@ -91,15 +96,103 @@ class ImageCheckRepository:
...
+    async def check_image(
+        self, user: APIUser, gitlab_user: APIUser | None, image_src: SessionLauncher
+    ) -> CheckResult: ...
+
+    @overload
+    async def check_image(self, user: APIUser, gitlab_user: APIUser | None, image_src: Image) -> CheckResult: ...
+
+    async def check_image(
+        self, user: APIUser, gitlab_user: APIUser | None, image_src: SessionLauncher | Image
+    ) -> CheckResult:
```
leafty, line 166: I don't think this works as intended.

Let's consider this situation: `user=AuthenticatedAPIUser(...), image_src=SessionLauncher(...)` with `image_src.environment.container_image` matching the private harbor registry. The image has been built successfully.

Then `__check_built_image_accessibility()` returns `None` (user has pull permission, and we set `image = Image.from_path(image_src.environment.container_image)`.

We continue below (L196):
* `image_provider` should be `None` (there is no OAuth 2.0 integration for the private harbor registry).
* `connected_user` is also `None`
* `connection` is also `None`

This means that we try to pull the image manifest with the public client (initialized at L196). Of course since this is the private harbor registry, then we will get a 401, 403 or 404 response (no public access to the image).

Considering that `connection` is None, we will return:
```python
        return CheckResult(
            accessible=False, # from `status_code == 200`
            platforms=None,
            response_code=4xx,
            image_provider=None,
            token=None, # from using a public client
            error=None, # never set in our case
        )
```

Why do we try to do both of the checks at once here?

components/renku_data_services/notebooks/blueprints.py
```diff
@@ -221,13 +223,28 @@ async def _check_docker_image(
...
+                logger.info(f"Checked image {query.image_url}: {result}")
+            elif query.launcher_id:
+                launcher_id = cast(ULID, ULID.from_str(query.launcher_id))
+                launcher = await self.session_repo.get_launcher(user=user, launcher_id=launcher_id)
+                result = await self.image_check_repo.check_image(
+                    user=user,
+                    gitlab_user=internal_gitlab_user,
+                    image_src=launcher,
+                )
+                logger.info(f"Checked image {launcher.environment.container_image}: {result}")
```
leafty: (See other comments for context)

Why do we call `self.image_check_repo.check_image()` in both cases?

To me it we would be more sound to do:
1. If `query.image_url` -> then check with `self.image_check_repo.check_image()` and keep the existing code for `check_image()`
2. If `query.launcher_id` -> then perform the repository permission check with `self.image_check_repo.__check_built_image_accessibility()`

components/renku_data_services/notebooks/image_check.py
leafty: Using `@overload` is useful if the returned type varies depending on the input types. This is not the case here so we can remove the overloads.

```suggestion
```

components/renku_data_services/notebooks/image_check.py
```diff
@@ -91,15 +96,103 @@ class ImageCheckRepository:
...
+            builds = await self.session_repo.get_environment_builds(user=user, environment_id=environment.id)
+
+            latest_successful_build = None
+            for build in builds:
+                if build.status == BuildStatus.succeeded:
+                    latest_successful_build = build
+                    break
+
+            if latest_successful_build is None or latest_successful_build.result is None:
+                raise errors.ValidationError(message="Image is not yet ready")
```
leafty: This condition should not happen, that is to say: if `latest_successful_build is None or latest_successful_build.result is None` then we have a private image for which we cannot know how it was built. The image exists, but we cannot trace back its origin.


```suggestion
                raise errors.ValidationError(message=f"Cannot get source repository for image {image}. You may need to rebuild it.")
```

Note: we can track as a new issure: "consider adding the source repository into the built image annotations. See: buildrun.spec.output.annotations".

components/renku_data_services/notebooks/image_check.py
```diff
@@ -91,15 +96,103 @@ class ImageCheckRepository:
...
+
+            if latest_successful_build.result.image != image:
+                raise errors.ProgrammingError(message="Database inconsistency")
```
leafty:
```suggestion
                raise errors.ProgrammingError(message=f"Cannot get source repository for image {image}: the build history is not consistent with the current image. You may need to rebuild it.")
```

components/renku_data_services/notebooks/image_check.py
```diff
@@ -91,15 +96,103 @@ class ImageCheckRepository:
...
+
+            if environment.build_parameters is None:
+                raise errors.ValidationError(message="Image was built but build parameters are missing")
```
leafty: Why do we check this?


```suggestion
```

components/renku_data_services/notebooks/image_check.py
```diff
@@ -91,15 +96,103 @@ class ImageCheckRepository:
...
+                internal_gitlab_user=gitlab_user if gitlab_user else APIUser(),
+            )
+
+            if repo_data.is_error:
+                raise errors.ValidationError(message="There was an issue accessing repo information")
+
+            if isinstance(repo_data.metadata, RepoMetadata) and not repo_data.metadata.pull_permission:
+                raise errors.ForbiddenError(
+                    message="You do not have pull access to the code repository used for this session."
+                )
```
leafty: In both cases we should return a forbidden error.

Also, maybe specify the repository in the error response:

```python
raise errors.ForbiddenError(
  message=f"You do not have pull access to the code repository {repository_url} which was used to build the image {image}."
)
```

components/renku_data_services/notebooks/image_check.py
sgaist: I thought I dropped it.

components/renku_data_services/notebooks/image_check.py
sgaist: `latest_successful is None` can happen during the initial build so should we rather have:

```
            if latest_successful_build is None:
                raise errors.ValidationError(message="Image is not yet ready")
            elif latest_successful_build.result is None:
                raise errors.ValidationError(message=f"Cannot get source repository for image {image}. You may need to rebuild it.")
```

About your note, ticket created: https://github.com/SwissDataScienceCenter/renku-data-services/issues/1346

components/renku_data_services/notebooks/image_check.py
sgaist: Done

components/renku_data_services/notebooks/image_check.py
sgaist: I see your point. You are correct, it does not work as intended for the private registry (but it does for the public builds).

My goal was to keep as much as possible the same flow when checking from a launcher than with just the image URL. So basically:
 
1. Check if the image is being built
2. Check if the user is allowed to access this image
3. Grab the information as with the URL

Point 2 as implemented is missing the separation between private and public builds.

I mixed the use of the secret to pull the image in Kubernetes with the authentication to the registry for this part and thus thought that the client that would be generated through the Image object would have access to the registry.

Just to be sure I get things correctly: should `check_image` be called only for non private builds images ? And thus `__check_built_image_accessibility` be called prior to any use of it (moved out of this class).

components/renku_data_services/notebooks/blueprints.py
sgaist: I'll refactor that following your other comments as well.

In the case of a private image build, what would the `ImageCheckResponse` look like ?

components/renku_data_services/notebooks/core_sessions.py
sgaist: I see your point.

It's because I added some of the checks in `check_image` to simplify the code globally but that was likely the wrong place to do them.

components/renku_data_services/notebooks/core_sessions.py
sgaist: Agreed, I'll refactor that.

components/renku_data_services/notebooks/core_sessions.py
leafty: Resolving as: no need to change the method name if it is reverted to its original implementation.

components/renku_data_services/notebooks/core_sessions.py
leafty: Let's revert to the original implementation.

components/renku_data_services/notebooks/image_check.py
leafty: I am not sure I understand why we would ever call `__check_built_image_accessibility()` with `image="image:unknown-at-the-moment"`.

This method should be called only if we match `image.startswith(builds_config.build_output_private_image_prefix)`.

Thus the image exists and has been pushed to harbor-private, so there must be a `last_successful_build`. If that is `None` then the build history was lost.

components/renku_data_services/notebooks/api.spec.yaml
leafty, line 1: It seems to me that the code changes implied by this API change are causing a lot of difficulties. Maybe we should drop them from this PR and focus on implementing the correct image pull secret checks at session start.

components/renku_data_services/notebooks/blueprints.py
leafty: See: https://github.com/SwissDataScienceCenter/renku-data-services/pull/1333/changes#r3374256633
Maybe we should drop the changes to this file for now.

components/renku_data_services/notebooks/image_check.py
leafty: See: https://github.com/SwissDataScienceCenter/renku-data-services/pull/1333/changes#r3374256633
Maybe we should drop the changes to the `check_image()` method.

components/renku_data_services/notebooks/image_check.py
```diff
@@ -91,15 +96,93 @@ class ImageCheckRepository:
...
+        self.git_repositories_repo = git_repositories_repo
+        self.session_repo = session_repo
         self.connected_services_repo = connected_services_repo
         self.oauth_client_factory = oauth_client_factory
 
-    async def check_image(self, user: APIUser, gitlab_user: APIUser | None, image: Image) -> CheckResult:
-        """Check access to the given image and provide image and access details."""
+    async def __check_built_image_accessibility(
+        self, user: APIUser, gitlab_user: APIUser | None, launcher: SessionLauncher
+    ) -> None:
```
leafty: Let's consider that we call this method only from `get_image_pull_secret()` and only in this branch:

```python
if (
        builds_config.enabled
        and builds_config.build_output_private_image_prefix is not None
        and launcher.environment.container_image.startswith(builds_config.build_output_private_image_prefix)
        and not user.is_anonymous
    ):
```

Then the implemtation here works; it will check that the user can read the source repository.

In `get_image_pull_secret()` we then provide the image pull secret for `private-harbor`.

leafty (commented):
See comments below to get this working with the minimal amount of code changes:

components/renku_data_services/notebooks/image_check.py
sgaist: I refactored the code so that we call `check_built_image_accessibility` explicitly when it make sense.

components/renku_data_services/notebooks/image_check.py
sgaist: Sorry, I saw your comments after I refactored the code.

With the new version, do you still think the now simpler changes in `check_image` should be dropped ?

components/renku_data_services/notebooks/api.spec.yaml
sgaist: Same as below, I only just saw this now.

Can you check the latest implementation to see if you still would like to drop this part ?

components/renku_data_services/notebooks/blueprints.py
```diff
@@ -221,13 +223,50 @@ async def _check_docker_image(
...
+                        )
+                    except (errors.ValidationError, errors.ProgrammingError, errors.ForbiddenError) as error:
+                        result = CheckResult(
+                            accessible=False,
+                            platforms=None,
+                            response_code=403 if isinstance(error, errors.ForbiddenError) else 422,
+                            error=error,
+                        )
+                    else:
+                        result = CheckResult(accessible=True, platforms=None, response_code=200)
```
leafty: Code is OK
Maybe we can have a comment here, e.g. `# NOTE: the image _should_ exist in the private registry but the current implementation does not check that.`.

components/renku_data_services/notebooks/image_check.py
```diff
@@ -48,7 +53,9 @@ class CheckResult:
     response_code: int
     image_provider: ImageProvider | None = None
     token: str | None = field(default=None, repr=False)
-    error: errors.UnauthorizedError | None = None
+    error: (
+        errors.UnauthorizedError | errors.ValidationError | errors.ProgrammingError | errors.ForbiddenError | None
+    ) = None
```
leafty, line 58: This may be too many error types, but we can fix that later.

Actually, reading the code in `blueprints.py`, the new error types will not generate any new useful data for the API client.

components/renku_data_services/notebooks/api.spec.yaml
leafty: See other comment: as it is, the updated check may not provide the most useful information to the client, so I think I would be happier having the "usability" changes being pulled into a new PR so that we can first fix the access controls on launching a session before fixing usability.

leafty (dismissed):
Implementation is OK as it is, but the changes to `GET /api/data/sessions/images` do not seem useful at the moment.

components/renku_data_services/notebooks/api.spec.yaml
sgaist: Done

components/renku_data_services/notebooks/image_check.py
sgaist: I reverted the changes so we can discuss the design/implications in the dedicated PR.

components/renku_data_services/notebooks/image_check.py
```diff
@@ -91,15 +98,91 @@ class ImageCheckRepository:
...
+        builds = await self.session_repo.get_environment_builds(user=user, environment_id=environment.id)
+
+        latest_successful_build = None
+        for build in builds:
+            if build.status == BuildStatus.succeeded:
+                latest_successful_build = build
+                break
+
+        if latest_successful_build is None:
+            raise errors.ValidationError(message="Image is not yet ready")
```
leafty: This is not what is happening: if `latest_successful_build is None` then we have an image (e.g. `harbor.cluster/private_builds:renku-build-123`) which presumably exists in the private registry but we cannot find a source build for it. So the image is not in a "Image is not yet ready" state. We just lost the build history (maybe the user switched from build from code to a custom image and back).

leafty (dismissed):
OK, see comment about inexact error message.

components/renku_data_services/notebooks/image_check.py
sgaist: I think we have double edged sword here:
- Until the first successful build this will be None
- After the first successful build, we have a different issue as you described.
What about using:
```suggestion
            raise errors.ValidationError(message="Image is not available")
```
This conveys the correct information in both cases.

leafty: > Until the first successful build this will be None

We explicitly do _not_ call this method while `image = "image:unknown-at-the-moment"` but only when it matches `harbor.domain/private/*` and at that point the image must already exist!

sgaist: Oh yes ! My bad ! I was too deep in the build logic itself rather than when this method is called.

So the message would rather be:
```suggestion
            raise errors.ValidationError(message="Successful build not found")
```

leafty (approved):

PR #1336, opened 2026-06-01
feat: add DELETE /apps endpoint

wesjdj:
/deploy


components/renku_data_services/renku_apps/repository.py
```diff
@@ -65,3 +68,25 @@ async def get_app(self, user: base_models.APIUser, app_name: str) -> App:
...
+    async def delete_app(self, user: base_models.APIUser, app_name: str) -> None:
+        """Delete an app by its name."""
+        if not user.is_authenticated or user.id is None:
+            raise errors.UnauthorizedError(message="You do not have the required permissions for this operation.")
+
+        runtime_state = await self.k8s_client.get_app_deployment(app_name)
+        if runtime_state is None:
+            logger.info(f"App with name {app_name} was not found.")
+            return None
+
```
olevski, line 81: You are missing the authorization check.

```suggestion
    async def delete_app(self, user: base_models.APIUser, app_name: str) -> None:
        """Delete an app by its name."""
        if not user.is_authenticated or user.id is None:
            raise errors.UnauthorizedError(message="You do not have the required permissions for this operation.")

        authorized = await self.authz.has_permission(user, ResourceType.project, launcher.project_id, Scope.WRITE)
        if not authorized:
            raise errors.MissingResourceError(
                message=f"Project with id '{launcher.project_id}' does not exist or you do not have access to it."
            )

        runtime_state = await self.k8s_client.get_app_deployment(app_name)
        if runtime_state is None:
            logger.info(f"App with name {app_name} was not found.")
            return None

```

olevski (changes requested):
This looks good Wes. I will come back to it when it is fully ready.

olevski (approved):

PR #1337, opened 2026-06-02
fix: slow resource usage query

olevski:
/deploy extra-values=dataService.dataTasks.enableResourceRequestTracking=true renku-ui=feat/usage-quotas

~I will submit a followup PR to remove the unused views separately.~ Done as part of this PR.


olevski:
TODO: Upload the backed up data from prod about resource usage and test the query with it.
EDIT: This has been done.

olevski:
What changes were made:
- the old views were removed
- the query removes unused pools/users/etc before it does windowing to speed things up, if windowing is done first and then filtering, the windowing/partitions always have to run on all of the data
- the usage is not queried at all for a pool where none of the classes have quota enforcement, this means by default the usage will not be queried at all in prod
- we can turn on usage querying by creating a special pool used only by few users where the classes have quota enforcement

Usage is reported in 2 places:
1. When we list all resource pools -> in this case for pool where none of the classes have enforcement we do not return usage information
2. When we explicitly call api/data/resource_pools/<id>/usage -> in this case when none of the classes of the pool have quota enforcement we return  404, the ui can handle this

eikek (approved):

PR #1341, opened 2026-06-03
feat: add mcp server

rokroskar:
Adds an MCP server. The server runs as a separate container alongside data-services. 

/deploy renku=000-add-mcp-server extra-values=mcpServer.enabled=true

AI use disclosure: The skeleton and usability guidelines were developed and written by me; claude code wrote the code, which I reviewed and tested. 

### Questions about the implementation: 

* there is a safety mechanism to prevent all admin users from using the mcp server. It's implemented as a part of the `_api` wrapper - I'm not sure this is a good way to do it, my initial instinct was to have a decorator do that job. 
* I don't know if you would like to have `DependencyManager` use consistently throughout, here the much simpler `MCPDependencies` is used

### Claude: 

```
$ claude mcp add --transport http --client-id renku-mcp --callback-port 8484 renku https://renku-ci-ds-1341.dev.renku.ch/mcp
```

When you start claude it should automatically try to log you in to the CI deployment. 

### Pi:

Add this as `.pi/mcp.json`:

```json
{
  "mcpServers": {
    "renku": {
      "type": "http",
      "url": "https://renku-ci-ds-1341.dev.renku.ch/mcp",
      "oauth": {
        "clientId": "renku-mcp",
        "redirectUri": "http://localhost:8484"
      }
    }
  }
}
```

Launch Pi and do:

```
> /mcp-auth
```

select renku and connect - it should log you in and get your token. 

### Codex

Add this to `.codex/config.toml`:

```toml
mcp_oauth_callback_port = 8484
mcp_oauth_callback_url = "http://localhost:8484/callback"

[mcp_servers.renku]
url = "https://renku-ci-ds-1341.dev.renku.ch/mcp"
enabled = true

[mcp_servers.renku.oauth]
client_id = "renku-mcp"
```

Run `codex mcp login renku`.

### Using the tools

Just prompt something like 

> Make me a renku project with this repository: https://github.com/rokroskar/renku-demo-air-quality-analysis and a session launcher. 

Related PR: https://github.com/SwissDataScienceCenter/renku/pull/4473


projects/renku_mcp_server/pyproject.toml
```diff
@@ -0,0 +1,30 @@
...
+packages = [
+  { include = "renku_data_services/mcp_api", from = "../../bases" },
+]
+
+[tool.poetry.dependencies]
+python = "^3.13"
+mcp = { extras = ["cli"], version = "^1.0" }
+httpx = "^0.28"
+uvicorn = "^0.34"
+starlette = "^0.46"
```
leafty: To be safe: can we get starlette 1.0.1? For context: https://ostif.org/disclosing-the-badhost-vulnerability-in-starlette/

rokroskar: yes, I was about to check that, thank you!

bases/renku_data_services/mcp_api/dependencies.py
```diff
@@ -0,0 +1,63 @@
...
+
+    base_url: str
+
+    @classmethod
+    def from_env(cls) -> MCPDependencies:
+        """Build MCPDependencies from environment variables."""
+        base_url = os.environ.get("RENKU_BASE_URL", "https://renkulab.io").rstrip("/")
+        return cls(base_url=base_url)
+
+    async def api(
```
leafty, line 25: It's weird to have an `api()` method defined on a dependency manager. It should be defined on a `http_client`. This should be changed to be something like:

```python
  def get_renku_httpx_client() -> httpx.AsyncClient:
    """Returns an async httpx client to use the backend API."""
    return httpx.AsyncClient() # with a token injector.
```

projects/renku_mcp_server/pyproject.toml
```diff
@@ -0,0 +1,30 @@
...
+[tool.poetry.dependencies]
+python = "^3.13"
+mcp = { extras = ["cli"], version = "^1.0" }
+httpx = "^0.28"
+uvicorn = "^0.34"
+starlette = ">=1.0.1"
+pydantic = "^2.10.6"
+
+[tool.poetry.scripts]
+renku-mcp = "renku_data_services.mcp_api.main:main"
```
leafty, line 26: This is not wrong, but unusual for this repository. Python scripts to start a server are just called as the entrypoint.

Instead of:
```
ENTRYPOINT ["tini", "-g", "--", "env/bin/renku-mcp"]
```

we tend to have:
```
ENTRYPOINT ["tini", "-g", "--", "env/bin/python", "-m", "renku_data_services.data_api.main", <other options>]
```

bases/renku_data_services/mcp_api/README.md
```diff
@@ -0,0 +1,134 @@
...
+    "cwd": "/path/to/projects/renku_mcp_server",
+    "env": {
+      "RENKU_BASE_URL": "https://<deployment>"
+    }
+  }
+}
+```
+
+The server discovers your token automatically from the `rnk` CLI token file — no `RENKU_ACCESS_TOKEN`
+needed after `rnk login`.
```
leafty, line 87: Uuhhh can we delay this until there is no more clear text token file?

projects/renku_mcp_server/pyproject.toml
rokroskar: a bit less noise if defined as a script, but sure makes sense to follow the same pattern

bases/renku_data_services/mcp_api/main.py
```diff
@@ -0,0 +1,237 @@
...
+def _rnk_token_paths() -> list[Path]:
+    """Paths where the official rnk CLI stores its token file (platform-specific)."""
+    home = Path.home()
+    paths: list[Path] = []
+    if sys.platform == "darwin":
+        paths.append(home / "Library" / "Application Support" / "io.renku.sdsc.renku-cli" / "token.json")
+    xdg = Path(os.environ.get("XDG_DATA_HOME", str(home / ".local" / "share")))
+    paths.append(xdg / "io.renku.sdsc.renku-cli" / "token.json")
+    if appdata := os.environ.get("APPDATA"):
+        paths.append(Path(appdata) / "io.renku.sdsc.renku-cli" / "token.json")
```
sgaist: Since ` "io.renku.sdsc.renku-cli" / "token.json"` is repeated, what about:
```suggestion
    home = Path.home()
    token_path =  "io.renku.sdsc.renku-cli" / "token.json"
    paths: list[Path] = []
    if sys.platform == "darwin":
        paths.append(home / "Library" / "Application Support" / token_path)
    xdg = Path(os.environ.get("XDG_DATA_HOME", str(home / ".local" / "share")))
    paths.append(xdg / token_path)
    if appdata := os.environ.get("APPDATA"):
        paths.append(Path(appdata) / token_path)
```
?

bases/renku_data_services/mcp_api/main.py
```diff
@@ -0,0 +1,237 @@
...
+        if not path.exists():
+            continue
+        try:
+            data = json.loads(path.read_text())
+            response = data.get("response") or data
+            access = response.get("access_token")
+            if access:
+                return access
+        except (OSError, json.JSONDecodeError, KeyError):
+            continue
```
sgaist, line 71: Since you are using `get`, I think `KeyError` will not happen.

So this could be refactored to:
```suggestion
        try:
            data = json.loads(path.read_text())
        except (OSError, json.JSONDecodeError):
            continue
        response = data.get("response") or data
        access = response.get("access_token")
        if access:
            return access
```

bases/renku_data_services/mcp_api/server.py
```diff
@@ -0,0 +1,957 @@
...
+
+async def _require_non_admin(ctx: Context) -> None:
+    """Raise if the current user is a Renku admin.
+
+    Admin accounts have platform-wide write access that bypasses normal
+    permission checks — running agent operations as an admin is dangerous.
+    Set RENKU_MCP_ALLOW_ADMIN=1 in the server environment to override.
+    """
+
+    if os.environ.get("RENKU_MCP_ALLOW_ADMIN"):
```
sgaist, line 70: Since it's a security related check, I would do an explicit check against `1` as documented in the function docstring.

```suggestion
    if os.environ.get("RENKU_MCP_ALLOW_ADMIN") == "1":
```

bases/renku_data_services/mcp_api/server.py
```diff
@@ -0,0 +1,957 @@
...
+
+def _is_stale_session(s: dict[str, Any]) -> bool:
+    """True if a session has a will_delete_at timestamp already in the past."""
+    wda = s.get("will_delete_at")
+    if not wda:
+        return False
+    try:
+        ts = datetime.datetime.fromisoformat(wda.replace("Z", "+00:00")).timestamp()
+        return ts < time.time()
+    except Exception:
```
sgaist: It's usually frown upon to have such a wide reaching `except` statement.

bases/renku_data_services/mcp_api/dependencies.py
rokroskar: do you mean here a renaming to make it clearer what is what or are you suggesting a deeper refactoring?

sgaist (changes requested):

bases/renku_data_services/mcp_api/README.md
rokroskar: this is only for spinning up the server locally - in practice I expect this to only be used for testing. In production the token is handled by the agent harness via oauth.

bases/renku_data_services/mcp_api/dependencies.py
leafty: 1. No human software engineer would put this `api()` here inside the dependency manager. The dependency manager would have a factory method (e.g. `get_renku_httpx_client()`) which creates a client configured to call the API.

2. The code found here is not necessary, you can achieve most of what it does with just:
    ```python
    headers: dict[str, str] = {
        "Authorization": f"Bearer {token}",
        "Content-Type": "application/json",
        "Accept": "application/json",
    }
    api_client = httpx.AsyncClient(headers=headers)
    ```

leafty: Oh and you can also have:

```python
prefix_url=f"{base_url}/api/data"
httpx.AsyncClient(url=prefix_url, headers=headers)
```

And then you can provide just the path part when calling `.get()`, etc.

bases/renku_data_services/mcp_api/server.py
```diff
@@ -0,0 +1,957 @@
...
+
+
+def _is_stale_session(s: dict[str, Any]) -> bool:
+    """True if a session has a will_delete_at timestamp already in the past."""
+    wda = s.get("will_delete_at")
+    if not wda:
+        return False
+    try:
+        ts = datetime.datetime.fromisoformat(wda.replace("Z", "+00:00")).timestamp()
+        return ts < time.time()
```
leafty: Weird parsing and comparison to "now".

bases/renku_data_services/mcp_api/server.py
```diff
@@ -0,0 +1,957 @@
...
+        return data
+
+    @mcp.tool()
+    async def connector_link(
+        ctx: Context,
+        connector_id: Annotated[str, Field(description="Connector ID")],
+        project_id: Annotated[str, Field(description="Project ID")],
+    ) -> dict[str, Any]:
+        """Link an existing data connector to a project."""
+        return await _deps(ctx).api(
```
SalimKayal, line 471: Why use `_deps(ctx).api` instead of `_api` ? `_api` has `require_non_admin` is there a good reason for this ?

bases/renku_data_services/mcp_api/server.py
```diff
@@ -0,0 +1,957 @@
...
+        query = {"namespace": namespace} if namespace else None
+        return await _api(ctx, "GET", "/projects", query=query)
+
+    @mcp.tool()
+    async def project_get(
+        ctx: Context,
+        project: Annotated[str, Field(description="Project ID or namespace/slug (e.g. 'myuser/my-project')")],
+    ) -> dict[str, Any]:
+        """Get a Renku project by ID or namespace/slug."""
+        return await _api(ctx, "GET", _project_path(project), _token(ctx))
```
SalimKayal, line 285: Is there a good reason to pass `_token` as the `body`  argument of the `_api` function ? also `_api` injects the token in the call. is there any need to pass it there ?

bases/renku_data_services/mcp_api/server.py
```diff
@@ -0,0 +1,957 @@
...
+
+        Before deleting, call session_list(project_id=<id>) for both session types and
+        job_list(project_id=<id>). Then:
+        - Inform the user about the running sessions and pending jobs that will be stopped, and ask for
+          explicit confirmation before proceeding.
+        - Running or pending sessions: stop them with session_delete.
+        - Hibernated or paused sessions: warn the user that unsaved work inside those
+          sessions will be lost, and ask for explicit confirmation before stopping them.
+        """
+        proj = await _api(ctx, "GET", _project_path(project), _token(ctx))
```
SalimKayal, line 323: same as above. is there any need to pass `_token` there ?

bases/renku_data_services/mcp_api/server.py
```diff
@@ -0,0 +1,957 @@
...
+        "port": env.get("port"),
+        "command": env.get("command"),
+        "args": env.get("args"),
+    }
+    return data
+
+
+def _project_path(ident: str) -> str:
+    """Turn a project ID or namespace/slug into an API path segment."""
+    import urllib.parse
```
SalimKayal, line 110: why not import this at the top level ?

bases/renku_data_services/mcp_api/server.py
```diff
@@ -0,0 +1,957 @@
...
+            pass
+    return ""
+
+
+def _deps(ctx: Context) -> MCPDependencies:
+    return ctx.request_context.lifespan_context["deps"]
+
+
+# Cache admin status per token so we only call /user once per session/request.
+_admin_cache: dict[str, bool] = {}
```
SalimKayal, line 59: no ttl for a cache ? don't we risk having this grow too much/become stale ? admittedly there shouldn't be too much admins but still.

bases/renku_data_services/mcp_api/server.py
```diff
@@ -0,0 +1,957 @@
...
+        state = "unknown"
+        logs: Any = None
+        poll = 3.0
+        while time.time() < deadline:
+            session, logs = await asyncio.gather(
+                _api(ctx, "GET", f"/sessions/{session_id}"),
+                _api(ctx, "GET", f"/sessions/{session_id}/logs"),
+                return_exceptions=True,
+            )
+            if isinstance(session, BaseException):
```
SalimKayal, line 938: isn't `BaseException` a bit broad ?

bases/renku_data_services/mcp_api/dependencies.py
```diff
@@ -0,0 +1,63 @@
...
+        params = {k: str(v) for k, v in (query or {}).items() if v is not None}
+        headers: dict[str, str] = {
+            "Authorization": f"Bearer {token}",
+            "Content-Type": "application/json",
+            "Accept": "application/json",
+        }
+        if extra_headers:
+            headers.update(extra_headers)
+
+        async with httpx.AsyncClient() as client:
```
SalimKayal, line 47: this is called every API call. is this really what we want here ?

SalimKayal (changes requested):

PR #1349, opened 2026-06-10
feat: add `pytest-testmon` as a dev dependency for faster testing loop

### LINKED ISSUE #1348: Add `pytest-testmon` to speed up the local test feedback loop
## Summary
Add [`pytest-testmon`](https://testmon.org/) as a Poetry dev dependency. It tracks which code each test executes and, on later runs, re-runs only the tests affected by your changes.

### Motivation
As the workspace grows, running the full suite on every edit gets slow, even though a typical change touches only a few components. Testmon builds a test / source dependency map and runs just the impacted tests. It tightens the local edit to test loop, is dev-only and opt-in (`--testmon`), and changes nothing about how tests are written.

## Scope
- Add `pytest-testmon` to the dev dependency group in `pyproject.toml`.
- Update `poetry.lock`.

SalimKayal:
## Summary
Adds `pytest-testmon` to the dev dependency group and updates the lockfile.

- `pyproject.toml`: add `pytest-testmon` under the dev dependency group.
- `poetry.lock`: regenerated to include `pytest-testmon` and its dependencies.
- `.gitignore` updated to ignore testmon files

## Motivation and Context
Speeds up the test feedback loop by re-running only the tests affected by a given change instead of the whole suite — a good fit for this Polylith workspace, where most changes are localized to a few components/bases. See #1348  for the full rationale.

## Usage
```bash
# Run only tests affected by your current changes
poetry run pytest --testmon
```

The first run is a normal full run that builds the `.testmondata` cache; subsequent runs are scoped to impacted tests.

## Notes
- This is dev-only and fully opt-in (`--testmon`); default `pytest` behavior is unchanged.
- `.testmondata` is a local cache and should not be committed — added to `.gitignore`
- No CI changes in this PR

fixes #1348


pyproject.toml
sgaist: Since you are touching `pyproject.toml`, I would suggest to take the opportunity and update the license field to use the SPDX version of the name `Apache-2.0` . There's a warning about it since some time.
```suggestion
license = "Apache-2.0"
```

sgaist (changes requested):
SGTM

sgaist (approved):

olevski (approved):

PR #1350, opened 2026-06-10
feat: Use an HTTP readiness probe for remote interactive sessions

SalimKayal:
## Summary

This PR makes remote interactive sessions (FirecREST/Slurm and RunAI) report Kubernetes readiness based on whether the session is *actually serving its frontend*, rather than merely on whether the tunnel has connected. It does so by selecting the readiness probe type per session based on both session type and location, sending an `http` probe for remote interactive sessions, `tcp` for local interactive sessions, and `none` for non-interactive sessions.

This change pairs with https://github.com/SwissDataScienceCenter/amalthea/pull/1135, which forwards `readinessProbe`, `port`, and `urlPath` into the remote session controller sidecar so its `/ready` endpoint actually evaluates these probe types.

## Motivation and Context

For remote sessions managed via FirecREST / RunAI, the controller pod was reporting ready immediately as soon as the tunnel connected, even if the IDE on the remote compute node had not actually started yet. As a result, a session could appear "ready" in the UI while the remote process was not yet serving the frontend, causing service routing to send users to a session that isn't usable.

We want the Kubernetes readiness state to reflect real, user-facing session availability so that routing only begins once the session is genuinely reachable. The Amalthea sidecar's `/ready` handler now interprets the probe type; `none` returns 200 immediately, `tcp` dials the local tunnel endpoint, and `http` performs an HTTP GET against the session's port and URL path. This PR is responsible for selecting the correct type per session.

## Changes

- Introduced a `_get_readiness_probe(session_type, session_location)` helper in `core_sessions.py` that centralizes readiness-probe selection:
  - Non-interactive sessions: `Type.none` (no probe), regardless of location.
  - Interactive **remote** sessions (FirecREST/RunAI): `Type.http`.
  - Interactive **local** sessions: `Type.tcp` (the existing default behavior).
- Updated `start_session` to call `_get_readiness_probe(...)` instead of the previous inline `Type.none` / default branch, so probe selection now accounts for session location.
- Added unit tests (`test_core_sessions_readiness_probe.py`) covering all three branches of the helper.
- Added an integration test (`test_start_server_remote_readiness_probe`) that provisions a remote (FirecREST) resource pool, starts a session, and asserts the created session's `readinessProbe.type == Type.http`.

## Notes

- **Depends on https://github.com/SwissDataScienceCenter/amalthea/pull/1135**, without it, the remote controller sidecar will not act on the `http`/`tcp` probe types. That PR requires no CRD changes, since `readinessProbe`, `port`, and `urlPath` already exist on the AmaltheaSession CRD.
- The behavior is **backward compatible**: in Amalthea, an empty/unknown probe type or a port of `0` falls back to the previous behavior of returning `200 OK`, so existing remote sessions continue to behave exactly as before until they're created with these values.
- Local sessions are unaffected in practice; they retain the default TCP probe via Kubernetes container probes.

### DIFF
--- a/components/renku_data_services/notebooks/core_sessions.py
+++ b/components/renku_data_services/notebooks/core_sessions.py
@@ -1127,7 +1127,7 @@ async def start_session(
                 stripURLPath=environment.strip_path_prefix,
                 env=env,
                 remoteSecretRef=remote_secret.ref() if remote_secret else None,
-                readinessProbe=ReadinessProbe(type=Type.none) if session_type.is_non_interactive else ReadinessProbe(),
+                readinessProbe=_get_readiness_probe(session_type, session_location),
             ),
             ingress=ingress_config.get_k8s_ingress() if session_type.is_interactive else None,
             extraContainers=session_extras.containers,
--- a/components/renku_data_services/notebooks/core_sessions.py
+++ b/components/renku_data_services/notebooks/core_sessions.py
@@ -1454,6 +1454,20 @@ async def patch_session(
     return await nb_config.k8s_v2_client.patch_session(session_id, user.id, patch_serialized)
 
 
+def _get_readiness_probe(session_type: SessionType, session_location: SessionLocation) -> ReadinessProbe:
+    """Return the readiness probe for a session based on type and location.
+
+    - Non-interactive sessions get no probe (Type.none).
+    - Interactive remote sessions (firecrest/runai) use an HTTP probe.
+    - Interactive local sessions use the default TCP probe.
+    """
+    if session_type.is_non_interactive:
+        return ReadinessProbe(type=Type.none)
+    if session_location == SessionLocation.remote:
+        return ReadinessProbe(type=Type.http)
+    return ReadinessProbe(type=Type.tcp)
+
+
 class _NamedResource(Protocol):
     """Represents a resource with a name."""
 
--- a/test/bases/renku_data_services/data_api/test_notebooks.py
+++ b/test/bases/renku_data_services/data_api/test_notebooks.py
@@ -329,6 +329,7 @@ async def test_not_found_server(
 
     assert res.status_code == 404, res.text
 
+
 @pytest.mark.xdist_group("sessions")  # Needs to run on the same worker as the rest of the sessions tests
 @pytest.mark.sessions
 @pytest.mark.asyncio
--- a/test/bases/renku_data_services/data_api/test_notebooks.py
+++ b/test/bases/renku_data_services/data_api/test_notebooks.py
@@ -12,7 +12,13 @@ from sanic_testing.testing import SanicASGITestClient
 
 from renku_data_services.base_models.core import APIUser
 from renku_data_services.data_api.dependencies import DependencyManager
-from renku_data_services.notebooks.crs import AmaltheaSessionSpec, AmaltheaSessionV1Alpha1, Resources, Session
+from renku_data_services.notebooks.crs import (
+    AmaltheaSessionSpec,
+    AmaltheaSessionV1Alpha1,
+    Resources,
+    Session,
+    Type,
+)
 
 
 @pytest.fixture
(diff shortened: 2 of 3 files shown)


sgaist (approved):

components/renku_data_services/notebooks/core_sessions.py
```diff
@@ -1454,6 +1454,20 @@ async def patch_session(
...
+    """Return the readiness probe for a session based on type and location.
+
+    - Non-interactive sessions get no probe (Type.none).
+    - Interactive remote sessions (firecrest/runai) use an HTTP probe.
+    - Interactive local sessions use the default TCP probe.
+    """
+    if session_type.is_non_interactive:
+        return ReadinessProbe(type=Type.none)
+    if session_location == SessionLocation.remote:
+        return ReadinessProbe(type=Type.http)
```
leafty, line 1467: For context: `http` probes are really not happy when the frontend is RStudio. Even with no authn, RStudio will redirect `/` to `/?some_token_something` and then back to `/`. The issue is that k8s doesn't take a 3xx redirect as a positive readiness check.

So if the user tries to run RStudio with a remote session, the session will never pass the readiness gate.

leafty (changes requested):

components/renku_data_services/notebooks/core_sessions.py
sgaist: It should based one their [documentation for the http probe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-liveness-http-request):
```
Any code greater than or equal to 200 and less than 400 indicates success. Any other code indicates failure.
``` 

Both readiness and liveness should behave the same in that regard.

leafty: Oh sorry, the initial redirect is with a 401 or 403 code from rstudio.

SalimKayal: Hm, then I have to figure something out for this... Any preliminary idea?

leafty: This is the context I remember, I think it's possible that the current rstudio images (e.g. with buildpacks) have a better behavior. Trying out the rocker images, they may work for http probes.

If we switch to http probes, the we should potentially switch for all sessions with and have an option available on the launcher to downgrade to TCP for not-nice frontends.

leafty (commented):
Looking at this again: I think this is OK for now, as this can help getting a better readiness probe. But it is not good to have the probe depend on the `SessionLocation` in the long term.

We should have a folllow-up change to set the probe to `http` in all cases, and have a new field `readinessProbeType` in the `EnvironmentORM` so that we can revert back to `tcp` in corner cases.

components/renku_data_services/notebooks/core_sessions.py
SalimKayal: there would be this https://docs.posit.co/ide/server-pro/admin/auditing_and_monitoring/server_health_checks.html but the problem is that we do not know which front end is running and which health check endpoint exists.

perhaps instead of this I could serve a health endpoint on the tunnel sidecar and probe on it ? this would allow us to "accept" 4xx as proof of readiness

leafty: When I probe by hand, I get redirected to `/browser_unsupported.htm` which gives a 200. So maybe it can work for now with RStudio. Also, I think having RStudio potentially broken for CSCS sessions for a limited time is acceptable.

SalimKayal:
> Looking at this again: I think this is OK for now, as this can help getting a better readiness probe. But it is not good to have the probe depend on the `SessionLocation` in the long term.

I could change that right now but it would need more extensive testing with the existing sessions.

> We should have a folllow-up change to set the probe to `http` in all cases, and have a new field `readinessProbeType` in the `EnvironmentORM` so that we can revert back to `tcp` in corner cases.

I don't like the idea of needing an extra piece of information but I don't see how to make it simpler. Perhaps we could add a health probe Buildpack to unify things at least on our side ?

leafty (approved):

Issue #1354, opened 2026-06-16
Refactor handling of default values in configuration objects


sgaist:
## Summary

`renku-data-services` uses multiple configuration objects for various aspects and features and some of them have default values.

Currently, most of these default values are used at the "point of use" where the configuration is required. This can be a bit surprising as reading the implementation of the configuration shows one reality (i.e. configuration is set to None because no environment variable was set) VS something is happening (i.e. builds using the default value for the build strategy although the configuration code does not indicate that).

While these defaults do make sense, having them handled in two separate places makes things harder to reason about and, sometimes, tests.

## Suggestion

Move the default value handling into the configuration classes and maybe make them [frozen](https://docs.python.org/3/library/dataclasses.html#frozen-instances) as I don't think they are meant to be changed at run time.

### Example

```python
@dataclass
class BuildsConfig:
    """Configuration for container image builds."""

    build_output_image_prefix: str

    @classmethod
    def from_env(cls) -> "BuildsConfig":
        """Create a config from environment variables."""
        build_builder_image = os.environ.get("BUILD_BUILDER_IMAGE", constants.BUILD_DEFAULT_BUILDER_IMAGE)
```

This would avoid having:

```python
        builder_image = self.builds_config.build_builder_image or constants.BUILD_DEFAULT_BUILDER_IMAGE
```
in the `db.py` file

Since this is a setting that is technically mandatory we could even have:

```python
@dataclass
class BuildsConfig:
    """Configuration for container image builds."""

    build_output_image_prefix: str

    @classmethod
    def from_env(cls) -> "BuildsConfig":
        """Create a config from environment variables."""
        build_builder_image = os.environ.get("BUILD_BUILDER_IMAGE", constants.BUILD_DEFAULT_BUILDER_IMAGE) or constants.BUILD_DEFAULT_BUILDER_IMAGE
```
So even in case of empty environment variable we would be good to go.

Since that would be a repeating pattern:

```python

def get_with_default(haystack: dict[str, str], needle: str, default: str) -> str:
    """Returns the default value if needle is not found or the value is None/empty."""
    return haystack.get(needle, default) or default


@dataclass
class BuildsConfig:
    """Configuration for container image builds."""

    build_output_image_prefix: str

    @classmethod
    def from_env(cls) -> "BuildsConfig":
        """Create a config from environment variables."""
[...]

PR #1355, opened 2026-06-17
feat: add feature flag for private repo builds

sgaist:
## Describe your changes

This PR moves the private repository build handling behind a feature flag.

This allows the Renku admins to provide the feature only if wanted alongside the build feature.

In order to make the code simpler, some of the configuration variable default value handling has been moved to the configuration class.

## Tickets and links

Related to #1333


components/renku_data_services/session/config.py
```diff
@@ -25,21 +25,28 @@ class BuildPlatformOverrides:
     tolerations: list[session_crs.Toleration] | None = None
 
 
+def get_environ_with_default(needle: str, default: str) -> str:
+    """Returns the default value if needle is not found or value is none/empty."""
+
+    return os.environ.get(needle, default) or default
```
leafty: How is this different from just using `.get(key, default)` on a dict-like object?

sgaist: `get` will return `default` if `key` is not in the dict however, it will return whatever value including `None`and `""`, if `key` is present. It is not what we want in this case.

Using that function, we ensure that we get `default` in all cases where we could have `None` or an empty string.

You can see the simplification in `db.py`. Albeit lightweight, it makes things cleaner: since we will use a default value down the line, let's have it directly in the configuration object from the start.

This relates to #1354

sgaist: It also has the benefit of not requiring None checks every time we want to use these configuration values.

leafty: > `get` will return `default` if `key` is not in the dict however, it will return whatever value including `None`and `""`, if `key` is present. It is not what we want in this case.
> 
> Using that function, we ensure that we get `default` in all cases where we could have `None` or an empty string.

Then, the correct form is `os.environ.get(key) or default_value`:
* if `key` is present and the `value` is truth-y -> use `value`
* if `key` is present and the `value` is not truth-y (e.g. `""` or `None`) -> use `default_value`
* if `key` is not present -> use `default_value`

No need to make the person reading the code jump to another method.

sgaist: Mind you, I had simplified that function before as I wanted to provide something generic for dict like objects but since it was only used on `os.environ`... 

And then I just missed the simple path...

Fixed

components/renku_data_services/session/config.py
```diff
@@ -50,14 +51,16 @@ class BuildsConfig:
...
+        private_builds_enabled = os.environ.get("BUILD_PRIVATE_REPO_BUILDS_ENABLED", "false").lower() == "true"
+
+        build_output_image_prefix = (
+            os.environ.get("BUILD_OUTPUT_IMAGE_PREFIX") or constants.BUILD_DEFAULT_OUTPUT_IMAGE_PREFIX
+        )
+        build_output_private_image_prefix = (
+            os.environ.get("BUILD_OUTPUT_PRIVATE_IMAGE_PREFIX") or constants.BUILD_DEFAULT_OUTPUT_PRIVATE_IMAGE_PREFIX
+        )
+
+        if private_builds_enabled and build_output_image_prefix == build_output_private_image_prefix:
```
leafty: Note: technically we should also check that `not build_output_image_prefix.startswith(build_output_private_image_prefix) and not build_output_private_image_prefix.startswith(build_output_image_prefix)` (i.e. one should not be the prefix of the other, because this will most likely result in permissions being mixed up, e.g. `renku.io/builds` and `renku.io/builds/private`).

leafty (dismissed):

components/renku_data_services/session/config.py
sgaist: Good point, I did not thought about that one.

Added along with tests

leafty (approved):

PR #1356, opened 2026-06-18
add apps launcher handling

wesjdj:
/deploy renku=add-apps-beta extra-values=apps.enabled=true


components/renku_data_services/renku_apps/repository.py
```diff
@@ -93,6 +94,15 @@ async def delete_app(self, user: base_models.APIUser, app_name: str) -> None:
...
         return None
 
+    async def delete_app_for_launcher(self, user: base_models.APIUser, launcher: SessionLauncher) -> None:
+        """Delete the app deployment backing the given launcher, if one exists."""
+        project = await self.project_repo.get_project(user, launcher.project_id)
+        runtime_state = await self.k8s_client.get_app_deployment_for_project(project, launcher)
+        if runtime_state is None:
+            return None
+        await self.delete_app(user, runtime_state.name)
+        return None
```
olevski: We should be able to use labels in k8s so that we can retrieve the app based on the launcher id. Or alternatively we can save the application name from k8s into the launcher table. But the launcher table is already getting complicated. So I would go towarrds finding the app with the launcher labels. That way you dont have to find the project or the launcher.

components/renku_data_services/project/blueprints.py
```diff
@@ -293,6 +293,12 @@ async def _patch(
             _: Request, user: base_models.APIUser, project_id: ULID, body: apispec.ProjectPatch, etag: str
         ) -> JSONResponse:
             project_patch = validate_project_patch(body)
+            if project_patch.visibility == Visibility.PRIVATE and await self.session_repo.project_has_app_launchers(
+                user, project_id
+            ):
+                raise errors.ValidationError(
+                    message="This project cannot be made private because it has one or more app launchers."
+                )
```
olevski: comment(non-blocking): This is fine. Just not sure what the desired behaviour is. but if we need to change the behaviour I propose we do it in a follow up PR. I asked for confirmation in slack here: https://swiss-data-science.slack.com/archives/C0B1W8RBQBX/p1782737951990709

components/renku_data_services/renku_apps/repository.py
olevski: You can also try to find it based on reconstructing the name from the project and launcher and if that fails then fall back to the labels.

olevski (changes requested):
One small thing. But if you prefer we can address everything in future/follow up PRs.

components/renku_data_services/project/blueprints.py
olevski: Got confirmation that the desired behaviour is to not require the user to delete the app. But the backend should turn off the app and then prevent the app from being restarted as long as the project is private. But lets come back to this in a followup PR.

olevski: Or if you really want to address it here - then go ahead.

wesjdj: Done, thanks!

components/renku_data_services/renku_apps/repository.py
wesjdj: Makes sense, thanks - done

olevski (approved):

PR #1359, opened 2026-06-22
feat(BuildStrategy): use flag to activate insecure registries

sgaist:
## Describe your changes

The current implementation of "insecure registries" support for builds is automatically activated by default.

This PR changes that through the use of the `shp-output-insecure` output parameter from Shipwright.

Changes are twofold:

- Update renku-data-services to use the new `BUILD_INSECURE_OUTPUT_ENABLED` environement variable
- Update the BuildStrategy to make use of it
- The list of insecure registries is now a proper list

This will also follow the Shipwright implementation more accurately.


olevski:
@sgaist what is the point of this? What does "insecure registries" mean exactly and what is it used for?

leafty:
> @sgaist what is the point of this? What does "insecure registries" mean exactly and what is it used for?

@olevski 

This tells the buildpack lifecycle to not verify TLS certificates for specific registries (when pulling, pushing, etc.). See also conversation here: https://github.com/SwissDataScienceCenter/renku-data-services/pull/1264#pullrequestreview-4126735483

This setting can help test building images end-to-end where the registry is deployed in a non-production setup with certificates not attached to a PKI (self-signed).

components/renku_pack_builder/manifests/buildstrategy.yaml
leafty, line 1: The script part looks complex now, with argument parsing involved. Do we want to maintain this?

sgaist: The idea was to keep it close to the Shipwright implementation however I am not against keeping it simpler if you prefer.

We can move back to a single string containing the list of registries.

components/renku_data_services/session/config.py
```diff
@@ -36,6 +36,7 @@ class BuildsConfig:
     build_builder_image: str | None = None
     build_run_image: str | None = None
     build_strategy_name: str | None = None
+    build_insecure_ouput_enabled: bool = False
```
olevski:
```suggestion
    build_insecure_output_enabled: bool = False
```

And then this repeats in a few other places I did not flag.

components/renku_data_services/session/k8s_client.py
```diff
@@ -223,10 +223,16 @@ async def create_image_build(self, params: models.ShipwrightBuildRunParams, user
                             crs.ParamValue(name="frontend", value=params.frontend),
                             crs.ParamValue(name="run-image", value=params.run_image),
                             crs.ParamValue(name="builder-image", value=params.builder_image),
-                            crs.ParamValue(name="insecure-registries", value=params.insecure_registries),
+                            crs.ParamValue(
+                                name="insecure-registries",
+                                values=[
+                                    crs.Value(value=registry) for registry in params.insecure_registries.split(",")
+                                ],
+                            ),
```
olevski: question: `"".split(",")` returns `[""]` will this be a problem?

components/renku_pack_builder/manifests/buildstrategy.yaml
```diff
@@ -37,11 +36,45 @@ spec:
...
+            fi
+
+            if [[ ! -z "$insecureRegistries" ]]; then
+              echo "> Using insecure registries: $insecureRegistries"
+
+              export CNB_INSECURE_REGISTRIES="${insecureRegistries}"
+            fi
+          fi
+
           set -euo pipefail
```
olevski: question: It is intentional that all of the new code runs before `set -euo pipefail`? And if yes, why?

components/renku_data_services/session/k8s_client.py
olevski: I think this will cause trouble further on into the strategy in the bash script in the yaml file.

olevski (changes requested):

components/renku_pack_builder/manifests/buildstrategy.yaml
sgaist: I had memory of things like exporting `CNB_INSECURE_REGISTRIES` failing but after retesting it's working properly...

Will move it up.

components/renku_data_services/session/k8s_client.py
sgaist: It should not however the BuildStrategy does not handle properly an empty value. Not that it fails but it will trigger the export of `CNB_INSECURE_REGISTRIES` with `,` as value.

I have refactored that part back to its original implementation as I have simplified the BuildStrategy as well.

components/renku_data_services/session/config.py
sgaist: Worse is, I remember spotting that issue but it looks I stashed it as some point.

Fixed

components/renku_pack_builder/manifests/buildstrategy.yaml
sgaist: I simplified it

olevski (approved):

Issue #1160, opened 2026-01-07
Projects with dashes in their name not found via generic search


eikek:
A project that has a dash in its name can be found when searched specifically for the name as in `name:my-project`, but it is not found when using generic search `my-project`.

olevski:
closed by #1362

PR #1362, opened 2026-06-25
fix: solr splitting project name

### LINKED ISSUE #1217: Searching for a project with dash in the name does not work
I have a public project called `ollama-test` on renkulab.io

https://renkulab.io/p/tasko.olevski/ollama-test

But when I search for `ollama-test` in the search I get nothing:

Image

Searching for just `ollama` returns results.

Image

Seen on renku 2.14.0

olevski:
/deploy

This fixes the basic problem of:
- create a project with name "test-project"
- type `test-project` in the search 
- there will be no results

It is because the name is tokenized in a way that splits on `-`, `_` and other characters.

What this does is:
- adds a new field which retains the name unchanged - other than lowercased
- add a migration so the new field is there
- modify the query for solr so that some matches significantly boost the match score:
  - fuzzy match on the name
  - full match on the unchanged name
  - full match on the slug (which is also not tokenized)

### DIFF
--- a/components/renku_data_services/search/core.py
+++ b/components/renku_data_services/search/core.py
@@ -49,13 +49,12 @@ async def update_solr(
     search_updates_repo: SearchUpdatesRepo,
     solr_client: SolrClient,
     batch_size: int,
-    solr_config: SolrClientConfig,
     schema_version: Literal["latest"] | int = "latest",
 ) -> list[Exception]:
     """Selects entries from the search staging table and updates SOLR."""
     counter = 0
     output: list[Exception] = []
-    migrator = SchemaMigrator(solr_config)
+    migrator = SchemaMigrator(solr_client.config)
     while True:
         schema_ready = await migrator.schema_version_is(schema_version)
         if not schema_ready:
@@ -106,7 +105,7 @@ async def _renku_query(
     authz_client: AuthzClient, ctx: Context, uq: SolrUserQuery, limit: int, offset: int
 ) -> SolrQuery:
     """Create the final solr query embedding the given user query."""
-    logger.debug(f"Searching as user: {ctx.role or "anonymous"}")
+    logger.debug(f"Searching as user: {ctx.role or 'anonymous'}")
     role_constraint: list[str] = [st.public_only()]
     match ctx.role:
         case AdminRole():
--- a/components/renku_data_services/search/core.py
+++ b/components/renku_data_services/search/core.py
@@ -39,17 +39,23 @@ from renku_data_services.solr.solr_client import (
     SolrQuery,
     SubQuery,
 )
+from renku_data_services.solr.solr_migrate import SchemaMigrator
 
 logger = logging.getLogger(__name__)
 
 
 async def update_solr(
-    search_updates_repo: SearchUpdatesRepo, solr_client: SolrClient, batch_size: int
+    search_updates_repo: SearchUpdatesRepo, solr_client: SolrClient, batch_size: int, solr_config: SolrClientConfig
 ) -> list[Exception]:
     """Selects entries from the search staging table and updates SOLR."""
     counter = 0
     output: list[Exception] = []
+    migrator = SchemaMigrator(solr_config)
     while True:
+        schema_ready = await migrator.schema_is_ready()
+        if not schema_ready:
+            await asyncio.sleep(2)
+            continue
         entries = await search_updates_repo.select_next(batch_size)
         if entries == []:
             break
--- a/components/renku_data_services/search/core.py
+++ b/components/renku_data_services/search/core.py
@@ -3,7 +3,6 @@
 import asyncio
 from collections.abc import Iterable
 from datetime import UTC, datetime
-from typing import Literal
 
 from authzed.api.v1 import (
     AsyncClient as AuthzClient,
@@ -40,7 +39,6 @@ from renku_data_services.solr.solr_client import (
     SolrQuery,
     SubQuery,
 )
-from renku_data_services.solr.solr_migrate import SchemaMigrator
 
 logger = logging.getLogger(__name__)
 
@@ -49,17 +47,11 @@ async def update_solr(
     search_updates_repo: SearchUpdatesRepo,
     solr_client: SolrClient,
     batch_size: int,
-    schema_version: Literal["latest"] | int = "latest",
 ) -> list[Exception]:
     """Selects entries from the search staging table and updates SOLR."""
     counter = 0
     output: list[Exception] = []
-    migrator = SchemaMigrator(solr_client.config)
     while True:
-        schema_ready = await migrator.schema_version_is(schema_version)
-        if not schema_ready:
-            await asyncio.sleep(2)
-            continue
         entries = await search_updates_repo.select_next(batch_size)
         if entries == []:
             break
--- a/components/renku_data_services/search/core.py
+++ b/components/renku_data_services/search/core.py
@@ -3,6 +3,7 @@
 import asyncio
 from collections.abc import Iterable
 from datetime import UTC, datetime
+from typing import Literal
 
 from authzed.api.v1 import (
     AsyncClient as AuthzClient,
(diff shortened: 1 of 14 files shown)


olevski:
Also I found a weird race condition. Basically what happens is:
- We use some default configuration in solr which makes it accept docs even if the schema is not set
- When such documents are received their value is always a list of something (i.e. strings or ints or whatever)
- So in CI deployments I think that the accetpance tests start very quickly before the schema migrations for solr complete
- Which means that some things are inserted in solr before the solr schema migrations are done
- Then some docs have fields list 'list[str]' but our models except `str` and all hell breaks loose
- This is because the acceptance tests start before the schema migrations are done and create projects for example, but some properties of a project are not in the solr schema yet so their type is `list[str]` rather than `str`. And the schema migrations do not replace the type - they just check if the field exists

That is why there is a guard to make sure no solr updates occur when we need to run a solr schema migration.

We can also change the migration code to fix this drift in the types. But this is more complicated.

components/renku_data_services/search/core.py
```diff
@@ -39,17 +40,27 @@
...
 
 logger = logging.getLogger(__name__)
 
 
 async def update_solr(
-    search_updates_repo: SearchUpdatesRepo, solr_client: SolrClient, batch_size: int
+    search_updates_repo: SearchUpdatesRepo,
+    solr_client: SolrClient,
+    batch_size: int,
+    solr_config: SolrClientConfig,
```
eikek: just a thought: We could also make the config available from the client itself. So there is less chance of injecting a different config than used by the client.

components/renku_data_services/search/core.py
```diff
@@ -39,17 +40,27 @@
...
 ) -> list[Exception]:
     """Selects entries from the search staging table and updates SOLR."""
     counter = 0
     output: list[Exception] = []
+    migrator = SchemaMigrator(solr_config)
     while True:
+        schema_ready = await migrator.schema_version_is(schema_version)
+        if not schema_ready:
+            await asyncio.sleep(2)
+            continue
```
eikek: There is a similar implementation in reprovision.py. I'm thinking whether the whole "wait for schema" could be somewhere else perhaps? (like in some startup routine for data-tasks) Is this necessary here, because the data-tasks doesn't wait for data-services to do the migration?

When calling `migrator.migrate()` it already checks for a running migration and would return `MigrateResult.empty()` in that case, otherwise it proceeds applying the migrations. Perhaps the complete "wait for schema" logic could be in the migrator itself? The `migrate` method could even be adopted to just wait in case it is running already. For example, rename `migrate` to `migrate0` and implementing `migrate` to wait for the migration to be done if `migrate0` returns an empty result? This would make logic like below `if not migrate_solr_schema:` unnecessary.

components/renku_data_services/search/core.py
```diff
@@ -39,17 +40,27 @@
...
 logger = logging.getLogger(__name__)
 
 
 async def update_solr(
-    search_updates_repo: SearchUpdatesRepo, solr_client: SolrClient, batch_size: int
+    search_updates_repo: SearchUpdatesRepo,
+    solr_client: SolrClient,
+    batch_size: int,
+    solr_config: SolrClientConfig,
+    schema_version: Literal["latest"] | int = "latest",
```
eikek: Why would one want to pass in a specific version? when doing solr updates, shouldn't it better go always to the latest?

components/renku_data_services/search/reprovision.py
```diff
@@ -78,7 +99,11 @@ async def _get_all_data_connectors(
                 yield dc
 
     async def init_reprovision(
-        self, admin: APIUser, reprovisioning: Reprovisioning, migrate_solr_schema: bool = True
+        self,
+        admin: APIUser,
+        reprovisioning: Reprovisioning,
+        migrate_solr_schema: bool = True,
+        schema_version: Literal["latest"] | int = "latest",
```
eikek, line 106: Same question for me as above: why is this argument necessary?

components/renku_data_services/solr/entity_schema.py
```diff
@@ -85,6 +86,38 @@ class Analyzers:
...
     )
 
+    # Analyzers for the `name` field only. Unlike `text_index`, these use the
+    # whitespace tokenizer so hyphenated values stay intact long enough for the
+    # wordDelimiterGraph filter to split them. That filter then emits the word
+    # parts ("test", "project"), the catenated form ("testproject") and -- via
+    # preserveOriginal -- the whole token ("test-project"), so the name field
+    # matches all of those (including the hyphenated whole, which the fuzzy query
+    # operator can only hit because preserveOriginal keeps it in the index).
+    name_index: Final[Analyzer] = Analyzer(
```
eikek, line 96: Great!

eikek (commented):
I was wondering if we could make a integration test in `bases/**/test_search.py` for that case?

components/renku_data_services/search/core.py
olevski: It is only needed for testing Eike. We have tests that go to a specific schema version (not the latest) then they inject some data and then go to the latest version. So we have to handle not always going to the latest version.

olevski: Specifically these tests https://github.com/SwissDataScienceCenter/renku-data-services/blob/fix-solr-splitting-words/test/bases/renku_data_services/data_api/test_search_migrations.py#L92

components/renku_data_services/search/core.py
olevski: Done in https://github.com/SwissDataScienceCenter/renku-data-services/pull/1362/commits/16d8de2e91415cdf61647f27e667826657bdd89c

components/renku_data_services/search/core.py
olevski: So I added the waiting for a solr schema version in the regular migrate call.

olevski: I think there is no need to have migrate0 and migrate.

components/renku_data_services/search/core.py
olevski: Now there is no version passed to core.update_solr.

components/renku_data_services/search/reprovision.py
olevski: This is for the tests at https://github.com/SwissDataScienceCenter/renku-data-services/blob/fix-solr-splitting-words/test/bases/renku_data_services/data_api/test_search_migrations.py#L92

olevski:
I also added a test to make sure an exact name match will rank the result at the top.

components/renku_data_services/search/reprovision.py
```diff
@@ -96,6 +121,9 @@ def log_counter(c: int) -> None:
             if c % 50 == 0:
                 logger.info(f"Inserted {c}. entities into staging table...")
 
+        if not migrate_solr_schema:
+            await self._wait_schema_version(schema_version)
```
eikek, line 125: Do we still need this? If `migrate` is waiting, wouldn't it be enough to only call `migrator.migrate()`? (or is this again for tests)

eikek (approved):

components/renku_data_services/search/reprovision.py
olevski: It is not for the tests but I would keep it just in case... Because if the migrate_solr_schema flag is to false, then you will simply run reprovision without migrating the schema or waiting for the schema version at all. So without this flag an admin could start a reprovisioning with outstanding solr schema migrations and win the race against the migrations. And if this happens the solr schema will be broken.

olevski: Oh nvm it is also still needed for the tests. Yes. It is hard to follow references from the LSP through the test fixtures.

olevski:
Also I just saw this PR: https://github.com/SwissDataScienceCenter/renku-data-services/pull/1162

These two are somewhat related. I will go ahead and merge this one and then we can check the older PR.

olevski:
Also this will close #1160 but github will not allow me to link this for some reason.

PR #1367, opened 2026-06-29
fix: allow unsetting quota

### LINKED ISSUE #1365: Removing quotas not possible
To reproduce:
- creata a resource pool
- there is no way to remote the quota from the admin panel (but that is ok)
- go to the swagger page, send request to patch the resource pool with `quota: null`
- the quota is still there

Please also document in the apispec how can the quota be removed.

I think the payload validation is ignoring or removing the quota field when set to None during validation

leafty:
Closes #1365.

Updated the API spec description (need to use `PUT` to remove a quota).

We can add and remove quotas with swagger:

1. Use `GET /resource_pools/{id}` to get the current json representation of a pool
2. Paste the json into the `PUT` body, remove the `"id"` field and set `"quota": null`.
3. Create a new quota with the `PATCH` request

/deploy


SalimKayal (commented):
otherwise seems pretty forward as a change.

components/renku_data_services/crc/api.spec.yaml
```diff
@@ -303,6 +303,9 @@ paths:
         - resource_pools
     patch:
       summary: Update specific fields of an existing resource pool
+      description: |
+        Note that the quota cannot be removed using the PATCH method.
+        To remove the quota, use the PUT method and unset the quota field.
```
SalimKayal, line 308: for my personal culture, why this choice ?

leafty: Because our code base cannot see the difference between a field set to null and a field not present in the payload. We could do it, but it would require some new type of validation we don't have yet.

SalimKayal: ok makes sense. thanks

SalimKayal (approved):
LGTM

PR #1371, opened 2026-07-02
feat: add support for inferred-frontend and no-implicit-frontend

SalimKayal:


leafty (commented):
Looks reasonable, will have a look at the deployment with the updated UI.

leafty (approved):

PR #1368, opened 2026-06-29
fix: dispatch metrics by k8s kind

leafty:
Closes #915.

/deploy extra-values=posthog.enabled=true,posthog.apiKey=phc_7hCT82FJG1oPBUN43wDyU93uSyymE0Iwo2fW4qb02pa,posthog.environment=renku-ci-ds-1368,posthog.host=https://eu.i.posthog.com,dataService.imageBuilders.enabled=true,dataService.imageBuilders.strategyName=renku-buildpacks-v3,dataService.imageBuilders.outputImagePrefix=harbor.dev.renku.ch/renku-build/,dataService.imageBuilders.nodeSelector.renku.io/node-purpose=user,dataService.imageBuilders.tolerations[0].effect=NoSchedule,dataService.imageBuilders.tolerations[0].key=renku.io/dedicated,dataService.imageBuilders.tolerations[0].operator=Equal,dataService.imageBuilders.tolerations[0].value=user

PR #1372, opened 2026-07-06
feat: add session type to some metrics

leafty:
Update session metrics to include a new `session_type` field used to differentiate interactive sessions and offline jobs.

Also, fixes a bug introduced in #1368 where session events were not sent because of case-sensitive matching with the `GVK` dataclass.

/deploy #notest renku-ui=main extra-values=posthog.enabled=true,posthog.apiKey=phc_7hCT82FJG1oPBUN43wDyU93uSyymE0Iwo2fW4qb02pa,posthog.environment=renku-ci-ds-1372,posthog.host=https://eu.i.posthog.com

### DIFF
--- a/components/renku_data_services/k8s/models.py
+++ b/components/renku_data_services/k8s/models.py
@@ -405,6 +405,16 @@ class GVK:
             version=version,
         )
 
+    def __eq__(self, value: object) -> bool:
+        """Override equality to be case-insensitive."""
+        if not isinstance(value, GVK):
+            return False
+        return (
+            self.kind.casefold() == value.kind.casefold()
+            and self.version.casefold() == value.version.casefold()
+            and (self.group.casefold() if self.group else None) == (value.group.casefold() if value.group else None)
+        )
+
 
 @dataclass
 class APIObjectInCluster:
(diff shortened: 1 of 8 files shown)


components/renku_data_services/k8s/models.py
```diff
@@ -405,6 +405,20 @@ def from_kr8s_object(cls, kr8s_obj: type[APIObject] | APIObject) -> Self:
             version=version,
         )
 
+    def __eq__(self, value: object) -> bool:
```
sgaist, line 408: Just curiosity, what is the motivation of making `__eq__` and `__hash__` case insensitive ?

sgaist (approved):

components/renku_data_services/k8s/models.py
leafty: The CRD names are not case sensitive, so `AmaltheaSession` is equivalent to `amaltheasession`. This is important with https://github.com/SwissDataScienceCenter/renku-data-services/blob/ff28f85831f5bdbeadb45415d86124766d445bdf/components/renku_data_services/notebooks/constants.py#L7-L8

There is an event filter where we match `gvk == AMALTHEA_SESSION_GVK` and this match needs to be case insensitive.

sgaist: Ok, thanks !

Issue #1298, opened 2026-05-05
Refactor resource pool user change to use a single db transaction


leafty:
TODO: handle authz changes as a single database transaction in `update_user_resource_pools`

 _Originally posted by @leafty in [#1290](https://github.com/SwissDataScienceCenter/renku-data-services/pull/1290/changes#r3188218267)_

cc @SalimKayal

PR #1374, opened 2026-07-06
Fix: reuse DB session in `get_resource_pool_users` to avoid connection-pool deadlock

### LINKED ISSUE #1373: TimeoutError: QueuePool limit of size 10 overflow 0 reached, connection timed out, timeout 30.00 (Background on this error at: https://sqlalche.me/e/20/3o7r)
Sentry Issue: [RENKU-DATA-SERVICES-8P](https://renku.sentry.io/issues/99224942/?referrer=github_integration)

```
TimeoutError: QueuePool limit of size 10 overflow 0 reached, connection timed out, timeout 30.00 (Background on this error at: https://sqlalche.me/e/20/3o7r)
(18 additional frame(s) were not displayed)
...
  File "<string>", line 2, in _connection_for_bind
    # Copyright (C) 2005-2025 the SQLAlchemy authors and contributors
```

SalimKayal:
## Summary

`GET /resource_pools/{resource_pool_id}/users` could exhaust the SQLAlchemy connection pool and deadlock.

`get_resource_pool_users()` opened a DB session and then, while still holding that session, called `kc_user_repo.get_user()` once per user. Each `get_user()` call opened a *new* DB session. If the pool was already at its limit, every request thread waited for an additional connection that could never be released, causing:

```
TimeoutError: QueuePool limit of size 10 overflow 0 reached, connection timed out, timeout 30.00
```

## Changes

- Wrapped `MemberRepository.get_resource_pool_users()` with `@with_db_transaction` so the method can reuse a single session.
- Added a `session` parameter to `UserRepo.get_user()` and decorated it with `@with_db_transaction`.
- Added `UserRepo.get_users_by_ids()` to fetch multiple users in a single batched query on the same session.
- Replaced the per-user `kc_user_repo.get_user(id=uid)` loop with a single call to `kc_user_repo.get_users_by_ids(user_ids, session=session)`.
- Updated `update_resource_pool_users()` to pass its own session to `get_resource_pool_users()`.
- Added a regression test verifying that listing resource-pool users does not issue per-user `get_user()` calls.

## Result

The endpoint now resolves all user data on the same DB session, eliminating the nested-session pattern that starved the connection pool.

fixes #1373

/deploy

### DIFF
--- a/components/renku_data_services/crc/db.py
+++ b/components/renku_data_services/crc/db.py
@@ -1053,6 +1053,7 @@ class MemberRepository(_Base):
         self.group_repo = group_repo
         self.project_repo = project_repo
 
+    @with_db_transaction
     @_only_admins
     async def get_resource_pool_users(
         self,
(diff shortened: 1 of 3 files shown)


SalimKayal:
this PR obviously does not fix

> Related to the comment here -> https://github.com/SwissDataScienceCenter/renku-data-services/issues/1298 but not the same: to fix the deadlock, we REALLY REALLY need to make sure that no DB session is requested from an existing DB session and this should be enforced by creating at most one DB session per API request.
> 
> This means that we should move away and forbid the coding paradigm where methods in `db.py` create their own DB session. DB sessions MUST be created from the `blueprint.py` handling method and passed around until closed when the API request is fulfilled.

as it requires a much deeper refactor

leafty (approved):
Thanks

What I don't like is that this works because the `@with_db_transaction` decorator doesn't wrap the method in a transaction if a DB session is passed to it. This blurs the lines of responsibility our Python methods have of the DB transactions.

SalimKayal:
We can (and should) take some time to refactor this more deeply at some point. But let's fix the issue at hand now.

PR #1378, opened 2026-07-08
Enable read-write storage for projects

eikek:
/deploy


components/renku_data_services/migrations/versions/fa511acd4581_add_project_storage.py
```diff
@@ -0,0 +1,121 @@
...
+    op.create_index(
+        op.f("ix_storage_project_storage_project_id"), "project_storage", ["project_id"], unique=True, schema="storage"
+    )
+    op.create_index(
+        op.f("ix_resource_pools_resource_requests_log_phase"),
+        "resource_requests_log",
+        ["phase"],
+        unique=False,
+        schema="resource_pools",
+    )
```
olevski: @eikek things fail because this index already exists and the call here fails. And then it keeps cycling. But the cyclies show errors from an earlier migrations because I think the migrations are partially completed when things crash.

Image

Image

olevski: That is why the CI deployment wont work. That index is already created in migration `c29f65247e88`. You can look it up. It will have the migration id in the filename.

eikek: Oh my… thank you @olevski !! I let alembic create that migration and only fixed the imports. I didn't see that it added this index.

components/renku_data_services/data_connectors/api.spec.yaml
```diff
@@ -70,6 +70,212 @@ paths:
           $ref: "#/components/responses/Error"
       tags:
         - data_connectors
+  /data_connectors/storage:
```
olevski, line 73: The pathing is going to get complicated. I think we can keep this as is for now. But I propose the following in a followup pr:
- there is already `/storage` endpoint but this is for Renku v1
- we remove this `/storage` endpoint and replace it with this

But lets not do this in thi PR.

components/renku_data_services/data_connectors/api.spec.yaml
```diff
@@ -744,6 +978,123 @@ components:
...
+          type: string
+        max_size:
+          type: integer
+          format: int64
+          description: Maximum size in GB
+      required:
+        - project_id
+        - max_size
+        - name
+        - namespace
```
olevski, line 1101: We will very likely add storage at other levels (group and user). And I think that is why you have made the payload for the storage use namespace/path rather than project id. Lets keep thing general here as well. So we can just provide the namespace. And also we can have an indication of what the namespace is (project, user or group).

components/renku_data_services/data_connectors/config.py
```diff
@@ -11,12 +11,30 @@
...
+    maximum_size: ByteSize
+
+    @classmethod
+    def from_env(cls) -> ProjectStorageConfig:
+        """Create a configuration from environment variables."""
+
+        storage_class = os.environ.get("PROJECT_STORAGE_STORAGE_CLASS") or "azurefile"
+        maximum_size = os.environ.get("PROJECT_STORAGE_MAX_SIZE_GB") or "10"
+        maximum_size = ByteSize.from_gibi(int(maximum_size))
+        return ProjectStorageConfig(storage_class=storage_class, maximum_size=maximum_size)
```
olevski: We should not set a default storage class here. Because depending on the cluster `azurefile` may not be accessible. And admins should not be allowed to put storage on the allowlist if they have not properly set a specific read-write-many storage class. So we can make this optional and then error out if it is not set. Or we can also add an enabled flag and if that is true then you have to set the storage class.

components/renku_data_services/data_connectors/core.py
```diff
@@ -211,6 +212,87 @@ async def validate_unsaved_data_connector(
...
+
+def validate_unsaved_project_storage(body: apispec.ProjectStoragePost) -> models.UnsavedProjectStorage:
+    """Validate the user input for a new project storage definition.
+
+    The namespace must be a project namespace. The project must be
+    enabled for project storages and the user must be an owner.
+    """
+    namespace_split = body.namespace.split("/")
+    if len(namespace_split) != 2:
+        raise errors.ValidationError(message=f"The namespace must be a project path: {body.namespace}")
```
olevski: There are dedicated classes for this. You can use them to parse this.

olevski: See `ProjectPath` in base_models/core.py

components/renku_data_services/data_connectors/orm.py
```diff
@@ -332,3 +333,89 @@ def dump(self) -> models.DepositJob:
...
+    """Maximum allowed size in bytes."""
+
+    updated_at: Mapped[datetime] = mapped_column(
+        "updated_at",
+        DateTime(timezone=True),
+        default=None,
+        server_default=func.now(),
+        onupdate=func.now(),
+        nullable=False,
+    )
```
olevski, line 362: I think we should move the permissions about this in Authzed.

components/renku_data_services/data_connectors/orm.py
```diff
@@ -332,3 +333,89 @@ def dump(self) -> models.DepositJob:
...
+
+    __tablename__ = "project_storage"
+
+    id: Mapped[ULID] = mapped_column(
+        "id", ULIDType, primary_key=True, server_default=text("generate_ulid()"), init=False
+    )
+    project_id: Mapped[ULID] = mapped_column(
+        ForeignKey(ProjectStorageAllowORM.project_id, ondelete="CASCADE"), index=True, nullable=False, unique=True
+    )
+    """ID of the project (must exist in project_storage_allow)."""
```
olevski, line 384: Can we do something to make this work for groups and usernamespaces in the future?

olevski (changes requested):
@eikek things look good. Lets touch base about this tomorrow. I am a bit torn how much we should be generalizing and anticipating future improvements now.

components/renku_data_services/data_connectors/api.spec.yaml
olevski: I opened a PR to remove the unused storage stuff https://github.com/SwissDataScienceCenter/renku-data-services/pull/1394

components/renku_data_services/data_connectors/core.py
eikek: `ProjectPath` doesn't have a parse from a single string, but I can add it.

components/renku_data_services/data_connectors/api.spec.yaml
olevski: So lets move this to the /storage path and to the storage folder in a follow up pr.

components/renku_data_services/data_connectors/api.spec.yaml
olevski: As discussed lets revisit this in a follow up pr

components/renku_data_services/data_connectors/core.py
olevski: Oh yeah it has a `from_strings` which will make it from a list of slugs. But not from one whole string. So yeah it should be easy to implement a `from_string` classmethod to instatiate that from one single string. You just have to split on `/` and then call the other method.

components/renku_data_services/data_connectors/orm.py
olevski: Consider / address in a follow up pr.

components/renku_data_services/data_connectors/orm.py
olevski: Lets revisit this in a follow up pr.

components/renku_data_services/data_connectors/config.py
olevski: We can also do this in a follow up pr if you prefer. Let me know.

olevski (approved):

PR #1379, opened 2026-07-08
fix: set work and mount dir correctly

### LINKED ISSUE #1377: Git clone fails for jobs with arbitrary images
Create a launcher from `busybox` image.

Use `["sleep"]` for the command `["10"]` for the args.

Launch the job. It will fail. 

Git clone failed with:
```
2026/07/08 12:27:06 Checking if clone path: /renku-cli exists
2026/07/08 12:27:06 Creating clone path
2026/07/08 12:27:06 failed to create clone path:mkdir /renku-cli: permission denied
```

Note: `renku-cli` is the repository getting cloned.

olevski:
/deploy

fixes #1377

### DIFF
--- a/components/renku_data_services/notebooks/api/amalthea_patches/init_containers.py
+++ b/components/renku_data_services/notebooks/api/amalthea_patches/init_containers.py
@@ -35,6 +35,7 @@ async def git_clone_container_v2(
     repositories: list[Repository],
     git_providers: list[GitProvider],
     workspace_mount_path: PurePosixPath,
+    work_dir: PurePosixPath,
     lfs_auto_fetch: bool = False,
     uid: int = 1000,
     gid: int = 1000,
@@ -51,11 +52,19 @@ async def git_clone_container_v2(
         read_only_etc_certs=True,
     )
 
+    # Trying to use work_dir as the place to clone is in place for backwards compatibility.
+    # We never had issues because before introducing jobs because we always set the work dir as a child of
+    # the storage mount path. But for running jobs with a custom image, then the work dir is usually not
+    # set to anything and defaults to `/`. In this case the work dir is not a child of the storage
+    # mount path and it is not writable in many cases.
+    git_clone_mount_path = work_dir
+    if not work_dir.relative_to(workspace_mount_path):
+        git_clone_mount_path = workspace_mount_path
     prefix = "GIT_CLONE_"
     env = [
         {
             "name": f"{prefix}MOUNT_PATH",
-            "value": workspace_mount_path.as_posix(),
+            "value": git_clone_mount_path.as_posix(),
         },
         {
             "name": f"{prefix}LFS_AUTO_FETCH",
--- a/components/renku_data_services/notebooks/api/amalthea_patches/init_containers.py
+++ b/components/renku_data_services/notebooks/api/amalthea_patches/init_containers.py
@@ -52,19 +52,11 @@ async def git_clone_container_v2(
         read_only_etc_certs=True,
     )
 
-    # Trying to use work_dir as the place to clone is in place for backwards compatibility.
-    # We never had issues because before introducing jobs because we always set the work dir as a child of
-    # the storage mount path. But for running jobs with a custom image, then the work dir is usually not
-    # set to anything and defaults to `/`. In this case the work dir is not a child of the storage
-    # mount path and it is not writable in many cases.
-    git_clone_mount_path = work_dir
-    if not work_dir.relative_to(workspace_mount_path):
-        git_clone_mount_path = workspace_mount_path
     prefix = "GIT_CLONE_"
     env = [
         {
             "name": f"{prefix}MOUNT_PATH",
-            "value": git_clone_mount_path.as_posix(),
+            "value": work_dir.as_posix(),
         },
         {
             "name": f"{prefix}LFS_AUTO_FETCH",
--- a/components/renku_data_services/notebooks/api/amalthea_patches/init_containers.py
+++ b/components/renku_data_services/notebooks/api/amalthea_patches/init_containers.py
@@ -35,7 +35,6 @@ async def git_clone_container_v2(
     repositories: list[Repository],
     git_providers: list[GitProvider],
     workspace_mount_path: PurePosixPath,
-    work_dir: PurePosixPath,
     lfs_auto_fetch: bool = False,
     uid: int = 1000,
     gid: int = 1000,
@@ -56,7 +55,7 @@ async def git_clone_container_v2(
     env = [
         {
             "name": f"{prefix}MOUNT_PATH",
-            "value": work_dir.as_posix(),
+            "value": workspace_mount_path.as_posix(),
         },
         {
             "name": f"{prefix}LFS_AUTO_FETCH",
--- a/components/renku_data_services/notebooks/api/classes/k8s_client.py
+++ b/components/renku_data_services/notebooks/api/classes/k8s_client.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 import base64
 import json
-from datetime import datetime
+from datetime import UTC, datetime
 from typing import cast
 
 import httpx
@@ -217,7 +217,10 @@ class NotebookK8sClient(SecretClient):
             )
         return sorted(
             sessions,
-            key=lambda sess: (sess.launcher_id, -(sess.metadata.creationTimestamp or datetime.min).timestamp()),
+            key=lambda sess: (
+                sess.launcher_id,
+                -(sess.metadata.creationTimestamp or datetime.min.replace(tzinfo=UTC)).timestamp(),
+            ),
         )
 
     async def get_session(self, name: str, safe_username: str) -> AmaltheaSessionV1Alpha1 | None:
--- a/components/renku_data_services/notebooks/api/classes/k8s_client.py
+++ b/components/renku_data_services/notebooks/api/classes/k8s_client.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import base64
+from datetime import datetime
 import json
 from typing import cast
 
@@ -214,7 +215,7 @@ class NotebookK8sClient(SecretClient):
                     sessions,
                 )
             )
-        return sorted(sessions, key=lambda sess: sess.metadata.name)
+        return sorted(sessions, key=lambda sess: sess.metadata.creationTimestamp or datetime.min, reverse=True)
 
     async def get_session(self, name: str, safe_username: str) -> AmaltheaSessionV1Alpha1 | None:
         """Get a specific session, None is returned if the session does not exist."""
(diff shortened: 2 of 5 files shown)


eikek (dismissed):

leafty (changes requested):
Wrong fix: you should update the fallback for `work_dir` and not mess up with using `workspace_mount_path`.

`workspace_mount_path` is NOT the worspace path. Usually it equals `work_dir`, but it may be set as a parent folder of `work_dir`.

What we need to do is re-factor this code block in `start_session` (and do the same for resume):

https://github.com/SwissDataScienceCenter/renku-data-services/blob/8853b75d74e7255f6094b23a62608f705965810c/components/renku_data_services/notebooks/core_sessions.py#L922-L929

olevski:
@leafty there is no guarantee that the workdir is a child of the storage dir. So I will push the same fix further up as you suggest. But that means that in some cases if the user deliberately sets a working directory that is not a child of the storage directory we will just simply ignore the working directory.

In this case if I use the `busybox` image then the working dir in the image is `/`. So it is not the fallback for the workdir that is wrong - the workdir is set in the image it is just not a child of the storage mount.

components/renku_data_services/notebooks/api/amalthea_patches/init_containers.py
```diff
@@ -51,11 +52,19 @@ async def git_clone_container_v2(
...
     )
 
+    # Trying to use work_dir as the place to clone is in place for backwards compatibility.
+    # We never had issues because before introducing jobs because we always set the work dir as a child of
+    # the storage mount path. But for running jobs with a custom image, then the work dir is usually not
+    # set to anything and defaults to `/`. In this case the work dir is not a child of the storage
+    # mount path and it is not writable in many cases.
+    git_clone_mount_path = work_dir
+    if not work_dir.relative_to(workspace_mount_path):
+        git_clone_mount_path = workspace_mount_path
```
leafty: Can we move up this script into `core_sessions.py` (into `start_session()` and `patch_session()`)?

The explanation is not fully correct: the fallback for the work dir is to use the `WORKDIR` defined in the container (we find it in the manifest). We never tried to use `busybox` as a session image before (duh, it has no frontend...), but we would have issues for any image where `WORKDIR` is not user-writeable. For example, we can make images built with buildpacks break if we set the mount and work dirs to `/workspace/source` (read-only folder).

This issue has always existed, which is why I think the fix should be general (try to define the correct `RENKU_MOUNT_DIR` and correct `RENKU_WORK_DIR`).

olevski: Yeah I can move it up.

leafty:
> @leafty there is no guarantee that the workdir is a child of the storage dir. So I will push the same fix further up as you suggest. But that means that in some cases if the user deliberately sets a working directory that is not a child of the storage directory we will just simply ignore the working directory.
> 
> In this case if I use the `busybox` image then the working dir in the image is `/`. So it is not the fallback for the workdir that is wrong - the workdir is set in the image it is just not a child of the storage mount.

(NOTE: see previous comment too, just above)

For images like `busybox`, I think what we have to do for now is ask that the mount and work dirs are set on the session/job launcher to be a new folder (e.g. `/work` or `/home/renku/work`) so that it is created by k8s on pod start.

components/renku_data_services/notebooks/core.py
```diff
@@ -22,15 +22,13 @@ async def docker_image_exists(config: NotebooksConfig, image_url: str, internal_
     return await image_repo.image_exists(parsed_image)
 
 
-async def docker_image_workdir(
-    config: NotebooksConfig, image_url: str, internal_gitlab_user: APIUser
-) -> PurePosixPath | None:
+async def docker_image_workdir(config: NotebooksConfig, image_url: str) -> PurePosixPath | None:
```
leafty: 👍 🎉  getting rid of ghosts 👻 🔨

components/renku_data_services/notebooks/core_sessions.py
```diff
@@ -828,6 +828,31 @@ async def _check_quota(resource_usage_service: ResourceUsageService, resource_po
...
+    nb_config: NotebooksConfig, environment: Environment
+) -> tuple[PurePosixPath, PurePosixPath]:
+    """Get the storage mount and work directories."""
+    work_dir = environment.working_directory
+    if not work_dir:
+        image_workdir = await core.docker_image_workdir(nb_config, environment.container_image)
+        work_dir_fallback = PurePosixPath("/home/jovyan/work")
+        work_dir = image_workdir or work_dir_fallback
+    storage_mount_fallback = work_dir
+    storage_mount = environment.mount_directory or storage_mount_fallback
```
leafty, line 839: Note: should we also check that `storage_mount` is not equal to `/`?

leafty (commented):
I think this would work for me. Will approve if you are happy with how jobs work on `busybox`.

components/renku_data_services/notebooks/core_sessions.py
olevski: Yeah probably. I am adding some tests. And there I can outline a bunch of expected scenarios and cases.

components/renku_data_services/notebooks/core.py
olevski: Haha yeah. I keep finding bits and pieces all over the place.

components/renku_data_services/notebooks/core_sessions.py
olevski: I will make this as ready when I am done with the unit tests and with the tests  in the deployment.

components/renku_data_services/notebooks/core.py
```diff
@@ -22,15 +22,13 @@ async def docker_image_exists(config: NotebooksConfig, image_url: str, internal_
     return await image_repo.image_exists(parsed_image)
 
 
-async def docker_image_workdir(
-    config: NotebooksConfig, image_url: str, internal_gitlab_user: APIUser
-) -> PurePosixPath | None:
+async def docker_image_workdir(image_url: str) -> PurePosixPath | None:
```
leafty, line 25: As a note (maybe this is a new issue): this method will not work for private images.

During session launch, we invoke `image_secret = await get_image_pull_secret(...)` to get the image secret as needed, so we can refactor the code to fetch the image manifest with credentials.

olevski: That is a good point. I will add this then.

olevski: So I will add this in a followup PR. That function would only ever work with images stored in the internal gitlab. So it is not something introduced in this PR.

components/renku_data_services/notebooks/core_sessions.py
```diff
@@ -1086,6 +1106,9 @@ async def start_session(
             [
                 SessionEnvItem(name="RENKU_JOB", value="1"),
                 SessionEnvItem(name="RENKU_SUBMISSION_ID", value=str(launch_request.submission_id)),
+                # NOTE: Without setting CNB_APP_DIR the working directory in jobs is
+                # where the repo used to build the image was cloned
+                SessionEnvItem(name="CNB_APP_DIR", value=work_dir.as_posix()),
```
leafty: Q: why do this only for jobs?

components/renku_data_services/notebooks/api/classes/k8s_client.py
```diff
@@ -214,7 +215,7 @@ async def list_sessions(
                     sessions,
                 )
             )
-        return sorted(sessions, key=lambda sess: sess.metadata.name)
+        return sorted(sessions, key=lambda sess: sess.metadata.creationTimestamp or datetime.min, reverse=True)
```
leafty: This should still be sorted by name first, otherwise, all submissions from different job launchers will be mixed.

See: https://docs.python.org/3/howto/sorting.html#sort-stability-and-complex-sorts
```python
s = sorted(student_objects, key=attrgetter('age'))     # sort on secondary key
sorted(s, key=attrgetter('grade'), reverse=True)       # now sort on primary key, descending
```

```suggestion
        by_creation_date = sorted(sessions, key=lambda sess: sess.metadata.creationTimestamp or datetime.min, reverse=True)
        return sorted(by_creation_date, key=lambda sess: sess.metadata.name)
```

olevski: The order is still being weird. Let me see.

olevski: Ok I can sort by launcher id and then by creation date in one call. When I do it by job name and then by creation date things order in a weird way.

components/renku_data_services/notebooks/core_sessions.py
olevski: Yeah I can apply it in all cases.

components/renku_data_services/notebooks/api/classes/k8s_client.py
```diff
@@ -214,7 +215,10 @@ async def list_sessions(
                     sessions,
                 )
             )
-        return sorted(sessions, key=lambda sess: sess.metadata.name)
+        return sorted(
+            sessions,
+            key=lambda sess: (sess.launcher_id, -(sess.metadata.creationTimestamp or datetime.min).timestamp()),
```
leafty: Unfortunately, this can crash:

```python
>>> from datetime import datetime
>>> datetime.min.timestamp()
  File "<stdin>", line 1, in <module>
    datetime.min.timestamp()
    ~~~~~~~~~~~~~~~~~~~~~~^^
ValueError: year 0 is out of range
```

It can be fixed with:

```python
>>> from datetime import datetime, timezone
>>> datetime.min.replace(tzinfo=timezone.utc).timestamp()
-62135596800.0
```

components/renku_data_services/notebooks/core_sessions.py
```diff
@@ -1070,6 +1090,9 @@ async def start_session(
         SessionEnvItem(name="RENKU_PROJECT_ID", value=str(project.id)),
         SessionEnvItem(name="RENKU_PROJECT_PATH", value=project.path.serialize()),
         SessionEnvItem(name="RENKU_LAUNCHER_ID", value=str(launcher.id)),
+        # NOTE: Without setting CNB_APP_DIR the working directory in jobs is
+        # where the repo used to build the image was cloned
```
leafty: Suggestion:

```suggestion
        # NOTE: CNB_APP_DIR sets the working directory for images built with cloud native buildpacks
```

leafty (changes requested):

components/renku_data_services/notebooks/api/classes/k8s_client.py
olevski: Oh wow I did not know that. Ok.

leafty (approved):
LGTM thanks 👍

components/renku_data_services/notebooks/core.py
olevski: The followup pr is: #1380

PR #1181, opened 2026-01-19
feat: allows searching data connector project links by DOI

mohammad-alisafaee:
* Adds a `/data_connectors/search?doi=...` endpoint to get a data connector from a given DOI
* Adds pagination to the `/data_connectors/{data_connector_id}/project_links` endpoint
* Returns project path when querying the endpoint

/deploy

PR #1381, opened 2026-07-10
feat: get data connector links by doi

olevski:
/deploy

This is a branch off of #1181 which excludes changes that need modification of the ui.

We can merge #1181 at a later time.

The endpoint added here is needed by Envidat.


olevski:
Note: I still need to rebase and remove some unnecessary changes.

components/renku_data_services/data_connectors/db.py
```diff
@@ -964,6 +964,55 @@ async def update_deposit(
             await session.refresh(res)
         return res.dump()
 
+    async def get_links(
+        self,
+        user: base_models.APIUser,
+        doi: str | None,
```
leafty:
```suggestion
        doi: DOI | None,
```

From `from renku_data_services.data_connectors.doi.models import DOI`

leafty (changes requested):

components/renku_data_services/data_connectors/api.spec.yaml
```diff
@@ -815,6 +868,8 @@ components:
           $ref: "#/components/schemas/Ulid"
         project_id:
           $ref: "#/components/schemas/Ulid"
+        project_path:
+          $ref: "#/components/schemas/ProjectPath"
```
leafty: Is this necessary? This seems to add complexity in the code we could do without (adding ORM links and related ORM queries which will slow down the response time). The `project_id` can be used to know the project details.

olevski: I think it may have been requested by Envidat. But I will remove it and we can add it later on if they really want it.

leafty: Well, `project_path` would NOT be a permanent identifier, so I don't know if it's a good idea to rely on it.

leafty (approved):

PR #1382, opened 2026-07-10
chore: make sure tests can run in parallel

olevski:
Some tests have to be limited to run in series. This includes anything that needs a kind cluster. when they are not added to the group that indicates that this is the case running the tests locally with multiple workers will fail.

### DIFF
--- a/test/bases/renku_data_services/data_api/test_connected_services.py
+++ b/test/bases/renku_data_services/data_api/test_connected_services.py
@@ -643,6 +643,7 @@ async def test_get_token_with_internal_token(
 # --- Integration Group Connected Services API Tests (E1-E2) ---
 
 
+@pytest.mark.xdist_group("sessions")  # Needs to run on the same worker as the rest of the sessions tests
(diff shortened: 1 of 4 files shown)


olevski:
The error message that shows up when these tests fail when running in parallel is this:
```
ERROR test/bases/renku_data_services/data_api/test_sessions.py::test_patch_session_launcher_from_code_update_command - subprocess.CalledProcessError: Command '['kind', 'create', 'cluster', '--name', 'k8s-cluster-01kx6fsdarvpr0tyy016sy5vd0', '--kubeconfig', '/tmp/pytest-of-vscode/pytest-27/popen-gw5/kube0/.kind-kubeconfig.yaml']' returned non-zero exit status 1.
```

olevski:
To replicate all I had to do is:
- launch a devcontainer
- run `make main_tests`

We dont see this in ci pipelines on github most likely because our runners are small and they only run 2 workers.

wesjdj (approved):

PR #1383, opened 2026-07-13
feat: add persisted logs

leafty:
Adds support for persisted logs from Amalthea sessions and image builds.

* Logs are collected from the Loki read API and are stored in the database when attributed to an existing user, session launcher or image build
* Persisted logs for sessions (interactive and jobs) are readable from their owner only
* Persisted logs for image builds are readable by project editors (+repository readers for private image builds)
* Persisted logs are removed once their TTL has expired (set from helm chart values)

Given that persisted logs rely on specific Loki configuration settings, the feature is toggled off by default in the renku helm chart.

/deploy renku=leafty/feat-peristed-logs amalthea-sessions=leafty/feat-run-id-jobs renku-ui=leafty/feat-peristed-logs extra-values=dataService.persistedLogs.enabled=true,dataService.imageBuilders.enabled=true,dataService.imageBuilders.strategyName=renku-buildpacks-v3,dataService.imageBuilders.outputImagePrefix=harbor.dev.renku.ch/renku-build/,dataService.imageBuilders.nodeSelector.renku.io/node-purpose=user,dataService.imageBuilders.tolerations[0].effect=NoSchedule,dataService.imageBuilders.tolerations[0].key=renku.io/dedicated,dataService.imageBuilders.tolerations[0].operator=Equal,dataService.imageBuilders.tolerations[0].value=user


components/renku_data_services/persisted_logs/orm.py
```diff
@@ -0,0 +1,99 @@
...
+    session_uid: Mapped[str | None] = mapped_column(nullable=True)
+    """The session UID for this session run."""
+
+    launcher_id: Mapped[ULID] = mapped_column(
+        ULIDType, ForeignKey(SessionLauncherORM.id, ondelete="CASCADE"), index=True, nullable=False
+    )
+    """The session launcher ID of the session."""
+
+    submission_id: Mapped[str | None] = mapped_column(nullable=True)
+    """The submission ID, if the session run corresponds to an offline job."""
```
olevski: What is an "offline job" and how is this different than a regular job? What makes it "offline"?

components/renku_data_services/persisted_logs/blueprints.py
```diff
@@ -0,0 +1,86 @@
...
+                    user=user,
+                    launcher_id=launcher_id,
+                    run_id=run_id,
+                    submission_id=query.submission_id,
+                )
+            if result is None:
+                # TODO: adjust error message when params are passed in the query
+                raise errors.MissingResourceError(
+                    message=f"Session launcher with id '{launcher_id}' does not have persisted logs."
+                )
```
olevski, line 54: Reminder to not forget.

components/renku_data_services/persisted_logs/collector.py
```diff
@@ -0,0 +1,334 @@
...
+        query = (
+            "{"
+            f'{PERSISTED_LOGS_SESSIONS_LABEL_KEY}="{PERSISTED_LOGS_SESSIONS_LABEL_VALUE}",'
+            f'{PERSISTED_LOGS_NAMESPACE_LABEL_KEY}="{self.config.namespace}"'
+            "}"
+        )
+        response = await self._get_logs(query=query, limit=limit, start=start, end=end)
+        async for item in self._process_session_logs(response):
+            yield item
+
```
olevski, line 61: I assume that if the start time is not given the start time is the present... It may be nice to clarify this in the docstring.

components/renku_data_services/persisted_logs/db.py
```diff
@@ -0,0 +1,422 @@
...
+            .join(
+                schemas.AmaltheaSessionLogORM,
+                schemas.SessionRunORM.id == schemas.AmaltheaSessionLogORM.run_id,
+                isouter=True,  # isouter makes it a left-join, not an outer join
+            )
+            .where(schemas.AmaltheaSessionLogORM.id.is_(None))
+        )
+        session_runs_res = await session.scalars(stmt)
+        session_run_ids = session_runs_res.all()
+        await session.execute(delete(schemas.SessionRunORM).where(schemas.SessionRunORM.id.in_(session_run_ids)))
```
olevski: You can use a subquery here ane execute just once. I think there is a subquery function you call from sqlalchemy to make a regular query a subquery. I dont think it is that hard to use.

components/renku_data_services/persisted_logs/api.spec.yaml
```diff
@@ -0,0 +1,237 @@
...
+          $ref: "#/components/schemas/Ulid"
+        session_uid:
+          type: string
+        launcher_id:
+          $ref: "#/components/schemas/Ulid"
+        submission_id:
+          type: string
+      required:
+        - id
+        - launcher_id
```
olevski, line 149: I think the submission ids are just for jobs right? And jobs also have session_uids, right? Or they do not?

olevski (changes requested):
Thanks Flora. Just a few minor things.

components/renku_data_services/persisted_logs/collector.py
leafty: > Defaults to one hour ago.

This is in the Loki docs which is linked in the doc string. The link is for the query range endpoint, so it points to exactly the relevant doc.

components/renku_data_services/persisted_logs/api.spec.yaml
leafty: `submission_id` will only exist for jobs -> optional field.
The session_uid should be there but its absence doesn't break the logs feature, so it's also optional.

components/renku_data_services/persisted_logs/orm.py
leafty: It's a job.

components/renku_data_services/persisted_logs/blueprints.py
leafty: Well, I want to make issues and fix the todos in another PR.

olevski: Ok no problem. Just wanted to bring it up in case that was not the intention.

olevski (approved):

PR #1384, opened 2026-07-13
feat: add option to set cpu limits

olevski:
/deploy 

Some clusters - particularly for openshift have defaulting/validating webhook or something similar that will apply a constant value that makes the cpu limit equal to 1 if unset. This causes problems because by default we do not set cpu limits. This makes it so that the admin can add a factor on the resource pool so that the cpu limit is set based on the request if the factor is set.

The automatic constant limit setting for cpus causes issues when the sessions request cpus that are greater than this constant value. In this case the session just fails to start. Because the cpu limit ends up being < request.


olevski:
The CI deployment has a public project with a resource pool that assigns limit and another that does not.

components/renku_data_services/notebooks/core_sessions.py
```diff
@@ -496,7 +498,7 @@ def resources_patch_from_resource_class(resource_class: ResourceClass) -> Resour
     return ResourcesPatch(requests=requests, limits=limits)
 
 
-def resources_from_resource_class(resource_class: ResourceClass) -> Resources:
+def resources_from_resource_class(resource_class: ResourceClass, cpu_limit_factor: float | None = None) -> Resources:
```
leafty, line 501: Why do we handle CPU resources differently from memory and GPU here? Why do we need a `cpu_limit_factor` field but no similar field for the other resources?

olevski: The reason is that we never put cpu limits.Just requests. For all things we put limits and requests. For cpu if we put limits the allowed usage is just capped even if there is spare capacity on the node. And when the node is overprovisioed linux can handle it without absolutely breaking or restarting the pods.

olevski: Also for gpus you have to set the limit equal to the request. You cannot omit it.

leafty: OK, this is understandable, but I think we should document this in the code and in the admin docs.

Handling of resource requests and limits in Kubernetes:
* CPU are burstable with low risk on the host node; which is why we either set no limit or a limit which is `f x CPU_request` with `f > 1`
* Memory limits should be equal to memory requests for sessions because over-subscription can force the host node to swap which may affect multiple sessions from multiple users.
* GPUs are not burstable, so the limit needs to be equal to the request.

With this, it becomes clear why CPU limits can be unset or set above the request while memory and GPUs must have `limit = request`.

olevski: Ok yeah I will add this to some of the docstrings in this PR here and open a PR in the renku repo.

components/renku_data_services/crc/core.py
```diff
@@ -195,6 +195,9 @@ def validate_resource_pool_post(body: apispec.ResourcePool) -> models.UnsavedRes
     remote = validate_remote(body=body.remote) if body.remote else None
     platform = __validate_runtime_platform(body=body.platform)
 
+    if body.cpu_limit_factor is not None and body.cpu_limit_factor < 1:
+        raise errors.ValidationError(message="The cpu limit factor has to be >= 1.")
+
```
leafty: Do we need this? Isn't it already validated by pydantic?

components/renku_data_services/crc/core.py
```diff
@@ -247,6 +254,9 @@ def validate_resource_pool_put_or_patch(
         remote=remote,
         cluster_id=ULID.from_str(body.cluster_id) if body.cluster_id else None,
         platform=platform,
+        cpu_limit_factor=RESET
+        if body.cpu_limit_factor is None or body.cpu_limit_factor == 0
+        else body.cpu_limit_factor,
```
leafty: This will force a reset on PATCH. RESET can be set if the method is PUT.

components/renku_data_services/crc/api.spec.yaml
```diff
@@ -2411,6 +2421,12 @@ components:
     IngressAnnotations:
       type: object
       additionalProperties: true
+    CpuLimitFactor:
```
leafty, line 2428: Note: maybe an additional type is needed to allow resetting on PATCH: it can be the union of `CpuLimitFactor` and the constant `0`.

components/renku_data_services/crc/core.py
olevski: Removed it.

components/renku_data_services/crc/api.spec.yaml
olevski: Added a special type for removing the value.

components/renku_data_services/migrations/versions/5cb5a8d5663b_add_cpu_limit_factor.py
```diff
@@ -0,0 +1,35 @@
...
+# revision identifiers, used by Alembic.
+revision = "5cb5a8d5663b"
+down_revision = "0a6cee40fe0d"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column("resource_pools", sa.Column("cpu_limit_factor", sa.Integer(), nullable=True), schema="resource_pools")
```
leafty: Huh is this migration outdated? The field is mapped to `sa.Integer()`, so if I pass `1.5` to the API, the actual value stored is `1`.

leafty (changes requested):

components/renku_data_services/migrations/versions/5cb5a8d5663b_add_cpu_limit_factor.py
olevski: Ah yeah at first I made it integer. But it should be float.

leafty (approved):
Code looks good.

Remark: maybe there should be updates to resource pool tests to add/update the `cpu_limit_factor` field. This can be done in another PR if not here.

olevski:
>  maybe there should be updates to resource pool tests to add/update the cpu_limit_factor field. This can be done in another PR if not here.

Agreed. I will open a follow up PR.

PR #1388, opened 2026-07-15
feat: update apispec for polymorphic resource class definition

SalimKayal:
## Summary

Adds a `kind` and optional FirecREST `remote` override to resource classes so that FirecREST-backed
classes can target specific Slurm systems/partitions and optionally let the HPC grid choose CPU,
memory, and GPU. A single resource pool can now contain classes with different Slurm target
configurations.

## Motivation and context

We originally targeted only CSCS, where the smallest allocation is a full node, so it was enough to
to let Slurm set the defaults for `CPUS_PER_TASK`, `GPUS`, `RAM`. As we expand to other grids
(Idiap) we need to tell Amalthea exactly how much of a node to request. We also want a single
resource pool to support multiple class-level target configurations and optionally which
`system_name`/`partition` to use, rather than forcing one pool per Slurm target.

## Behavior

- Resource classes now have a `kind` (`local`, `firecrest`, `runai`). The kind must match the parent
  pool's remote kind; `firecrest` classes require integer CPU and may only carry `remote` overrides
  when the pool is FirecREST.
- FirecREST classes support per-class `remote` overrides for `system_name` and `partition`, falling
  back to the pool defaults when omitted. This lets a single pool hold multiple classes targeting
  different Slurm systems or partitions.
- FirecREST classes can set `ignore_resource_class_values` to tell Amalthea to ignore the CRD CPU,
  memory, and GPU values and let the HPC grid choose them. The class still stores and reports the
  values for display and quota matching.
- On session launch, FirecREST classes emit `RSC_FIRECREST_SYSTEM_NAME`, `RSC_FIRECREST_PARTITION`
  (when set), and `RSC_FIRECREST_IGNORE_RESOURCE_CLASS_VALUES` (when true). CPU, memory, and GPU
  continue to be passed through the session CRD resources as before.
- Changing a pool's remote kind auto-converts existing classes to the new kind. The update is
  rejected if any class still has FirecREST remote overrides when moving away from FirecREST, or if
  any class has non-integer CPU when moving to FirecREST.
- The API remains backward compatible: resource class schemas default `kind` to `local`, so existing
  clients and legacy classes keep working.

## Changes

- Updated `crc/api.spec.yaml` and regenerated `crc/apispec.py`:
  - `ResourceClass`/`ResourceClassWithId` now include `kind: RemoteKind` (default `local`) and an
    optional `remote: RemoteClassConfigurationFirecrest`.
  - `RemoteClassConfigurationFirecrest` gains `ignore_resource_class_values` (default `false`).
  - Added `if/then/else` schema rules so `remote` is forbidden when `kind != firecrest` and `cpu`
    must be an integer when `kind == firecrest`.

- Extended `crc/models.py` with `FirecrestClassRemote` and added `kind`/`remote` to
  `UnsavedResourceClass`, `ResourceClass`, and `ResourceClassPatch`.

- Extended `crc/orm.py` with a `remote_json` column for class-level FirecREST overrides. Class
  `kind` is derived at read time from the parent pool's `remote_json.kind`.

- Added an Alembic migration that adds `resource_classes.remote_json`.

- Updated `crc/core.py` validators to:
  - Enforce pool-kind/class-kind matching.
  - Require integer CPU for FirecREST classes.
  - Reject `remote` overrides on non-FirecREST classes.
  - Derive `kind` correctly on PUT/PATCH when omitted.
  - Validate class compatibility when a pool's remote kind changes.

- Updated `crc/db.py` to persist `remote_json`, reject `kind` changes, and apply pool remote changes
  before class updates so class conversion sees the new pool kind.

- Updated `crc/blueprints.py` to pass the existing class kind into standalone class PUT/PATCH
  validators.

- Updated `notebooks/core_sessions.py` to compute effective FirecREST env vars from class overrides
  and emit `RSC_FIRECREST_IGNORE_RESOURCE_CLASS_VALUES` when requested.

- Added unit tests for the new validators and updated FirecREST resource-pool and session-launch
  tests.

## Notes

- Resource class `kind` is not stored on the `resource_classes` table; it is derived at read time
  from the parent pool's `remote_json.kind` (defaulting to `local`). This keeps the migration
  minimal (only `remote_json` is added) while still making existing FirecREST/Run:AI classes appear
  with the correct kind.
- The UI can still read/write pool-level `system_name`/`partition` for now; they serve as the
  defaults for classes without per-class overrides.


components/renku_data_services/crc/api.spec.yaml
```diff
@@ -1323,6 +1323,21 @@ components:
           $ref: "#/components/schemas/NodeAffinityList"
         quota_enforced:
           $ref: "#/components/schemas/QuotaEnforced"
+        kind:
```
leafty: I don't understand why we repeat `kind` in the resource class when it cannot be different from the parent resource pool.

components/renku_data_services/crc/api.spec.yaml
```diff
@@ -1323,6 +1323,21 @@ components:
...
+        remote:
+          $ref: "#/components/schemas/RemoteClassConfigurationFirecrest"
+      if:
+        properties:
+          kind:
+            const: "firecrest"
+      then:
+        properties:
+          cpu:
+            $ref: "#/components/schemas/CpuInteger"
```
leafty: Why do we add this complexity in the API spec? I would rather have this handled by a `math.floor()` inside the remote session controller on the amalthea side than make a complex API spec.

components/renku_data_services/crc/api.spec.yaml
```diff
@@ -1367,18 +1398,72 @@ components:
         default_storage: 10
         quota_enforced: false
     ResourceClassPatch:
-      allOf:
-        - $ref: "#/components/schemas/ResourceClassProperties"
+      type: object
```
leafty: Note: removing the `allOf` kind of negates having the `ResourceClassProperties` type.

components/renku_data_services/crc/api.spec.yaml
SalimKayal: it allows the conditional in the api spec. I hesitated a lot with this because datamodel code generator does not support this. I can change that if you think it's more trouble than otherwise

components/renku_data_services/crc/api.spec.yaml
SalimKayal: because floor of 0.5 is not what we want ? but again this can be validated elsewhere (i.e. in core.py)

components/renku_data_services/crc/api.spec.yaml
SalimKayal: this is so that we can have a nullable kind. if we remove kind this can come back

SalimKayal (commented):
So basically all your remarks are linked to the kind in resource class. should I remove this and manage this aspect differently ?

components/renku_data_services/crc/api.spec.yaml
leafty: I mean, we can have:
```python
cpu_int = max(math.floor(cpu), 1)
```

leafty: to have a minimum of 1 and then truncating if set to a non-whole value

components/renku_data_services/crc/api.spec.yaml
SalimKayal: `kind` is gone, per #1404

components/renku_data_services/crc/api.spec.yaml
SalimKayal: out of the api spec per #1404

components/renku_data_services/crc/api.spec.yaml
SalimKayal: reverted

components/renku_data_services/crc/api.spec.yaml
```diff
@@ -2188,6 +2194,20 @@ components:
...
         - system_name
+    RemoteClassConfigurationFirecrest:
+      type: object
+      additionalProperties: false
+      properties:
+        system_name:
+          $ref: "#/components/schemas/RemoteConfigurationFirecrestSystemName"
+        partition:
+          $ref: "#/components/schemas/RemoteConfigurationFirecrestPartition"
+        ignore_resource_class_values:
```
leafty: Would it make more sense to move this field to the resource class? It seems that all the classes in a remote HPC pool would set this to the same value.

components/renku_data_services/crc/core.py
```diff
@@ -86,9 +114,16 @@ def validate_resource_class_patch_or_put(
     | apispec.ResourceClass
     | apispec.ResourceClassWithId,
     method: Literal["PATCH", "PUT"],
+    *,
+    existing_kind: models.RemoteConfigurationKind | None = None,
 ) -> models.ResourceClassPatch | models.ResourceClassPatchWithId:
-    """Validate the patch to a resource class."""
+    """Validate the patch/put of a resource class."""
     rc_id = body.id if isinstance(body, (apispec.ResourceClassPatchWithId, apispec.ResourceClassWithId)) else None
+    kind = existing_kind if method == "PATCH" else None
```
leafty: Looks like this line may break `PUT`.

components/renku_data_services/crc/core.py
```diff
@@ -101,6 +136,17 @@ def validate_resource_class_patch_or_put(
     tolerations: list[str] | None = [] if method == "PUT" else None
     if body.tolerations:
         tolerations = sorted(t.root for t in body.tolerations or [])
+    remote: models.FirecrestClassRemote | None = None
+    if body.remote is not None:
+        if kind != models.RemoteConfigurationKind.firecrest:
```
leafty, line 139: See above: it seems kind is always set to `None` when `method == 'PUT'`. So, how do we set the `remote` field when using the `PUT` method?

components/renku_data_services/crc/core.py
```diff
@@ -299,6 +421,11 @@ def validate_resource_pool_update(existing: models.ResourcePool, update: models.
                 message=f"Resource class '{rc.id}' does not exist in resource pool '{existing.id}'."
             )
         idx, existing_rc = found
+        if rc.kind is not None and rc.kind != existing_rc.kind and rc.kind != new_pool_kind:
+            raise errors.ValidationError(message="The resource class kind cannot be changed.")
```
leafty: Why can't we change the `kind`? Qhile there is validation code to check when we change the `kind`...

SalimKayal: this is is unreachable under the new design, the kind is always the pool kind.

I will remove this.

components/renku_data_services/crc/core.py
SalimKayal: from #1404 the remoteness comes from the resource pool

components/renku_data_services/crc/core.py
SalimKayal: well caught. thanks.

components/renku_data_services/crc/api.spec.yaml
SalimKayal: I wondered a bit about this. Here's why I went this way.

If we keep the resource pools/classes as is it works because the resource pool's values are used.

if we change the values in the resource class, it overrides the pool's defaults

this means that if we do not change anything in the configs it should still work, if we do we can have a single resource pool per firecrest api endpoint. i.e.` CSCS Alps` with `Daint`/`Eiger`/`Bristen`/`Clariden` Systems and `debug`/`normal` partitions. I can change that if you think it's poor design.

leafty: I think it would be more likely that an admin forgets about this field when updating some classes. If we have `forward_resource_values` at the resource pool level, then it is less likely to be mis-configured (only needs to be set once per pool).

leafty (commented):
Comment: without a `platform` override (`amd64 | arm64`), it is not possible to merge clusters in a single resource pool.

SalimKayal:
> Comment: without a `platform` override (`amd64 | arm64`), it is not possible to merge clusters in a single resource pool.

Should I add it while I'm at it ?

leafty:
> > Comment: without a `platform` override (`amd64 | arm64`), it is not possible to merge clusters in a single resource pool.
> 
> Should I add it while I'm at it ?

Sure.

components/renku_data_services/crc/api.spec.yaml
SalimKayal: the default for the `forward_resource_values` is false. we may want to have some classes with defaults and some with config. Also we could have a default in the pool and an override in the class ? Or do we disregard this feature for admins and simplify it at the pool level ?

SalimKayal:
> > > Comment: without a `platform` override (`amd64 | arm64`), it is not possible to merge clusters in a single resource pool.
> > 
> > 
> > Should I add it while I'm at it ?
> 
> Sure.

this is a bigger redesign as we need to use it to validate the image builds. do I simply wire it in for now then add the ui/image-building stuff in a follow up PR or is all of this a follow up pr ?

leafty:
> > > > Comment: without a `platform` override (`amd64 | arm64`), it is not possible to merge clusters in a single resource pool.
> > > 
> > > 
> > > Should I add it while I'm at it ?
> > 
> > 
> > Sure.
> 
> this is a bigger redesign as we need to use it to validate the image builds. do I simply wire it in for now then add the ui/image-building stuff in a follow up PR or is all of this a follow up pr ?

We should leave it out then.

components/renku_data_services/crc/api.spec.yaml
leafty: Is there a case where a resource pool would have a mix bag of `forward: true` and `forward: false` for classes? It seems to me that the setting could be left at the pool level.

SalimKayal:
Info: `ignore_resource_class_values` renamed to `forward_resource_value`

components/renku_data_services/crc/api.spec.yaml
SalimKayal: At CSCS they were talking about allowing partial node allocation because some people were complaining that 4xGB200 is a bit much for some tasks.

components/renku_data_services/crc/api.spec.yaml
```diff
@@ -2188,6 +2194,21 @@ components:
...
         - system_name
+    RemoteClassConfigurationFirecrest:
+      type: object
+      additionalProperties: false
+      properties:
+        system_name:
+          $ref: "#/components/schemas/RemoteConfigurationFirecrestSystemName"
+        partition:
+          $ref: "#/components/schemas/RemoteConfigurationFirecrestPartition"
+        forward_resource_value:
```
leafty:
```suggestion
        forward_resource_values:
```

components/renku_data_services/crc/core.py
```diff
@@ -210,13 +257,69 @@ def validate_resource_pool_post(body: apispec.ResourcePool) -> models.UnsavedRes
...
+        # Reject class-level remote data when the new pool kind is not FirecREST.
+        has_remote = rc.remote is not None or (existing is not None and existing.remote is not None)
+        if new_pool_kind != models.RemoteConfigurationKind.firecrest and has_remote:
+            raise errors.ValidationError(
+                message="Remove class-level remote configuration before converting the pool away from FirecREST."
+            )
+        # FirecREST requires integer CPU.
+        cpu = rc.cpu if rc.cpu is not None else (existing.cpu if existing is not None else None)
+        if new_pool_kind == models.RemoteConfigurationKind.firecrest and cpu is not None and not cpu.is_integer():
+            raise errors.ValidationError(message="FirecREST resource classes require an integer value for cpu.")
```
leafty, line 299: Can we not do `ceil()` instead? This could be annoying to deal with. (Or the admin would have to delete and re-create).

leafty (commented):
Looks good now, we should deploy and test.

leafty (dismissed):

components/renku_data_services/crc/core.py
SalimKayal: The check only fires on writes (PUT/PATCH of the pool, POST of a class). GETs don't validate, so existing fractional-cpu classes still read back fine and don't need migrating.

Failing at write time is actually the right moment to be explicit: the admin is already making a change, so a clear error tells them to set an integer cpu in the same request rather than silently coercing a value they typed. 

For the case where fractional data does slip through (legacy rows, etc.), I've switched the launch-time coercion in core_sessions.py from int() (truncated down) to ceil() (rounds up). That achieves the "never under-provision" result without hiding the write-time mismatch behind a silent fix.

leafty (dismissed):

leafty (approved):

PR #1389, opened 2026-07-15
fix: distinguish union for patching launchers envs

### LINKED ISSUE #1385: Changing the name of a launcher does not work
To reproduce:
- create a launcher
- edit its name
- the response will be:

Image

I suspect some value is missing or being added to the payload.

olevski:
fixes #1385 

I am not exactly sure exactly what change caused  this. But I added explicit discrimination for the union in the patch. So it eliminates the error. And we do not depend on how Pydantic handles this case by default.

/deploy


leafty (approved):

leafty (approved):
OK, though I think the issue stems from mis-use of `allOf`, `anyOf`, or `oneOf` types in the api spec.

PR #1391, opened 2026-07-28
feat(apps): add build-from-code custom frontend, data connectors, etc

wesjdj:


components/renku_data_services/session/models.py
```diff
@@ -473,4 +493,22 @@ class ShipwrightBuildStatusUpdate:
...
+        working_directory=BUILD_WORKING_DIRECTORY,
+        mount_directory=BUILD_MOUNT_DIRECTORY,
+        uid=BUILD_UID,
+        gid=BUILD_GID,
+        environment_kind=EnvironmentKind.CUSTOM,
+        environment_image_source=EnvironmentImageSource.build,
+        # A bring-your-own app is unlikely to be aware of the session's URL prefix,
+        # so strip it (like RStudio) and serve the app at "/". Apps that *are*
+        # prefix-aware can read $RENKU_BASE_URL_PATH instead.
+        strip_path_prefix=True,
```
olevski: question: All apps are running on a subdomain and the path for each app is `/`, right?. So we do not need to set this at all.

components/renku_data_services/session/models.py
```diff
@@ -92,6 +99,19 @@ class UnsavedBuildParameters:
...
+        """The frontend token passed to the buildpacks via BP_RENKU_FRONTENDS.
+
+        This is the persisted ``frontend_variant`` for every RenkuLab-provided frontend.
+        The "custom" frontend has no built-in UI, so it maps to the buildpack token
+        "none": the buildpacks then produce a runnable image without injecting a
+        frontend and defer the launch process to the repository Procfile.
+        """
+        if self.frontend_variant == FrontendVariant.custom:
+            return "none"
+        return self.frontend_variant
```
olevski: nitpick: It may be nicer to resolve `self.frontend_variant` into the `FrontendVariant` enum inside this proprety and just return the enum rather than a string.

components/renku_data_services/renku_apps/api.spec.yaml
```diff
@@ -125,18 +99,13 @@ components:
...
-      enum:
-        - running
-        - hibernated
 
     AppName:
       type: string
       minLength: 5
+      # The k8s limit is 63 characters (the name becomes a hostname label) but we leave
+      # some leeway. Must match APP_NAME_MAX_LENGTH in renku_apps/core.py, which is what
+      # generate_app_name() truncates to.
```
olevski: nitpick: should we add this to the help field rather than just in a comment?

components/renku_data_services/renku_apps/core.py
```diff
@@ -1,8 +1,38 @@
...
+_LAUNCHER_ID_SUFFIX_LENGTH: Final[int] = 8
+"""Trailing characters of the launcher ULID that make an app name unique per launcher."""
+
+_SLUG_MAX_LENGTH: Final[int] = APP_NAME_MAX_LENGTH - _LAUNCHER_ID_SUFFIX_LENGTH - 1
+
+
+def generate_app_name(project_slug: str, launcher_id: ULID) -> str:
+    """Generate a DNS-1035 label name for an app, bounded to APP_NAME_MAX_LENGTH."""
+    suffix = str(launcher_id)[-_LAUNCHER_ID_SUFFIX_LENGTH:].lower()
+    return f"{_slug_label(project_slug)}-{suffix}"
```
olevski, line 48: Please add a few test cases for this to make sure it works as expected. It should not take long to add them.

components/renku_data_services/renku_apps/core.py
```diff
@@ -1,8 +1,38 @@
...
+
+
+def _slug_label(slug: str) -> str:
+    """Coerce a project slug (which may hold dots, underscores or a leading digit) into a DNS-1035 label."""
+    label = re.sub(r"-+", "-", re.sub(r"[^a-z0-9]", "-", slug.lower()))
+    if not label[:1].isalpha():
+        label = f"app-{label}"
+    return label[:_SLUG_MAX_LENGTH].rstrip("-")
```
olevski, line 56: question: I understand the outermost re.sub call - that one replaces repeated `-` in the name with a single `-`. But the second I am confused about. It replaces the first character or number of slug with a dash? Is this because you are trying to join with dash in `f"app-{label}"`. 

Also you have this: `re.sub(r"[^a-z0-9]", "-", slug.lower())` but I suspect you want `re.sub(r"^[a-z0-9]", "-", slug.lower())`, the beggining of string character should be outside of the set.

components/renku_data_services/renku_apps/data_connectors.py
```diff
@@ -0,0 +1,94 @@
+"""Launch-time selection of the data connectors an app may mount.
+
+Apps are publicly and anonymously reachable and run as ``DUMMY_RENKU_APP_USER_ID``
+rather than a real user. The set of data connectors an app mounts is therefore a
+**security boundary**: mounting a connector that carries a user's decrypted
+credentials would expose that user's data to anonymous visitors. This module
+defines the fail-closed filter that decides which of a project's linked connectors
+are safe to expose through an app.
```
olevski: nitpick: Do we really need a separate file for these 2 functions in this file? Why not move them to core.py?

components/renku_data_services/renku_apps/data_connectors.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+    try:
+        return (
+            dc.visibility == Visibility.PUBLIC
+            and not list(validator.get_private_fields(dc.storage.configuration))
+            and dc.storage.configuration.get("type") not in _OAUTH2_INTEGRATION_STORAGE_TYPES
+        )
+    except Exception:
+        logger.warning("Excluding data connector %s from app: predicate error", getattr(dc, "id", "?"), exc_info=True)
+        return False
```
olevski: issue: please handle the specific exception or exceptions, not all exceptions, this can hide real bugs or issues

components/renku_data_services/renku_apps/data_connectors.py
```diff
@@ -0,0 +1,94 @@
...
+    async for dc in dc_secret_repo.get_data_connectors_with_secrets(user, project_id):
+        if is_app_mountable(dc.data_connector, validator):
+            survivors.append(dc)
+
+    logger.info(
+        "Selected %d data connector(s) to mount for project %s",
+        len(survivors),
+        project_id,
+    )
+    return survivors
```
olevski: question: If this is supposed to select dcs for apps and we know we want public dcs only why allow `user` to be passed? It seems like we want user to always be the the one for anonymous user. So just hardcode it inside the function.

components/renku_data_services/renku_apps/k8s_client.py
```diff
@@ -191,8 +265,58 @@ def _resources_from_resource_class(resource_class: ResourceClass) -> dict[str, A
...
 
 
+def _app_labels(session_launcher: SessionLauncher, project: Project) -> dict[str, str]:
+    """Labels shared by an app's Knative Service and its owned data-connector resources."""
+    return {
+        "renku.io/safe-username": DUMMY_RENKU_APP_USER_ID,
+        "renku.io/project-slug": project.slug.lower(),
+        "renku.io/project-namespace": project.namespace.path.serialize().replace("/", "-").lower(),
+        "renku.io/project-id": str(project.id),
+        "renku.io/project-id-slug": str(project.id)[18:26].lower(),
```
olevski, line 319: question: What is the purpose of saving the project id slug in here like this? Seems a bit redundant.

components/renku_data_services/renku_apps/k8s_client.py
```diff
@@ -191,8 +265,58 @@ def _resources_from_resource_class(resource_class: ResourceClass) -> dict[str, A
...
+
+
+def _service_owner_reference(app_name: str, uid: str) -> dict[str, Any]:
+    """Owner reference making the app's Knative Service the owner of its PVC/Secret."""
+    return {
+        "apiVersion": KNATIVE_SERVICE_GVK.group_version,
+        "kind": KNATIVE_SERVICE_GVK.kind,
+        "name": app_name,
+        "uid": uid,
+        "controller": True,
```
olevski: Is there a reason for `controller: True`? My understanding is this is when a resource is owned by a k8s controller which is solely responsible for reconciling the resource. But in this case we do not have a controller for apps.

components/renku_data_services/renku_apps/k8s_client.py
```diff
@@ -45,19 +106,54 @@ def __init__(
...
+        data_connectors: list[DataConnectorWithSecrets],
     ) -> AppRuntimeState:
         """Create a deployment for the given app and return its observed runtime state."""
         cluster = await self.__client.cluster_by_id(self.__cluster_id)
-        app_name = _generate_app_name(project, session_launcher)
-        manifest = _build_app_deployment_manifest(session_launcher, app_name, resource_class, project)
+        app_name = generate_app_name(project.slug, session_launcher.id)
+        labels = _app_labels(session_launcher, project)
+
+        work_dir = session_launcher.environment.working_directory or _DEFAULT_WORK_DIR
```
olevski: If we follow the same idea as what the session do then we try to get the work dir from the image before we fall back to a hardcoded default.

olevski: See line 892 in components/renku_data_services/notebooks/core_sessions.py to see how we do this for sessions.

components/renku_data_services/renku_apps/k8s_client.py
```diff
@@ -68,8 +164,33 @@ async def create_app_deployment(
         created = await self.__client.create(
             meta.with_manifest(manifest.model_dump(exclude_none=True, mode="json")), refresh=True
         )
+
+        owner_reference = _service_owner_reference(app_name, str(created.manifest.metadata.uid))
+        for resource in dc_resources:
+            await self._create_owned(resource.secret, _SECRET_GVK, resource.name, cluster, owner_reference)
+            await self._create_owned(resource.pvc, _PVC_GVK, resource.name, cluster, owner_reference)
+
```
olevski, line 198: We need a way to handle partial creation or failed creation. A few things here:
- what happens if the pvc or secret with the same name already exists from a failed create atempt that has not been cleaned up
- how do we clean or recover up if creating the DC resources fail

olevski: You may want to address this in a followup pr if you prefer. Or we can sit and talk about this. You can take a look at how this is handled in the sessions.

components/renku_data_services/renku_apps/k8s_client.py
```diff
@@ -191,8 +265,58 @@ def _resources_from_resource_class(resource_class: ResourceClass) -> dict[str, A
...
+    resource_class: ResourceClass | None,
+    project: Project,
+    labels: dict[str, str],
+    volumes: list[dict[str, Any]],
+    volume_mounts: list[dict[str, Any]],
+    affinity: dict[str, Any] | None,
+    tolerations: list[dict[str, Any]],
 ) -> KnativeService:
     """Build a Knative Service manifest derived from the session launcher."""
     environment = session_launcher.environment
```
olevski, line 365: I missed this from a previous review or PR. So I cannot select the code I am actually talking about but it is just below this point. When you create the security context you also need to: drop all capabilities, set privileged to False and also set privilege escallation to false.

components/renku_data_services/renku_apps/k8s_client.py
```diff
@@ -207,23 +331,41 @@ def _build_app_deployment_manifest(
...
+    env: list[dict[str, str]] = [
+        {"name": "RENKU_SESSION_IP", "value": "0.0.0.0"},  # nosec B104
+        {"name": "RENKU_SESSION_PORT", "value": str(environment.port)},
+    ]
+    reserved_names = {var["name"] for var in env}
+    env.extend(
+        {"name": var.name, "value": var.value or ""}
+        for var in session_launcher.env_variables or []
+        if var.name not in reserved_names
+    )
```
olevski: If you set the dictionary first with session_launcher.env_variables and then you set the `RENKU_*` env variables then you dont have to check for thie reserved names. Because if there are clashes you will overwrite with the `RENKU_*` env vars which is what we want.

components/renku_data_services/renku_apps/k8s_client.py
```diff
@@ -132,53 +253,6 @@ async def list_app_deployments(self, project_id: ULID | None = None) -> AsyncGen
...
-        )
-        patch_body: dict[str, Any] = {
-            "spec": {"template": {"metadata": {"annotations": {_MAX_SCALE_ANNOTATION: max_scale}}}}
-        }
-        updated = await self.__client.patch(meta, patch_body)
-        return _extract_runtime_state(KnativeService.model_validate(updated.manifest))
-
 
 def _resources_from_resource_class(resource_class: ResourceClass) -> dict[str, Any]:
     """Build a k8s container resources block from a resource class."""
```
olevski, line 284: Note for when we merge to main... We added a new feature recently on the resource pool that sets a factor to derive cpu limits based on cpu request. We should take that into account here. The factor is optional and if unset then there are no cpu limits.

components/renku_data_services/renku_apps/k8s_client.py
olevski: It should look something like this:
```
securityContext:
      allowPrivilegeEscalation: false
      capabilities:
        drop:
        - ALL
      privileged: false
      runAsNonRoot: true
      runAsUser: <some-id>
      runAsGroup: <some-id>
```

olevski (changes requested):

components/renku_data_services/session/models.py
wesjdj: Yep, you're right, thanks!

components/renku_data_services/session/models.py
wesjdj: I've removed this now anyway since I'm using the newly-added `none` frontend instead

components/renku_data_services/renku_apps/api.spec.yaml
wesjdj: Good point, done, thanks

components/renku_data_services/renku_apps/data_connectors.py
wesjdj: Done, thanks!

components/renku_data_services/renku_apps/k8s_client.py
wesjdj: I was using the project id slug as the app URL suffix, but you're right, I can probably find a way to get the same value through knative.

components/renku_data_services/renku_apps/k8s_client.py
wesjdj: You're right, I've removed it and kept `blockOwnerDeletion`.

components/renku_data_services/renku_apps/core.py
wesjdj: Correct, I wanted the dash to be the delimiter between the app prefix and the label.

components/renku_data_services/renku_apps/data_connectors.py
wesjdj: Agreed, done!

components/renku_data_services/renku_apps/k8s_client.py
wesjdj: Done, added `_resolve_work_dir`

components/renku_data_services/renku_apps/data_connectors.py
wesjdj: Done, it is now except `errors.ValidationError`

components/renku_data_services/renku_apps/k8s_client.py
wesjdj: Done, all six fields are set.

components/renku_data_services/renku_apps/k8s_client.py
wesjdj: Done, thanks!

components/renku_data_services/renku_apps/core.py
wesjdj: Done, thanks!

components/renku_data_services/renku_apps/k8s_client.py
wesjdj: Cleanup is handled now. The service is created first, then the PVC and secret with an owner reference pointing at the service. If either the PVC or the secret fails, the service is deleted, and Kubernetes then cascades the delete to whatever was already created.

components/renku_data_services/renku_apps/k8s_client.py
wesjdj: Good to know, thanks. I'll derive the CPU limit from the request once that lands on this branch, and leave it unset when there's no factor.

components/renku_data_services/renku_apps/core.py
wesjdj: On the second one, the `^` inside the brackets is a negation rather than an anchor, so `[^a-z0-9]` is "anything that isn't a letter or digit" and that's what swaps the dots and underscores for dashes.

components/renku_data_services/renku_apps/k8s_client.py
olevski: That sounds good. Thanks.

olevski (approved):

olevski (approved):

PR #1396, opened 2026-07-30
Enable read-write storage for projects

eikek:
Feature target branch for project storage PRs.

/deploy #notests


olevski:
Cleanup tasks:
- the 20 character limit on the storage class name is too small and it should have been also documented in the apispec as well, increase to 253 characters. It can only contain lowercase character, 0-9, `.` and `-`. It must end and start with a alphanumeric character.

olevski (approved):

PR #1397, opened 2026-07-31
Move project storage code from data_connectors to storage module

eikek:
Moves the project storage code from the `data_connectors` module into the `storage` module.

/deploy


components/renku_data_services/storage/api.spec.yaml
```diff
@@ -89,6 +89,248 @@ paths:
           $ref: "#/components/responses/Error"
       tags:
         - storage
+
+  /data_connectors/storage:
```
olevski: We do not need the `/data_connectors` part here anymore. Just `/storage` is enough imo.

olevski (changes requested):
One small thing Eike. You just forgot to update the paths in the apispec yaml file.

components/renku_data_services/storage/api.spec.yaml
eikek: Thanks! 🤦🏼

olevski (approved):

PR #1400, opened 2026-08-03
fix: rollback mount dir fallback logic

olevski:
/deploy

### DIFF
--- a/components/renku_data_services/notebooks/core_sessions.py
+++ b/components/renku_data_services/notebooks/core_sessions.py
@@ -145,6 +145,7 @@ async def get_extra_init_containers(
         git_providers=git_providers,
         workspace_mount_path=storage_mount,
         work_dir=work_dir,
+        mount_dir=storage_mount,
         uid=uid,
         gid=gid,
     )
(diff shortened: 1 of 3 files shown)


olevski:
A few changes were done:
- I added back the portion of code that appends `/work` to the work directory and makes that the fallback for the mount directory. Without this we overwrite the work directory from the image which results in files in the work directory that may be needed by the user being overwritten.
- The git clone init container clones to the work directory. But if the work directory is not a child of the mount directory then the cloned repos never carry over to the session. So in this case we clone straight to the mount directory.

components/renku_data_services/notebooks/core_sessions.py
```diff
@@ -849,22 +849,14 @@ async def get_mount_work_dir(
         image_workdir = await image_check_repo.image_workdir(user, parsed_image)
         work_dir_fallback = PurePosixPath("/home/jovyan/work")
         work_dir = image_workdir or work_dir_fallback
-    storage_mount_fallback = work_dir
+    # NOTE: The fallback should have /work appended to the path to avoid collisions
+    # between the mount and workdir and for backwards compatibility reasons.
+    # If /work is not added then we overwrite the workdir from the image which may contain files that are needed.
+    storage_mount_fallback = work_dir / "work"
```
olevski, line 855: This puts back the logic that we already had.
See https://github.com/SwissDataScienceCenter/renku-data-services/blob/v0.78.1/components/renku_data_services/notebooks/core_sessions.py#L927

leafty (dismissed):

olevski:
Sorry @leafty I have to update some of the tests. I will ask you  to reapprove.

leafty (approved):

PR #1406, opened 2026-08-07
chore: centralize rclone config handling

olevski:
Centralize rclone configuration handling

Renku-specific rclone "providers" (polybox, switchDrive, openbis, the fake Switch S3 provider, envidat_v1) have to be rewritten into real rclone configs before rclone sees them. That translation was duplicated in four places — RCloneValidator.get_real_configuration, two transform_* statics on the validator, and a hand-rolled copy in RCloneStorage.config_string — which had already drifted apart.

This consolidates it into a single convert_rclone_configuration() in storage/rclone.py, applied at the two boundaries where rclone actually consumes a config:

- RCloneStorage.__init__ — the INI written into the csi-rclone secret at session launch
- RCloneValidator.test_connection — the connection-test subprocess

Also in here

- get_rclone_validator() — an lru_cached shared instance, replacing per-call construction in RCloneStorage and a per-model deep copy in RCloneConfig's PrivateAttr default.
- DOI metadata unified on schema.org. The Invenio and Dataverse response parsers and the rclone backend metadata subprocess call are replaced by a single get_metadata() path. Global data connector prevalidation no longer shells out to rclone at all.

Two still left todo:

- core.py:230 — ~the provider override TODO. Nothing sets configuration["provider"] for rclone's doi backend now that get_doi_metadata is gone. This is the only remaining item that could change behaviour.~ the stuff we implemented in the rcloen cli can auto-detect the stuff that we let through, for other things we modify the storage config
- models.py:128 — ~PrevalidatedGlobalDataConnector.rclone_metadata is no longer rclone metadata and nothing reads it.~ done this was removed

/deploy


components/renku_data_services/data_connectors/core.py
```diff
@@ -424,40 +391,6 @@ def validate_data_connector_secrets_patch(
...
-    def __init__(self, *, convert_charrefs: bool = True) -> None:
-        super().__init__(convert_charrefs=convert_charrefs)
-        self._text = ""
-
-    @property
-    def text(self) -> str:
-        return self._text
-
-    def handle_data(self, data: str) -> None:
-        self._text += data
```
olevski, line 458: All of this simply got moved to another file to avoid circular imports. Moved to `components/renku_data_services/data_connectors/doi/metadata.py`

components/renku_data_services/notebooks/api/schemas/cloud_storage.py
```diff
@@ -193,48 +198,6 @@ def config_string(self, name: str) -> str:
...
-            self.configuration["url"] = "https://polybox.ethz.ch/remote.php/webdav/"
-        elif access == "personal" and storage_type == "switchDrive":
-            self.configuration["url"] = "https://drive.switch.ch/remote.php/webdav/"
-
-        # Extract the user from the public link
-        if access == "shared" and storage_type in {"polybox", "switchDrive"}:
-            public_link = self.configuration.get("public_link", "")
-            user_identifier = public_link.split("/")[-1]
-            self.configuration["user"] = user_identifier
-
```
olevski, line 237: All of these already existed as separate functions that modify the rclone config in `storage/rclone.py`. So now instead of having the same definition in 2 places we just use those transforming functions every time we need to do something like this.

components/renku_data_services/storage/models.py
```diff
@@ -2,164 +2,19 @@
...
-        case "s3":
-            return from_s3_url(parsed_url)
-        case "azure" | "az":
-            return from_azure_url(parsed_url)
-        case "http" | "https":
-            return _from_ambiguous_url(parsed_url)
-        case _:
-            raise errors.ValidationError(message=f"Scheme '{parsed_url.scheme}' is not supported.")
-
-
```
olevski, line 162: All of this moved to storage/rclone.py

components/renku_data_services/storage/rclone.py
```diff
@@ -519,10 +445,233 @@ def get_private_fields(
...
+
+    match parsed_url.scheme:
+        case "s3":
+            return from_s3_url(parsed_url)
+        case "azure" | "az":
+            return from_azure_url(parsed_url)
+        case "http" | "https":
+            return _from_ambiguous_url(parsed_url)
+        case _:
+            raise errors.ValidationError(message=f"Scheme '{parsed_url.scheme}' is not supported.")
```
olevski, line 677: Moving stuff over from notebooks/api/schemas/cloud_storage.py

components/renku_data_services/data_connectors/doi/metadata.py
```diff
@@ -139,6 +39,111 @@ async def _get_envidat_metadata(metadata_url: str) -> models.DOIMetadata | None:
...
+    def __init__(self, *, convert_charrefs: bool = True) -> None:
+        super().__init__(convert_charrefs=convert_charrefs)
+        self._text = ""
+
+    @property
+    def text(self) -> str:
+        return self._text
+
+    def handle_data(self, data: str) -> None:
+        self._text += data
```
olevski, line 151: Stuff moved over from data_connectors/core.py

components/renku_data_services/storage/rclone.py
```diff
@@ -226,59 +181,9 @@ def inject_default_values(self, config: Union[RCloneConfig, dict[str, Any]]) ->
...
-    @staticmethod
-    def transform_envidat_config(configuration: RCloneConfig | dict[str, Any]) -> RCloneConfig | dict[str, Any]:
-        """Used to convert the configuration for Envidat into a real configuration."""
-        storage_type = configuration.get("type")
-        if storage_type is None:
-            return configuration
-        if storage_type != ENVIDAT_V1_PROVIDER:
-            return configuration
-        configuration["type"] = "doi"
-        return configuration
```
olevski, line 281: Both of these are just pulled out as standalone functions further down in the same file.

components/renku_data_services/data_connectors/doi/models.py
```diff
@@ -160,11 +106,39 @@ class SchemaOrgDataset(BaseModel):
...
     publisher: SchemaOrgPublisher | None = None
+    provider: SchemaOrgProvider | None = None
+    _parsed_keywords: list[str] | None = PrivateAttr(default=None, init=False)
+
+    @staticmethod
+    def _parse_keywords(val: Any) -> list[str]:
+        if isinstance(val, list):
+            if not all([isinstance(i, str) for i in val]):
+                raise errors.ValidationError(message=f"Cannot parse keywords {val} for Schema.org dataset")
+            return [i.strip() for i in val if len(i) > 0]
```
eikek: Here I'm thinking what if a element of the list is a string with a `,`? Because for a single str we are splitting on `,` so it seems to be allowed input.

components/renku_data_services/notebooks/data_sources.py
```diff
@@ -221,23 +221,23 @@ async def _get_oauth2_configuration_part(
...
+                f"{data_connector.storage.configuration['type']} "
                 f"because no provider of kind {provider_kind.value} was found."
             )
             return None
         connection = provider.connected_user.connection if provider.connected_user else None
         if connection is None:
             logger.info(
                 f"Skipping data connector {str(data_connector.id)} of type "
-                f"{data_connector.storage.configuration["type"]} "
+                f"{data_connector.storage.configuration['type']} "
```
eikek: single quotes :-)

eikek (approved):
👍🏼

components/renku_data_services/data_connectors/doi/models.py
olevski: I changed it to take this into account. .split is now called on every member of the list if the input is a list.

eikek (approved):

PR #1418, opened 2026-08-19
feat: show Group’s Resource Pool

boucherjeanphi:
/deploy #notests


components/renku_data_services/crc/db.py
```diff
@@ -1701,6 +1701,34 @@ async def update_user(self, api_user: base_models.APIUser, keycloak_id: str, **k
...
+            raise errors.MissingResourceError(
+                message=f"The group with slug {group_slug} does not exist or you do not have access to it"
+            )
+
+        async with self.session_maker() as session, session.begin():
+            allowed_ids = [int(rp) async for rp in self.authz.get_group_resource_pools(str(group.id))]
+            stmt = select(schemas.ResourcePoolORM).options(selectinload(schemas.ResourcePoolORM.classes))
+            stmt = stmt.where(
+                or_(
+                    schemas.ResourcePoolORM.public == true(),
```
olevski: I think this is redundant. Because a public resource pool may not be assigned to the group explicitly. And even if it was that information will come from authzed anyhow.

I think this will show all resource pool explicitly associated with the group + all public resource pools.

But we only want to show all resource pools explicitly associated with the group. So I would remove this part.

olevski: P.s. If you make this change we should change the tests too. Just a reminder to not forget.

olevski (changes requested):
It looks great Jean Philippe. One small remark. For this we should probably confirm with Laura.

olevski (approved):

